PR #1, opened 2015-12-02
Feature nxp 18434 document view layout

Gabez0r:


app/elements/nuxeo-document-view/nuxeo-document-view.html
```diff
@@ -78,6 +288,112 @@
...
+            'SerialDocumentReview': 'Serial document review'
+          };
+          this.tags = ['tag1', 'tag2', 'tag3'];
+        },
+
+        _getProcessLabel: function(process) {
+          return this.processLabels[process];
+        },
+
+        _formatDate: function(date) {
```
akervern: Why not using momentjs to standardize all date format?

nelsonsilva: :+1:  we need to start building a helpers library. I think we can add these to either a behaviour we mixin or just add it as a base Polymer elements feature to make it directly available in databinding expressions

PR #5, opened 2016-04-29
Feature nxp 19337 routing

afjusto:


app/elements/nuxeo-routing.html
```diff
@@ -0,0 +1,54 @@
...
+limitations under the License.
+
+Contributors:
+  Andre Justo <ajusto@nuxeo.com>
+-->
+<script>
+  (function() {
+    Polymer.Base._addFeature({
+      router: {
+        browse: function(params) {
```
nelsonsilva: Let's go with actual arguments instead of a 'params' array to have proper names

nelsonsilva:
👍  (with remarks)

PR #20, opened 2017-03-03
NXP-21679: display a banner to open the mobile app from the Web UI

ataillefer:


elements/nuxeo-mobile/nuxeo-mobile-banner.html
```diff
@@ -0,0 +1,147 @@
...
+    <style>
+      :host {
+        text-align: right;
+        position: fixed;
+        z-index: 500;
+        bottom: 0;
+        left:0;
+        right: 0;
+        background-color: rgba(255, 255, 255, .8);
+        padding: 1vh;
```
Gabez0r: This displays a white stripe on the bottom even if browsing on desktop.

i18n/messages-fr.json
```diff
@@ -280,6 +280,7 @@
   "label.dublincore.title":"Titre",
   "label.dublincore.version":"Version",
   "label.myCollections.title":"Mes collections",
+  "label.mobile.openInApp": "Ouvrir l'application",
```
Gabez0r: No need to add the translation here. It will be overridden from Crowdin.

elements/nuxeo-mobile/nuxeo-mobile-banner.html
ataillefer: Right, fixing it.

i18n/messages-fr.json
ataillefer: :+1:

Gabez0r:
👍

guirenard:
:+1:

elements/nuxeo-mobile/nuxeo-mobile-banner.html
adilioaraujo: **Suggestion:** don't know if it's against the mockups, but changing the banner button position to the center will make the create button accessible (and also the Administrator menu icon) that are being overlapped by the banner.

```
#mobileBanner {
  text-align: center;
  position: fixed;
  z-index: 500;
  bottom: 0;
  left: 80px;
  right: 80px;
  padding: 1vh;
}
```

lkemen: Removing the bar background is risky since we don't know how people will customise the login page and the UIs. That's why we have one, and we need to keep the bar consistency between the login page, the JSF UI & WEB UI.
Even if it's not perfect I would go for having the left menu displayed on top of the mobile banner when sliding it. We need to fix the overlay on the WEB UI side but that's it.

wdyt?

adilioaraujo: yes you're right, I just did a quick test and the create button and menu icon were unreachable... but yes, it can be fixed with z-index or something and keep the background.

lkemen: I'm afraid we should resolve it that way and play with indexes between the default things of the UI, the side menu, the banner & the popups overlay 🎆

ataillefer: OK merging as it is, thanks guys!

PR #22, opened 2017-03-09
Fix nxp 19961 task web ui

guirenard:


elements/nuxeo-app/nuxeo-app.html
```diff
@@ -303,7 +303,8 @@
 
           <nuxeo-slot slot="PAGES" model="[[actionContext]]"></nuxeo-slot>
 
-          <nuxeo-layout id="home" name="home" href="[[resolveUrl('../nuxeo-home.html')]]" error="[[i18n('app.homeNotFound')]]"></nuxeo-layout>
+          <nuxeo-layout id="home" name="home" href="[[resolveUrl('../nuxeo-home.html')]]" error="[[i18n('app.homeNotFound')]]" on-element-changed="_homeLoaded">
```
nelsonsilva: `<nuxeo-layout>` has a `model` attribute, similar to the one in `<nuxeo-slot>` do it would be better to set up a computed property like `_homeModel(tasks)` which would be recomputed when the tasks are loaded.

guirenard:
`model="[[_homeModel(tasks)]]"` works great! Awesome. Thanks @nelsonsilva

PR #29, opened 2017-03-16
NXP-21746 : disables clipboard buttons when not in a container. Adds …

pncosta:
…tip message to help user

### DIFF
--- a/elements/nuxeo-clipboard/nuxeo-clipboard.html
+++ b/elements/nuxeo-clipboard/nuxeo-clipboard.html
@@ -87,6 +87,17 @@ Contributors:
         @apply(--nx-actions);
       }
 
+      .tip {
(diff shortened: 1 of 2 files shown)


elements/nuxeo-clipboard/nuxeo-clipboard.html
```diff
@@ -87,6 +87,17 @@
         @apply(--nx-actions);
       }
 
+      .tip {
+        border: 1px solid #e6e9ef;
```
adilioaraujo: Avoid using hardcoded color, in this case could be something like var(--nuxeo-text-light, #e6e9ef);
However, a suggestion is to remove .tip class, rename .emptyResult and reuse it for both the empty and the tip messages.

PR #41, opened 2017-04-13
NXP-20659 : Replaces date text ocurrences with  with nuxeo-date. Adds

pncosta:
relative time to document activity.


pncosta:
for discussion.
for now, not doing on layouts to avoid conflits with @adilioaraujo

PR #42, opened 2017-04-17
NXP-21539 and NXP-21541: move default contributions to slot contents

Gabez0r:


elements/elements.html
```diff
@@ -202,14 +201,14 @@
...
 <link rel="import" href="search/nuxeo-search-form.html">
-<link rel="import" href="nuxeo-collections/nuxeo-collection-default-actions.html">
 <link rel="import" href="nuxeo-collections/nuxeo-collection-move-up-action.html">
 <link rel="import" href="nuxeo-collections/nuxeo-collection-move-down-action.html">
 <link rel="import" href="nuxeo-collections/nuxeo-collection-move-top-action.html">
 <link rel="import" href="nuxeo-collections/nuxeo-collection-move-bottom-action.html">
 <link rel="import" href="nuxeo-collections/nuxeo-collection-remove-action.html">
 <link rel="import" href="nuxeo-suggester/nuxeo-suggester.html">
 <link rel="import" href="nuxeo-tasks/nuxeo-tasks-list.html">
+<link rel="import" href="nuxeo-tasks/nuxeo-tasks-drawer.html">
```
nelsonsilva: could you plz do a quick review with @adilioaraujo on the naming of drawer / main page elements?

elements/nuxeo-app.html
```diff
@@ -240,61 +234,22 @@
...
-          <nuxeo-slot slot="SEARCH_MENU_BUTTONS" model="[[actionContext]]"></nuxeo-slot>
-          <nuxeo-menu-icon name="tasks" icon="icons:check-box" id="menuTasks" label="app.tasks" badge="[[taskCount]]"></nuxeo-menu-icon>
-          <nuxeo-menu-icon name="favorites" icon="icons:star" label="app.favorites"></nuxeo-menu-icon>
-          <nuxeo-menu-icon name="collections" icon="icons:book" label="app.collections"></nuxeo-menu-icon>
-          <nuxeo-menu-icon name="personalWorkspace" icon="icons:folder-shared" label="app.personalSpace"></nuxeo-menu-icon>
-          <nuxeo-menu-icon name="clipboard" icon="icons:content-paste" label="app.clipboard" badge="[[clipboardCount]]"></nuxeo-menu-icon>
-          <nuxeo-menu-icon name="administration" icon="icons:build" label="app.administration" class="admin-icon" hidden$="[[!_hasAdministrationPermissions(currentUser)]]"></nuxeo-menu-icon>
+          <nuxeo-slot slot="DRAWER_ITEMS" model="[[actionContext]]"></nuxeo-slot>
+          <nuxeo-menu-icon name="administration" icon="icons:build" label="app.administration" class="admin-icon"
```
nelsonsilva: Let's create a ticket to move these to a new slot afterwards

elements/nuxeo-app.html
```diff
@@ -462,7 +406,18 @@
 
           actionContext: {
             type: Object,
-            computed: '_actionContext(currentDocument, currentUser)'
+            computed: '_actionContext(currentDocument, currentUser, tasks, currentTask, taskCount,' +
+            ' clipboard, clipboardDocCount)'
```
nelsonsilva: do we need `taskCount` and `clipboardDocCount`? can't we bind to properties in `tasks` and `clipboard`?

elements/nuxeo-app.html
```diff
@@ -501,16 +459,21 @@
 
           window.addEventListener('nuxeo-documents-deleted', function(e) {
             this._fetchTasks();
-            this.$.navTree.removeDocuments(e.detail.documentIds);
+            var navTree = this.$$('navTree');
```
nelsonsilva: We might have multiple trees so perhaps we should do `$$('nuxeo-document-tree').forEach(..)`

elements/nuxeo-app.html
```diff
@@ -501,16 +459,21 @@
 
           window.addEventListener('nuxeo-documents-deleted', function(e) {
```
nelsonsilva: why are these listeners registered here and in the `listeners` block?

elements/nuxeo-app.html
```diff
@@ -501,16 +459,21 @@
...
           window.addEventListener('nuxeo-documents-deleted-error', function(e) {
             this._documentsDeletedError(e.detail.response, e.detail.documents);
           }.bind(this));
 
-          this.clipboard = this.$.clipboard;
-
-          this.$.collectionResults.nxProvider = this.$.collectionsForm.nxProvider;
+          window.addEventListener('selected-tab-changed', function(e) {
+            if (e.srcElement === this.$$('#browser')) {
+              this.set('docAction', e.detail.value);
```
nelsonsilva: I guess this is to handle the `@<action>` part of the route, perhaps we should make this less specific to the browser and pass if to any main page content

elements/nuxeo-app.html
```diff
@@ -544,14 +507,20 @@
...
+            browser.docAction = action
+          }
           this.$.doc.get().then(function(doc) {
             if (this.docId && !doc.isVersion) {
               this.docId = '';
               this.docPath = doc.path;
               window.page('/browse' + doc.path);
             } else {
-              this.$.recent.add(doc);
+              var recent = this.$$('#recent');
```
nelsonsilva: good candidate for a datastore singleton because even if the recents list is not there we might want to track them.. let's open a ticket about it

elements/nuxeo-app.html
```diff
@@ -846,6 +836,71 @@
...
+        _clipboardUpdated: function(e) {
+          this.clipboard = this.clipboard || this.$$('#clipboard');
+          this.set('clipboardDocCount', e.detail.docCount);
+        },
+
+        _updateDrawer: function() {
+          this.$['drawer-pages'].items.forEach(function(drawerPage) {
+            switch (drawerPage.tagName) {
+              case 'NUXEO-COLLECTIONS':
+                this.collectionsForm = drawerPage;
```
nelsonsilva: we should probably use a getter for this instead of a property and we would look it up only when the page changed to a `nuxeo-collection`

PR #48, opened 2017-05-08
NXP-19179: dirty check document properties when editing

nelsonsilva:


nelsonsilva:
@Gabez0r @adilioaraujo updated to copy over dirty properties from the document since we can't do partial updates of complex properties or lists

PR #50, opened 2017-05-12
NXP-22300: initialize complex properties on document creation and import

Gabez0r:


elements/nuxeo-document-creation/nuxeo-document-creation-behavior.html
```diff
@@ -18,166 +18,216 @@
...
 
-    suggesterChildren: {
-      type: Array,
-      observer: '_suggesterChildrenChanged',
-      notify: true
+    _initializeFields: function(doc, node, nodePath) {
+      if (typeof nodePath === 'undefined') {
+        nodePath = '';
+      }
+      var prefix = (node['@prefix'] ? node['@prefix'] : node['name']) + ':';
```
nelsonsilva: nit: the ":" is not part of the prefix :P I'd remove the parenthesis and add the ":" when the field is actually used

elements/nuxeo-document-creation/nuxeo-document-creation-behavior.html
```diff
@@ -18,166 +18,216 @@
...
+      Object.keys(node.fields).forEach(function(field) {
+        if (node.fields[field] !== null && typeof node.fields[field] === 'object' &&
+            'type' in  node.fields[field] && node.fields[field].type === 'complex') {
```
nelsonsilva: why not just `node.fields[field].type && .. `instead of the `in`

nelsonsilva: nit: also, I'd probably use `fieldName` and introduce a `var field = node.fields[field]` to make it a bit more concise

PR #56, opened 2017-05-30
NXP-22442: add default selection when the drawer is opened on mobile

pncosta:


elements/nuxeo-app.html
```diff
@@ -644,6 +644,10 @@
           if (drawerPanel.narrow) {
             drawerPanel.openDrawer();
           }
+          if (!this.selectedTab){
+            var drawer = this.$['drawer-pages'];
+            this.selectedTab = drawer.childElementCount > 0 ? drawer.firstElementChild.getAttribute('name') : '';
```
nelsonsilva: can we make `selectedTab` use two way binding and then just do `drawer.selectIndex(0)`?
or just do `drawer.selectIndex(0)` followed by `this.selectedTab = drawer.selected`

PR #65, opened 2017-07-04
NXP-22679: add nuxeo-search-page and refactor search layouts

Gabez0r:


Gabez0r:
We need a better name for `nuxeo-search-page`.

elements/search/nuxeo-search-page.html
```diff
@@ -0,0 +1,354 @@
...
+        overflow: hidden;
+        white-space: nowrap;
+        display: block;
+      }
+
+      ::content .capitalize {
+        text-transform: capitalize;
+      }
+
+      ::content .unfocused-line.paper-input-container {
```
adilioaraujo: This selector can be removed

elements/search/nuxeo-search-page.html
```diff
@@ -0,0 +1,354 @@
...
+        background-color: transparent;
+      }
+
+      ::content paper-input {
```
adilioaraujo: This selector can be removed.

elements/search/nuxeo-search-page.html
```diff
@@ -0,0 +1,354 @@
...
+        --paper-input-container-underline-focus: {
+          background-color: white;
+        };
+
+        --paper-input-container-label: {
+          font-family: var(--nuxeo-app-font);
+        };
+      }
+
+      ::content #checkbox.paper-checkbox {
```
adilioaraujo: This selector can be removed

elements/search/nuxeo-search-page.html
```diff
@@ -0,0 +1,354 @@
...
+         * An object propagating key/values served by enclosing slot contents.
+         */
+        model: {
+          type: Object
+        },
+        _opened: {
+          type: Boolean,
+          value: false
+        },
+        _nxProvider:HTMLElement
```
adilioaraujo: Missing space `_nxProvider: HTMLElement`

adilioaraujo:
:+1:

elements/search/nuxeo-results-view.html
```diff
@@ -0,0 +1,341 @@
...
+         * If `true`, display a toggle control to enable or disable auto mode.
+         */
+        displayAutoControl: {
+          type: Boolean,
+          value: false
+        },
+        /**
+         * An object propagating key/values served by enclosing slot contents.
+         */
+        model: {
```
Gabez0r: This only makes sense if we decide to inject the context on `nuxeo-slot`.

Gabez0r:
`nuxeo-search-page` was renamed to `nuxeo-results-view`.

elements/search/nuxeo-results-view.html
```diff
@@ -0,0 +1,341 @@
...
+        var count = this._getFilterCount();
+        return this._opened || count === 0 ? 'hidden' : '';
+      },
+
+      _paramsChanged: function() {
+        if (this.params) {
+          // if the supplied params are a string, parse them; otherwise clone the object
+          this._params = JSON.parse(typeof this.params === 'string' ? this.params : JSON.stringify(this.params));
+          // cache the number of this.params to avoid having to parse json again if this.params is a string
+          this._paramsCount = Object.keys(this._params).length;
```
nelsonsilva: if only use case for `_paramsCount` is the filter count then I'd rather just store `_filterCount`

elements/search/nuxeo-results-view.html
```diff
@@ -0,0 +1,341 @@
...
+       provider="[[provider]]"
+       page-size="[[pageSize]]"
+       aggregations="{{aggregations}}"
+       enrichers="thumbnail, permissions, highlight"
+       params="[[_params]]"
+       quick-filters="{{quickFilters}}"
+       schemas="[[schemas]]"
+       loading="{{loading}}"
+       headers='{"X-NXfetch.document": "properties", "X-NXtranslate.directoryEntry": "label"}'
+       fetch-aggregates
```
nelsonsilva: we don't expose `headers`, `enrichers` and `fetch-aggregates` as properties?

elements/search/nuxeo-results-view.html
```diff
@@ -0,0 +1,341 @@
...
+       on-error="_onError">
+    </nuxeo-page-provider>
+
+    <div class="form" hidden$="[[!showFilters]]">
```
nelsonsilva: Why not a `dom-if` to avoid stamping? I'd rather hide when we allow easy toggling, when it's meant to be set once then we can avoid instantiating a bunch of stuff with a conditional template...

elements/search/nuxeo-results-view.html
```diff
@@ -0,0 +1,341 @@
...
+         * The aggregations returned by `provider`.
+         **/
+        aggregations: {
+          type: Object,
+          observer: '_aggregationsChanged'
+        },
+        /**
+         * If `true`, the current element is visible.
+         **/
+        visible: Boolean,
```
nelsonsilva: Would rather not have elements that depend on a `visible` flag to be set to true for them to work unless they're page elements

elements/search/nuxeo-results-view.html
```diff
@@ -0,0 +1,341 @@
...
+         * If `true`, automatically execute the search each time a param is changed.
+         */
+        auto: {
+          type: Boolean,
+          value: false
+        },
+        /**
+         * If `true`, display a toggle control to enable or disable auto mode.
+         */
+        displayAutoControl: {
```
nelsonsilva: can we please get rid of this :P

elements/search/nuxeo-results-view.html
nelsonsilva: so let's remove it for now.... if we don't have a ticket about it please open one so we can address it later

elements/search/nuxeo-search-form.html
```diff
@@ -522,7 +525,7 @@
       },
 
       get form() {
-        return this.$.layout.element;
+        return this.$.layout.form;
```
nelsonsilva: since we introduce `nuxeo-search-form-layout` then perhaps we can proxy the methods we need from the form so we just set `id="form"` and interact with it instead of having to expose the form in  `nuxeo-search-form-layout` (like we did for `validate` in `nuxeo-document-layout`)

elements/search/nuxeo-results-view.html
Gabez0r: 👍

elements/search/nuxeo-results-view.html
Gabez0r: 👍

elements/search/nuxeo-results-view.html
Gabez0r: 👍

elements/search/nuxeo-results-view.html
Gabez0r: 👍

elements/search/nuxeo-results-view.html
Gabez0r: This is the number of the original parameters passed to the element, which will be subtracted to the filter count displayed to the user. So, `_filterCount` would be a misleading name.

elements/search/nuxeo-results-view.html
Gabez0r: We're using in on most slots to provide lazy loading and prevent unnecessary requests.

PR #80, opened 2017-09-22
NXP-22817: unify nuxeo search results

Gabez0r:
Former `nuxeo-results-view` now accepts optionally a search form like the former `nuxeo-search-results`. Former `nuxeo-search-results` renamed to `nuxeo-search-page`, exposing the same API as `nuxeo-results-view` and wrapping it on a `nuxeo-page`.


afjusto:
👍

PR #82, opened 2017-09-29
NXP-21794: allow to close the mobile banner

ataillefer:


elements/nuxeo-mobile/nuxeo-mobile-banner.html
```diff
@@ -24,7 +25,8 @@
   <template>
     <style>
       #mobileBanner {
-        text-align: right;
+        display: flex;
+        justify-content: space-between;
         position: fixed;
         z-index: 2;
```
guirenard: seems to be not enough when drawer is opne

guirenard: screenshot from 2017-09-29 14-02-15

ataillefer: Yes `110` should be better

PR #135, opened 2018-02-16
NXP-24369: fix recently viewed documents

adilioaraujo:


elements/nuxeo-home.html
```diff
@@ -217,6 +230,25 @@
...
         }
+      },
+
+      _recentDocumentsChanged: function(docs) {
+        if (docs && docs.length > 0) {
+          var ids = docs.map(function(k) {return "'" + k.uid + "'"}).join(',');
+          this.$.recentProvider.query = 'SELECT * FROM Document WHERE ecm:uuid IN (' + ids + ')';
+          this.$.recentProvider.fetch().then(function(result) {
+            var recents = [];
+            result.entries.forEach(function(entry) {
```
afjusto:
```
var recents = docs.filter(function(doc) {
    return result.entries.filter(function(entry) {
        return entry.uid === doc.uid;
    });
});
```

afjusto:
👍

elements/nuxeo-home.html
Gabez0r: This may lead to false positives, since filter expects its callback to return a boolean but the inner filter will return an array. If no results are found by the inner filter, the empty array is still evaluated to `true`. We should check for the length instead:
```
var recents = docs.filter(function(doc) {
    return result.entries.filter(function(entry) {
        return entry.uid === doc.uid;
    }).length > 0;
});
```

guirenard:
:+1:

elements/nuxeo-home.html
```diff
@@ -217,6 +224,23 @@
...
+      _recentDocumentsChanged: function(docs) {
+        if (docs && docs.length > 0) {
+          var storage = this.$.recentStorage;
+          var provider = this.$.recentProvider;
+          var ids = docs.map(function(doc) {return "'" + doc.uid + "'"}).join(',');
+          provider.query = 'SELECT * FROM Document WHERE ecm:uuid IN (' + ids + ') AND ecm:isTrashed = 0';
+          provider.fetch().then(function(result) {
+            docs.forEach(function(doc) {
+              if (result.entries.findIndex(function(entry) {return entry.uid === doc.uid;}) === -1) {
+                storage.remove(doc);
```
Gabez0r: won't this re-trigger `_recentDocumentsChanged `?

afjusto:
👍

Gabez0r:
👍

elements/nuxeo-home.html
adilioaraujo: nope

PR #157, opened 2018-02-28
NXP-24391: improve fulltext search

adilioaraujo:


afjusto:
Do not search when `focus` is lost.

afjusto:
👍

PR #175, opened 2018-03-27
NXP-24711: fix create and import forms not submitting on enter press

mnixo:
The check `e.detail.keyboardEvent.target.tagName === 'INPUT'` is removed as it doesn't work anymore. Currently the `tagName` returns `NUXEO-DOCUMENT-LAYOUT` instead of `INPUT`. The `_submitKeyHandler` method should only be triggered by pressing the enter key from an element inside the form, so it seems to be unnecessary to check for the tag name. Will backport to the `9.10` branch when (and if) this PR is approved.

### DIFF
--- a/elements/document/nuxeo-document-create.html
+++ b/elements/document/nuxeo-document-create.html
@@ -363,10 +363,8 @@ limitations under the License.
         });
       },
 
-      _submitKeyHandler: function(e) {
-        if (e.detail.keyboardEvent.target.tagName === 'INPUT') {
-          this._create();
-        }
+      _submitKeyHandler: function() {
+        this._create();
       },
 
       _canCreate: function() {
--- a/elements/document/nuxeo-document-import.html
+++ b/elements/document/nuxeo-document-import.html
@@ -1254,11 +1254,11 @@ limitations under the License.
         });
       },
 
-      _submitKeyHandler: function(e) {
+      _submitKeyHandler: function() {
         if (this.stage === 'upload' && this._canImport()) {
           this._import();
         }
-        if (this.stage === 'customize' && e.detail.keyboardEvent.target.tagName === 'INPUT') {
+        if (this.stage === 'customize') {
           if (this._hasNextFile()) {
             this._nextFile();
           } else if (this._canImportWithMetadata()) {


guirenard:
:+1:

guirenard:
FYI, I reported to the team that it is a shame we cannot create the doc when hitting enter and I've been told it was on purpose.

IMHO, allowing to create just hitting enter is needed in term of UX

mnixo:
It would be cool to also have the escape key press to close the dialog... But for this one I just wanted to add this because it works for 8.10 and it doesn't after that, and it sure doesn't seem to be intentional.

tiagocardoso:
Guillaume,
I thought that was a bug collateral for Polymer 2 migration.
Creating on Enter was working on 9.3 (AFAIK) and we agree that it should
continue to work.

Miguel, as for escape. If the form is not filled, I agree.

elements/document/nuxeo-document-create.html
```diff
@@ -364,9 +364,7 @@
       },
 
       _submitKeyHandler: function(e) {
```
guirenard: remove `e` param -> build failure because of lint error

elements/document/nuxeo-document-import.html
```diff
@@ -1258,7 +1258,7 @@
         if (this.stage === 'upload' && this._canImport()) {
           this._import();
         }
-        if (this.stage === 'customize' && e.detail.keyboardEvent.target.tagName === 'INPUT') {
+        if (this.stage === 'customize') {
```
guirenard: idem

guirenard:
:+1:

elements/document/nuxeo-document-create.html
```diff
@@ -363,10 +363,8 @@
         });
       },
 
-      _submitKeyHandler: function(e) {
-        if (e.detail.keyboardEvent.target.tagName === 'INPUT') {
-          this._create();
-        }
+      _submitKeyHandler: function() {
+        this._create();
```
Gabez0r: The problem here is that, without this check and on some browsers, just hitting enter on the description box, to type in a new line, will trigger an unintended creation. However, as far as I know, there isn't still a standard cross-browser solution to [identify the original target](https://developer.mozilla.org/en-US/docs/Web/API/Event/Comparison_of_Event_Targets). On Chome we can rely on `event.path` and on Firefox on `event.originalTarget`, but the other browsers are kind of let in the dark.

PR #200, opened 2018-04-18
NXP-24350: add support to DnD documents to sibling containers

afjusto:


adilioaraujo:
:+1:

elements/nuxeo-results/nuxeo-document-content.html
```diff
@@ -334,9 +336,11 @@
       },
 
       _dragoverImport: function(e) {
-        e.preventDefault();
-        this.fire('notify', {message: this.i18n('documentContentView.drag.import'), duration: 0});
-        this._toggleDragging(true);
+        if (this.selectedItems.length === 0) {
```
Gabez0r: This will likely prevent documents from being imported if we happen to have documents selected.

Gabez0r:
👍

elements/nuxeo-app.html
```diff
@@ -956,6 +958,40 @@
...
+
+      _moveDocumentsToContainer: function(documents, target) {
+        this.$.moveDocumentsOp.op = 'Document.Move';
+        this.$.moveDocumentsOp.params = {target: target.uid};
+        this.$.moveDocumentsOp.input = 'docs:' + documents.map(function(doc) {
+          return doc.uid;
+        }).join(',');
+        this.$.moveDocumentsOp.execute().then(function() {
+          this.fire('document-updated');
+          this._toast(this.i18n('app.documents.dropped', documents.length, target.title));
```
nelsonsilva: nit: label key seems weird as it's a move, not really a drop, no?

elements/nuxeo-app.html
```diff
@@ -956,6 +958,40 @@
...
+      _addDocumentsToCollection: function(documents, target) {
+        this.$.moveDocumentsOp.op = 'Document.AddToCollection';
+        this.$.moveDocumentsOp.params = {collection: target.uid};
+        this.$.moveDocumentsOp.input = 'docs:' + documents.map(function(doc) {
+          return doc.uid;
+        }).join(',');
+        this.$.moveDocumentsOp.execute().then(function() {
+          this.fire('document-updated');
+          this._toast(
+              this.i18n(documents.length === 1 ? 'app.document.addedToCollection' : 'app.documents.addedToCollection')
```
nelsonsilva: nit: are we really keeping this pattern which different messages for single and multiple or going with something like "Document(s) added to the collection"?

elements/nuxeo-results/nuxeo-document-content.html
```diff
@@ -107,7 +107,8 @@
                          empty-label="[[emptyLabel]]"
                          empty-label-when-filtered="[[emptyLabelWhenFiltered]]"
                          selection-enabled display-quick-filters
-                         display-sort$="[[_canSort(document)]]" sort-label="[[i18n('documentContentView.sort.label')]]" sort-options="[[sortOptions]]">
+                         display-sort$="[[_canSort(document)]]" sort-label="[[i18n('documentContentView.sort.label')]]" sort-options="[[sortOptions]]"
+                         draggable$="[[_hasWritePermission(document)]]" is-droptarget="[[_isDroptarget]]">
```
nelsonsilva: The `is-droptarget` naming makes it look like a boolean property affecting the list itself... something like `droptarget-filter` would probably be clearer....

elements/nuxeo-results/nuxeo-document-content.html
```diff
@@ -369,6 +370,17 @@
 
       _computeVisible: function() {
         this.visible = Boolean(this.offsetWidth || this.offsetHeight);
+      },
+
+      // function used by nuxeo-data-grid and nuxeo-data-table to check if a list item is a drop target
+      _isDroptarget: function(el) {
+        var model = this.modelForElement(el);
+        return model && (Nuxeo.FiltersBehavior.hasFacet(model.item, 'Folderish') ||
```
nelsonsilva: `this.hasFacet(...)` should work, if it doesn't then we're evaluating this function in the wrong context...

elements/nuxeo-app.html
afjusto: indeed, will update

elements/nuxeo-app.html
afjusto: dunno, we already had those for collections so I just re-used them

elements/nuxeo-results/nuxeo-document-content.html
afjusto: agree, will rename it

elements/nuxeo-results/nuxeo-document-content.html
afjusto: this function is evaluated in each list implementing the draggable behavior (`iron-data-table`, `nuxeo-data-list` and `nuxeo-data-grid`) and those don't have the filter behavior

PR #221, opened 2018-05-07
NXP-24972: include cloud services and authorized apps on elements.html

mnixo:
Please let me know if there's a more appropriate location other than at the end of `elements.html`.


mnixo:
The performance drawback is significant. Another approach should be considered to handle this particular problem for the catalog generation in Designer.

PR #233, opened 2018-05-29
NXP-22921: Asset Diff Web UI

Gabez0r:


elements/diff/imports.html
```diff
@@ -0,0 +1,19 @@
+<link rel="import" href="elements/nuxeo-default-diff.html">
```
nelsonsilva: Missing copyright header

elements/diff/nuxeo-diff-behavior.html
```diff
@@ -0,0 +1,411 @@
...
+
+<script>
+  var Nuxeo = Nuxeo || {};
+
+  /**
+   * `Nuxeo.DiffBehavior` provides common behavior to elements that display differences between document properties.
+   *
+   * @polymerBehavior
+   */
+  Nuxeo.DiffBehavior = [Nuxeo.I18nBehavior, {
```
nelsonsilva: I'd rather explicitly mixing the i18n behavior than have it as part of the diff behavior

elements/diff/nuxeo-diff-behavior.html
```diff
@@ -0,0 +1,411 @@
...
+    _showArrayItem: function(arrdelta, showAll) {
+      return showAll || arrdelta.modified;
+    },
+
+    _incLevel: function(level) {
+      return level + 1;
+    },
+
+    /**
+     * Computes the indentatio style for labels.
```
nelsonsilva: nit: typo `indentation`

elements/diff/nuxeo-diff-behavior.html
```diff
@@ -0,0 +1,411 @@
...
+      }
+      return value;
+    },
+
+    _isSimpleDelta: function(delta) {
+      return this._isNotObjectNorArray(this._unwrapDelta(delta));
+    },
+
+    _isSimple: function(delta, originalValue) {
+      var simple;
```
nelsonsilva: nit: I'd rather just do `if(..) { return...} else { return ...}`

elements/diff/nuxeo-diff-behavior.html
```diff
@@ -0,0 +1,411 @@
...
+        simple = this._isSimpleDelta(delta);
+      } else {
+        simple = !this._isObject((Array.isArray(originalValue) && originalValue.length > 0) ?
+          originalValue[0] : originalValue);
+      }
+      return simple;
+    },
+
+    _getAllKeys: function(delta, originalValue, showAll) {
+      if (showAll) {
```
nelsonsilva: nit: ternary operator

elements/diff/nuxeo-diff-behavior.html
```diff
@@ -0,0 +1,411 @@
...
+      return type ? type.replace(/\[\]$/, '') : 'string';
+    },
+
+    /* delta helpers */
+
+    _nochanges: function(delta) {
+      return !delta;
+    },
+
+    _added: function(delta, hideAdditions) {
```
nelsonsilva: nit: `_hasAdditions`, `_hasDeletions`, etc.. would be more readable imho
Also, public computed properties would probably be useful and make the templates more readable

elements/diff/nuxeo-diff-behavior.html
```diff
@@ -0,0 +1,411 @@
...
+      return this._isObject(delta) && delta._t === 'a';
+    },
+
+    _objectInnerChanges: function(delta) {
+      return this._isObject(delta) && !delta._t;
+    },
+
+    /* delta helpers: unidiff */
+
+    _generateTextDiffHunks: function(text) {
```
nelsonsilva: these should be moved outside the behavior as they are basically static helpers

elements/diff/nuxeo-diff.html
```diff
@@ -0,0 +1,584 @@
...
+<link rel="import" href="../../bower_components/paper-toggle-button/paper-toggle-button.html">
+<link rel="import" href="../../bower_components/nuxeo-elements/nuxeo-document.html">
+<link rel="import" href="../../bower_components/nuxeo-ui-elements/nuxeo-filters-behavior.html">
+<link rel="import" href="../../bower_components/nuxeo-ui-elements/nuxeo-i18n-behavior.html">
+<link rel="import" href="../../bower_components/nuxeo-ui-elements/nuxeo-routing-behavior.html">
+<link rel="import" href="../../bower_components/nuxeo-ui-elements/widgets/nuxeo-select.html">
+<link rel="import" href="../../bower_components/nuxeo-ui-elements/widgets/nuxeo-card.html">
+<link rel="import" href="nuxeo-object-diff.html">
+
+<script type='text/javascript' src="../../node_modules/jsondiffpatch/dist/jsondiffpatch.umd.js"></script>
```
nelsonsilva: not published to bower?

elements/diff/nuxeo-diff.html
```diff
@@ -0,0 +1,584 @@
...
+
+      .diffPickers > .switchSides {
+        display: flex;
+      }
+
+      #diffPickerPane > .switchSides {
+        display: none;
+      }
+
+      @media (max-width: 720px) {
```
nelsonsilva: not a fan of media queries in elements tbh,,,

elements/diff/nuxeo-diff.html
```diff
@@ -0,0 +1,584 @@
...
+           */
+          docIds: {
+            type: Array,
+            value: []
+          },
+          /**
+           * The list of documents in `docIds`, fetched using the `enrichers`, `schemas`and `headers` specified in the
+           * element.
+           */
+          documents: {
```
nelsonsilva: so this is read only?

elements/diff/nuxeo-diff.html
```diff
@@ -0,0 +1,584 @@
...
+        created: function() {
+          if (!_customLoadPromise) {
+            _customLoadPromise = new Promise(function(resolve, reject) {
+              this.importHref(this.resolveUrl('imports.html'), resolve, reject);
+            }.bind(this));
+          }
+        },
+
+        _resize: function() {
+          if (document.body.offsetWidth <= 720) {
```
nelsonsilva: This should be based on the element's clientWidth not the body's

elements/diff/nuxeo-diff.html
```diff
@@ -0,0 +1,584 @@
...
+            this._setUnifiedView(this._unifiedView);
+            this.set('_showUnifiedViewControl', true);
+          }
+        },
+
+        _switchSides: function() {
+          var left = this._leftUid;
+          var right = this._rightUid;
+          //this._leftUid = null; // prevent diff with the same id
+          this.set('_rightUid', left);
```
nelsonsilva: you really only need to store one of them in a local var for the switch, right ? :P

elements/diff/nuxeo-diff.html
```diff
@@ -0,0 +1,584 @@
...
+            return current.then(next);
+          }, Promise.resolve([]));
+        },
+
+        _docIdsChanged: function() {
+          if (this.docIds && this.docIds.length > 1) {
+            var calls = [];
+            this.documents = [];
+            this.docIds.forEach(function(uid) {
+              calls.push(function() {
```
nelsonsilva: Do we really want a promise sequence or just a `Promise.all(...).then(....)` ? 
Something like `Promise.all(this.docIds.map(...)).then(...)`

elements/diff/nuxeo-diff.html
```diff
@@ -0,0 +1,584 @@
...
+        },
+
+        _fetchCommonSchemas: function(left, right) {
+          // if both have the same type, only do a single fetch
+          if (left && right && left.type === right.type) {
+            return this._fetchSchemas(left).then(function(response) {
+              return response.schemas;
+            });
+          }
+          return this._fetchSchemas(left).then(function(response1) {
```
nelsonsilva: `Promise.all([this._fetchSchemas(left), this._fetchSchemas(right)]).then(([left, right]) => ....)` (without the destructured assignement :P )

elements/diff/nuxeo-diff.html
```diff
@@ -0,0 +1,584 @@
...
+            return;
+          }
+          this._updateUrl(this._leftUid, this._rightUid);
+          var left = this.documents.find(function(doc) {
+            return doc.uid === this._leftUid;
+          }.bind(this));
+          var right = this.documents.find(function(doc) {
+            return doc.uid === this._rightUid;
+          }.bind(this));
+          this.set('_left', left);
```
nelsonsilva: afaik a `this.property = ...` is enough... the `set` is really only need for structured data, right?

elements/diff/nuxeo-diff.html
```diff
@@ -0,0 +1,584 @@
...
+            }.bind(this));
+            this.set('_delta', delta);
+          }.bind(this));
+        },
+
+        diff: function() {
+          _customLoadPromise.then(this._diff.bind(this));
+        },
+
+        _updateUrl: function(leftUid, rightUid) {
```
nelsonsilva: 🤔 the push/pop state in the element makes it difficult to move to elements and/or reuse in other applications... perhaps this should be set up in the app itself, wdyt?

elements/diff/nuxeo-document-diff-button.html
```diff
@@ -0,0 +1,71 @@
...
+-->
+<dom-module id="nuxeo-document-diff-button">
+  <template>
+    <style>
+      :host {
+        display: inline-block;
+      }
+    </style>
+
+    <template id="availability" is="dom-if" if="[[_isAvailable(selectedDocuments)]]">
```
nelsonsilva: why the id?

elements/diff/nuxeo-versions-diff-button.html
```diff
@@ -0,0 +1,94 @@
...
+            return this._getMajor(doc) === this._getMajor(this.document) &&
+                   this._getMinor(doc) === this._getMinor(this.document);
+          }.bind(this));
+          // and put the current one if the beginning of the list
+          if (currentIndex > 0) {
+            var current = versions[currentIndex];
+            versions.splice(currentIndex, 1);
+            versions.unshift(current);
+          }
+          this.fire('nuxeo-diff-documents', {
```
nelsonsilva: 🤔 I'd expect us to fire a `navigate` and somehow have the app pass the current selection to the diff element which you then get the versions of the documents....

elements/diff/nuxeo-diff.html
Gabez0r: I'm cascading the two `_fetchSchemas` calls on purpose, in order to have them sequenced, because they share the same `nuxeo-resource` element to make the request.

elements/diff/nuxeo-document-diff-button.html
Gabez0r: copy paste 😬

elements/diff/nuxeo-diff.html
Gabez0r: The app itself doesn't know about what the left and the right documents are, but `nuxeo-diff-page` does. Moving the logic there.

elements/diff/nuxeo-diff.html
Gabez0r: right!

afjusto:
👍

elements/diff/nuxeo-versions-diff-button.html
Gabez0r: Won't work with the compare versions actions, which build its own list of documents (which are the versions).

elements/diff/imports.html
Gabez0r: added

elements/diff/nuxeo-diff.html
Gabez0r: nope!

elements/diff/nuxeo-diff.html
Gabez0r: yes

elements/diff/nuxeo-diff.html
Gabez0r: We need it sequenced because they share the same `nuxeo-document` to issue the requests.

elements/diff/nuxeo-diff.html
Gabez0r: Me neither, but we need this behavior. An alternative would be to have a property like `responsive-width`, like `paper-drawer-panel` has. WDYT?

elements/diff/nuxeo-diff-behavior.html
Gabez0r: Public computed properties would be nice, but it would mean they would be computed even for custom elements that do not need those checks.

PR #255, opened 2018-06-18
NXP-25188: remove browsersync and rely on polyserve

nelsonsilva:


gulpfile.js
```diff
@@ -165,36 +160,18 @@ gulp.task('build', ['clean'], function(cb) {
...
-      }
-    },
-    // Run as an https by uncommenting 'https: true'
-    // Note: this uses an unsigned certificate which on first access
-    //       will present a certificate warning in the browser.
-    // https: true,
-    server: {
-      baseDir: ['.tmp', '.'],
-      middleware: [require('proxy-middleware')(proxyOptions)]
+    proxy: {
```
Gabez0r: need `hostname: '0.0.0.0',` in the options, in order for it to be accessible from outside

nelsonsilva: 👍 guess it should be optional but we can probably have it by default for convenience

Gabez0r:
👍

nmpcunha:
👍

PR #258, opened 2018-06-20
NXP-25202: style unresolved nuxeo-app

Gabez0r:
Note that the `[unresolved]` selector precedes all css rules to prevent bleed on shady css.

### DIFF
--- a/elements/nuxeo-app.html
+++ b/elements/nuxeo-app.html
@@ -539,6 +539,8 @@ limitations under the License.
             this.showError(404, e.reason.message, this._errorUrl());
           }
         }.bind(this));
+
+        this.removeAttribute('unresolved');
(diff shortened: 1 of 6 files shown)


guirenard:
:+1:

Gabez0r:
I agree that the css file would be more elegant and maintainable, but it's an additional request that can take a little bit on slower networks. So I believe it would be a fair tradeoff, since we shouldn't be changing this css too often...

Gabez0r:
Ended up moving the styles to a separate `css` file because we can inline the file directly on `index.jsp` as a static include, thus saving that additional request. :)

guirenard:
:+1:

nmpcunha:
👍

PR #274, opened 2018-07-05
NXP-25021: display a login banner when the session is expired

Gabez0r:
Depends on nuxeo/nuxeo-elements#27.

### DIFF
--- a/elements/nuxeo-app.html
+++ b/elements/nuxeo-app.html
@@ -40,6 +40,7 @@ limitations under the License.
 <link rel="import" href="nuxeo-app/nuxeo-page.html">
 <link rel="import" href="nuxeo-app/nuxeo-page-item.html">
 <link rel="import" href="nuxeo-app/nuxeo-offline-banner.html">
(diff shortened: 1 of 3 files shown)


elements/nuxeo-app.html
```diff
@@ -521,7 +524,8 @@
         'nuxeo-documents-deleted': '_documentsDeleted',
         'nuxeo-documents-untrashed': '_documentsUntrashed',
         'nuxeo-documents-dropped': '_documentsDropped',
-        'nuxeo-diff-documents': '_diffDocuments'
+        'nuxeo-diff-documents': '_diffDocuments',
+        'session-expired': '_handleExpiredSession'
```
afjusto: would it work if we register the listener in the `nuxeo-expired-session` element instead?

afjusto:
👍

elements/nuxeo-app.html
Gabez0r: It must be registered on a common parent, either on app or at the document level. :(

guirenard:
:+1:

PR #278, opened 2018-07-09
NXP-25340: fix add to clipboard action

afjusto:


elements/nuxeo-document-actions/nuxeo-clipboard-toggle-button.html
```diff
@@ -95,20 +100,15 @@
...
 
-      set clipboard(clipboard) {
-        if (!clipboard) {
+      _clipboardChanged: function() {
+        if (!this.clipboard) {
           return;
         }
-        this._clipboard = clipboard;
-        this._clipboard.addEventListener('documents-changed', this._update.bind(this));
+        this.clipboard.addEventListener('documents-changed', this._update.bind(this));
```
Gabez0r: Not that this is likely to happen, but shouldn't we be adding the listener o the new value and removing it from the old value (if not `undefined`)?

atchertchian:
see https://jira.nuxeo.com/browse/NXP-25925

PR #287, opened 2018-07-18
NXP-22453: add routing support to subpage navigation in nuxeo browser

afjusto:


Gabez0r:
👍

elements/nuxeo-browser.html
```diff
@@ -151,6 +148,21 @@
...
         this.navigateTo('browse', this.document.path);
+      },
+
+      _updateUrl: function() {
+        var currentUrl = window.location.href;
+        var idx = currentUrl.indexOf('?p=');
+        var newUrl = idx > -1 ? currentUrl.slice(0, idx) : currentUrl;
+        // do not append 'p' queryParam if current tab is the default one
+        if (this.selectedTab !== this.properties.selectedTab.value) {
+          newUrl += '?p=' + this.selectedTab;
```
Gabez0r: if we were to support the scenario above where we could have several query params, we could do something along these lines:
```
...
params = new URLSearchParams(window.location.hash.substring(window.location.hash.lastIndexOf('?')));
params.set('p', this.selectedTab);
newUrl += '?' + params.toString();
...
```

elements/nuxeo-browser.html
```diff
@@ -93,23 +93,20 @@
...
           value: 'view',
-          notify: true
+          notify: true,
+          observer: '_updateUrl'
         },
         cvParams: {
           type: Object,
           notify: true
         },
-        toto: {
```
nelsonsilva: 😂

elements/nuxeo-browser.html
```diff
@@ -151,6 +148,34 @@
...
+        var queryParams;
+        if (queryParamsIdx > -1) {
+          newUrl = currentUrl.slice(0, queryParamsIdx);
+          queryParams = new URLSearchParams(currentUrl.substring(queryParamsIdx));
+        } else {
+          newUrl = currentUrl;
+          queryParams = new URLSearchParams();
+        }
+        // remove 'p' queryParam if current tab is the default one
+        if (this.selectedTab === this.properties.selectedTab.value) {
```
nelsonsilva: What's up with `this.selectedTab` and `this.properties.selectedTab`?

afjusto: `this.selectedTab` -> current value
`this.properties.selectedTab.value` -> default value (https://github.com/nuxeo/nuxeo-web-ui/blob/master/elements/nuxeo-browser.html#L95)

The spec says we don't want the parameter to be added when switching to the default view (aka `view` tab)

nelsonsilva: 👍

nelsonsilva:
👍

Gabez0r:
👍

PR #315, opened 2018-08-31
NXP-25700: add cache service worker

nelsonsilva:
SW on dev mode only so no need for reload timestamp as hotreload is dev mode only.
SW matching js and html files for agressive caching of these. Other resources are not dynamically loaded so they rely on default static cache (which is disable for js and html files).
SW does network first so it hits browser cache (no stale responses unless offline).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ bower_components
 node_modules
 npm-debug.log
 dist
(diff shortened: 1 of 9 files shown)


guirenard:
:+1:

PR #319, opened 2018-09-07
NXP-25659: Add option to block modifier keys.

nmpcunha:


nmpcunha:
For now, a flag was added to block the usage of modifier keys.

Anyway, since there is already a ticket to review our accessibility shortcuts (https://jira.nuxeo.com/browse/DESIGN-250), it would be interesting to think further about this, since there are some limitations from iron-a11y-keys that won't allow us to specify shortcuts with and without modifier keys: e.g. for one action having both shortcuts _ctrl+a_ or _v_ only

elements/nuxeo-keys/nuxeo-keys.html
```diff
@@ -102,6 +93,39 @@
...
+          res.push(bind[0]);
+          return res;
+        }, []).filter(function(bind) {
+          return bind.key.toLowerCase() === this._transformKey(keyboardEvent.key) && (keyboardEvent.altKey
+              || keyboardEvent.ctrlKey || keyboardEvent.metaKey || keyboardEvent.shiftKey) === !!bind.hasModifiers;
+        }.bind(this));
+      },
+
+      /**
+       * Transforms the key to match iron-a11y-keys logic.
```
nmpcunha: This is copied (and simplified) function from iron-a11y-keys-behavior

nmpcunha:
The two different commits represent two different approaches we can take.

elements/nuxeo-keys/nuxeo-keys.html
```diff
@@ -77,6 +93,40 @@
...
         }
         this.fire('pressed', e.detail, {});
+      },
+
+      _getMatchingKeyBindings: function(keyboardEvent) {
+        return this.$.a11y._keyBindings[keyboardEvent.type].reduce(function(res, bind) {
+          res.push(bind[0]);
+          return res;
+        }, []).filter(function(bind) {
+          return bind.key.toLowerCase() === this._transformKey(keyboardEvent.key)
```
Gabez0r: couldn't we ditch the reduce and just do a bind[0] here?

elements/nuxeo-keys/nuxeo-keys.html
Gabez0r: Please add this information in the comment and the link to the code as well.

Gabez0r:
👍

guirenard:
:+1:

guirenard:
glad to now be able to `Ctrl+c` within web ui :ok_hand:

PR #320, opened 2018-09-11
NXP-25512: define a dedicated prop for document fetch properties

guirenard:


elements/nuxeo-app.html
```diff
@@ -1162,9 +1162,10 @@
       },
 
       _computeHeaders: function() {
+        var fProps = (Nuxeo.UI && Nuxeo.UI.config && Nuxeo.UI.config.fetchedProperties) || 'properties';
         return (Nuxeo.UI && Nuxeo.UI.config && Nuxeo.UI.config.headers) || {
           'X-NXfetch.directoryEntry': 'parent',
-          'X-NXfetch.document': 'lock,properties',
+          'X-NXfetch.document': 'lock,' + fProps,
```
Gabez0r: I wouldn't hardcode the lock, as ideally one should be able to disable it if not needed. I understand that main purpose of this might be to make overriding it easier on the `ConfigurationService` contribution. But, this is rather a limitation of how we override arrays when contributing for this service, and Web UI should be abstracted from that, in my opinion. Other than that 👍 .

guirenard: Ok but on the other hand, the fact we need to request the lock fetch (and not an enricher) is a very weird legacy and the purpose is to kind of hide it

PR #331, opened 2018-10-04
NXP-25146: add responsive action menus to web ui

Gabez0r:
Depends on nuxeo/nuxeo-ui-elements#276.


guirenard:
:+1:

elements/diff/nuxeo-document-diff-button.html
```diff
@@ -64,6 +75,10 @@
         this.fire('nuxeo-diff-documents', {
           documents: this.selectedDocuments
         });
+      },
+
+      _computeLabel: function() {
```
nmpcunha: Should you use i18n argument?

nmpcunha:
👍

elements/diff/nuxeo-document-diff-button.html
Gabez0r: It's just a trigger.

PR #333, opened 2018-10-05
NXP-25701: Fix activity vertical alignment

atoom75:


guirenard:
:+1:

elements/nuxeo-document-activity/nuxeo-document-activity.html
```diff
@@ -29,9 +29,9 @@
 <dom-module id="nuxeo-document-activity">
   <template>
     <style>
-      .row {
-        @apply --layout-horizontal;
-        @apply --layout-center;
```
Gabez0r: Just removing `--layout-center` gives the result we want.

elements/nuxeo-document-activity/nuxeo-document-activity.html
```diff
@@ -45,13 +45,11 @@
     </style>
 
     <template is="dom-repeat" items="[[activities]]">
-      <div class="row">
+      <p>
```
Gabez0r: This adds an extra line between activity rows, which is not desirable.

elements/nuxeo-document-activity/nuxeo-document-activity.html
```diff
@@ -45,13 +45,11 @@
...
         <nuxeo-user-tag user="[[item.principalName]]"></nuxeo-user-tag>
-        <div class="value">
```
Gabez0r: Because the `value` class is not being used, the margins are a little bit off.

nmpcunha: And, just for the record, if your changes/refactor turn a class obsolete, please remove it in order to clean-up our code.

elements/nuxeo-document-activity/nuxeo-document-activity.html
nmpcunha: You have also the demos for iron-flex-layout here: https://www.webcomponents.org/element/@polymer/iron-flex-layout/demo/demo/index.html

atoom75: @Gabez0r Yes I know those mixins, they are applying flex properties to their container, but it seemed to me a little bit odd to use flexboxes here. I would have not display the `nuxeo-user-tag` in its own column because I don't know which its length could be. If it's long, the risk is that we will not have so much space for the activity value which will be display in many lines. That's why I remove this class and use an inline display, to allow the activity value to be display under the `nuxeo-user-tag`. But I agree with you, when the activity can be display in only one line, it's enough to remove the `--layout-center` property.

Gabez0r: According to the mockups linked to the ticket (hope you have access to them 🤞), the text is not supposed to go under the tag but instead remain on its own column, which is what we already have in place. We just need the user tag to be aligned to the top.

atoom75: I've just seen the solution proposed by Isabel and it confirms your suggestion so I'm going to implement it like that, and we will optimize it afterwards if necessary.

atoom75:
Code updated following your recommendations!

nmpcunha:
I think you could squash those commits, but let's wait for other review.
Despite that, 👍

atoom75:
Ok I've removed from git history my initial commit and the revert one to keep only the useful commit

Gabez0r:
👍

nmpcunha:
👍

PR #337, opened 2018-10-11
NXP-24880: introduce skipAggregagtes option on nuxeo-search-form to a…

guirenard:
…llow dynamic aggrgation computation

### DIFF
--- a/elements/search/nuxeo-results-view.html
+++ b/elements/search/nuxeo-results-view.html
@@ -79,7 +79,6 @@ limitations under the License.
     </style>
 
     <nuxeo-page-provider id="provider"
-                         auto$="[[_isAuto(auto, visible, searchForm)]]"
                          provider="[[provider]]"
                          page-size="[[pageSize]]"
                          aggregations="{{aggregations}}"
@@ -90,6 +89,7 @@ limitations under the License.
                          loading="{{loading}}"
                          headers="[[headers]]"
                          fetch-aggregates
+                         skip-aggregates$="[[skipAggregates]]"
                          on-error="_onError">
     </nuxeo-page-provider>
 
@@ -258,6 +258,11 @@ limitations under the License.
         },
 
         /**
+         * If `true`, aggregagtes from page provider definition will not be computed.
+         */
+        skipAggregates: Boolean,
+
+        /**
          * Currently selected view
          */
         view: Object,
@@ -296,10 +301,6 @@ limitations under the License.
 
       toggleExpand: function() {
         this.$$('#collapse').toggle();
-      },
-
-      _isAuto: function() {
-        return this.visible && this.auto && !this.searchForm;
       },
 
       _expandIcon: function(opened) {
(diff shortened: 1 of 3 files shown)


elements/search/default/nuxeo-default-search-form.html
```diff
@@ -136,9 +140,21 @@
         searchTerm: {
           type: String,
           notify: true
+        },
+        skipAggregates: {
```
Gabez0r: If we are supposed to receive this from the model (i.e., allow someone to pass it through as an attribute), then we shouldn't be setting it directly with an observer, or we'll lose what was passed from the model as soon as the parameters change. If we do not need this to be set from the outside, then we could use a computed property instead.

elements/search/nuxeo-search-form.html
```diff
@@ -826,6 +833,9 @@ <h2>[[i18n('searchForm.delete.heading')]]</h2>
             }
           }
         }.bind(this));
+        this.form.addEventListener('skip-aggregates-changed', function(e) {
+          this.notifyPath('skipAggregates', e.detail.value);
+        }.bind(this));
```
Gabez0r: Perhaps we should reflect all the changes on this file on `nuxeo-results-view` and `nuxeo-search-page` as well?

elements/search/default/nuxeo-default-search-form.html
guirenard: Well, here it is just an example on how to leverage this `skipAggregates` property when designing your search form to do not compute the aggregations if `fulltext` parameter not set (hence speed up the request).

Not really intended to be merged.

elements/search/nuxeo-search-form.html
guirenard: WDYM?

Gabez0r: What I meant was that `nuxeo-results-view` and `nuxeo-search-page` should perhaps expose the same skipAggregates flags, since they share similar API.

elements/search/nuxeo-results-view.html
```diff
@@ -90,6 +89,7 @@
                          loading="{{loading}}"
                          headers="[[headers]]"
                          fetch-aggregates
+                         skip-aggregates$="[[skipAggregates]]"
```
Gabez0r: We need to reflect this on `nuxeo-search-page` as well, since it wraps `nuxeo-results-view`.

elements/search/nuxeo-results-view.html
```diff
@@ -257,6 +257,11 @@
           observer: '_searchFormChanged'
         },
 
+        /**
+         * If `true`, aggregagtes from page provider definition will not be computed.
+         */
+        skipAggregates: Boolean,
```
Gabez0r: We need to reflect this on nuxeo-search-page as well, since it wraps nuxeo-results-view.

elements/search/nuxeo-results-view.html
Gabez0r: Nevermind this comment if this only needs to be read from the form. 😁

Gabez0r:
👍

PR #342, opened 2018-10-19
ELEMENTS-772: Fixed size for nuxeo-document-thumbnail

atoom75:


elements/nuxeo-document-thumbnail/nuxeo-document-thumbnail.html
```diff
@@ -31,6 +31,7 @@
         height: 32px;
         width: 32px;
         position: relative;
+        flex: 0 0 auto;
```
Gabez0r: nitpicking: I'd use `flex: none;` instead, for improved readability

nmpcunha:
I would suggest to squash those two commits.
After that and amending the message 👍

PR #344, opened 2018-10-19
NXP-25344: Use CSS variables and fix nuxeo-suggester display on mobile

atoom75:
Not really sure about the aim of this task, so I add 2 new variables in the default theme for the `paper-input` element, and I fixed a display bug on mobile when the result title is long. For that second point, you can find more informations here: https://css-tricks.com/flexbox-truncated-text/

### DIFF
--- a/elements/nuxeo-suggester/nuxeo-suggester.html
+++ b/elements/nuxeo-suggester/nuxeo-suggester.html
@@ -149,6 +149,10 @@ limitations under the License.
         @apply --layout-vertical;
       }
 
+      .item .details {
+        min-width: 1px;
+      }
+
       .item + .item {
         border-top: 1px solid var(--nuxeo-border);
       }
@@ -210,8 +214,8 @@ limitations under the License.
         }
 
         paper-input {
-          width: calc(100% - 90px);
-          margin-left: 1.2rem;
+          width: var(--nuxeo-suggester-media-width, calc(100% - 90px));
+          margin-left: var(--nuxeo-suggester-media-margin-left, 1.2rem);
         }
 
         #results {
--- a/themes/base.html
+++ b/themes/base.html
@@ -527,6 +527,12 @@ limitations under the License.
       position: relative;
       top: var(--nuxeo-app-top);
     };
+
+    --nuxeo-suggester-width: 65%;
+
+    --nuxeo-suggester-media-width: calc(100% - 90px);
+
+    --nuxeo-suggester-media-margin-left: 1.2rem;
   }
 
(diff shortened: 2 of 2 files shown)


elements/nuxeo-suggester/nuxeo-suggester.html
```diff
@@ -210,8 +214,8 @@
         }
 
         paper-input {
-          width: calc(100% - 90px);
-          margin-left: 1.2rem;
+          width: var(--paper-input-media-width);
+          margin-left: var(--paper-input-media-margin);
```
nmpcunha: As a "good practice" we should provide a default value as a failover for the variable.
e.g. var(--paper-input-media-margin, 1.2rem);

guirenard:
:+1:

guirenard:
(I agree with @nuno 's comment though)

elements/nuxeo-suggester/nuxeo-suggester.html
```diff
@@ -210,8 +214,8 @@
         }
 
         paper-input {
-          width: calc(100% - 90px);
-          margin-left: 1.2rem;
+          width: var(--nuxeo-suggester-media-width, calc(100% - 90px));
+          margin-left: var(--nuxeo-suggester-media-margin, 1.2rem);
```
Gabez0r: Since we might want to add variables to the shorthand `margin` or for the other margins, please include the `-left` in the name:
```suggestion
          margin-left: var(--nuxeo-suggester-media-margin-left, 1.2rem);
```

atoom75: ok it's fixed!

Gabez0r:
👍

nmpcunha:
👍

themes/default/theme.html
```diff
@@ -157,6 +157,11 @@
     --nuxeo-diff-added-color: #B4EFCB;
     --nuxeo-diff-deleted-color: #E6B1B1;
     --nuxeo-diff-label-color: #D4D4D9;
+
+    /* Suggester */
+    --nuxeo-suggester-width: 65%;
+    --nuxeo-suggester-media-width: calc(100% - 90px);
+    --nuxeo-suggester-media-margin-left: 1.2rem;
```
Gabez0r: Sorry, did not notice the file name, but this should be placed on the base theme instead `themes/base.html`, so that every theme can benefit from it.

atoom75: @Gabez0r Ok it's done, tell me if it's ok

nmpcunha:
👍

PR #348, opened 2018-10-26
NXP-25637: unify sorting, filtering and result actions on listings

Gabez0r:
Depends on nuxeo/nuxeo-ui-elements#292.


elements/nuxeo-results/nuxeo-document-content.html
```diff
@@ -106,8 +113,7 @@
                          class="results"
                          empty-label="[[emptyLabel]]"
                          empty-label-when-filtered="[[emptyLabelWhenFiltered]]"
-                         selection-enabled display-quick-filters
-                         display-sort$="[[_canSort(document)]]" sort-label="[[i18n('documentContentView.sort.label')]]" sort-options="[[sortOptions]]"
```
nelsonsilva: What happened to the `sort-label`?

elements/nuxeo-results/nuxeo-results.html
```diff
@@ -244,6 +333,49 @@
         this.view = null;
       },
 
+      _displayQuickFilters: function() {
+        // XXX check previous view properties for compatibility
+        return this.view && !this.view.handlesFiltering &&
```
nelsonsilva: so, if the view handles filtering I'll never be able to use quick filters from `nuxeo-results`....
what about `this.view && ((this.view.handlesFiltering && this.view.hasAttribute('display-quick-filters')) || this.displayQuickFilters)`?

elements/search/default/nuxeo-default-search-results.html
```diff
@@ -32,11 +32,15 @@
       }
     </style>
 
-    <nuxeo-results name="default" nx-provider="[[nxProvider]]" selected-items="{{selectedItems}}" display-mode="list">
+    <nuxeo-results name="default"
+                   nx-provider="[[nxProvider]]"
```
nelsonsilva: how I wish we could rename this to just `provider`....

elements/search/nuxeo-search-form.html
```diff
@@ -288,12 +289,15 @@ <h1>[[i18n(label)]]</h1>
         </div>
 
         <div id="queue" hidden$="{{!queue}}">
+          <template is="dom-if" if="[[_displayQuickFilters(_quickFilters)]]">
```
nelsonsilva: nit: for these one liners / single element I'd be in favour of using `hidden` as the dom-if will just toggle visibility anyway after the first stamp.

elements/nuxeo-results/nuxeo-document-content.html
Gabez0r: It was removed. It was decided it was no longer needed since the sort widget was self-explanatory and already inline with what we're doing with sorting on table headers.

elements/search/default/nuxeo-default-search-results.html
Gabez0r: Unfortunately we're accessing this from other elements, so if we were to do it, I guess it'd make sense on a dedicated ticket just dedicated to code refactor... and with more time... 😬

elements/search/nuxeo-search-form.html
Gabez0r: That was my first approach, unfortunately, the computed binding to `_displayQuickFilters` won't run if `_quickFilters` is undefined, which made the filters still visible (taking up space), but empty. Alternatives to still use `hidden` would be to add another computed property instead, with a default value to false...

elements/nuxeo-results/nuxeo-results.html
Gabez0r: You'll not be able to use quick filters for **that** particular view only, but they will still show up for other views. Note that this binding has the view as a dependency. If we know that the current view already shows filters, there's no point in showing them again, right? 🤔

nelsonsilva:
👍

elements/nuxeo-results/nuxeo-results.html
```diff
@@ -116,22 +129,35 @@
...
+        <div class="rightHand">
+          <slot name="actions"></slot>
+
+          <!-- 0.9 compatibility -->
+          <nuxeo-slot slot="DOCUMENT_LISTING_ACTIONS" model="[[actionContext]]"></nuxeo-slot>
+
+          <nuxeo-slot slot="RESULTS_ACTIONS" model="[[actionContext]]"></nuxeo-slot>
+
+          <template is="dom-repeat" items="[[_displayModes]]">
+            <paper-icon-button class="displayMode"
```
nmpcunha: Indentation issue

nmpcunha:
👍

guirenard:
:+1:

PR #353, opened 2018-10-30
NXP-25637: prevent fetch error when loading a view on nuxeo-results

Gabez0r:
Depends on nuxeo/nuxeo-ui-elements#295.


guirenard:
:+1:

PR #355, opened 2018-10-31
Task nxp 25938 diff title

atoom75:


elements/nuxeo-diff-page.html
```diff
@@ -23,17 +23,23 @@
...
   <template>
     <style include="nuxeo-styles">
+    .title a {
+      color: var(--nuxeo-link-hover_-_color);
+    }
     </style>
     <nuxeo-page>
       <div class="header" slot="header">
-        <span class="title">[[i18n('app.title.diff')]]</span>
+        <span class="title">[[i18n('app.title.diff')]] <a class="title" href$="[[_buildPermalink(_left)]]">[[_left.title]]</a> [[i18n('label.versus')]] <a class="title" href$="[[_buildPermalink(_right)]]">[[_right.title]]</a></span>
```
Gabez0r: 1. Please format and break this line, it's too long.
2. Title class applied to both span and the anchors inside it, which seems to be a mistake.
3. Title class is not needed, since there are no uses of `<a>` that are not titles. Or we can keep the title class on the anchors and just use it for the rule.

elements/nuxeo-diff-page.html
```diff
@@ -23,17 +23,23 @@
 <dom-module id="nuxeo-diff-page">
   <template>
     <style include="nuxeo-styles">
+    .title a {
```
Gabez0r: Just `.title` is enough.

i18n/messages.json
```diff
@@ -643,6 +643,7 @@
   "label.ui.aggregate.to_30_s": "Less than 30 s",
   "label.ui.aggregate.to_500_px": "Less than 500 px",
   "label.version": "Version",
+  "label.versus": "Vs.",
```
Gabez0r: This label should be `diffPage.versus` by convention, since it is only used on `nuxeo-diff-page.html` has no foreseeable reuse.

atoom75:
@Gabez0r I've updated my commit following your recommendations

nuxeo-csv
```diff
@@ -0,0 +1 @@
+../nuxeo-csv/nuxeo-csv-web-ui/src/main/resources/web/nuxeo.war/ui/nuxeo-csv
```
guirenard: Ooops look like you pushed your sym link ;)

nuxeo-drive
```diff
@@ -0,0 +1 @@
+../nuxeo-drive-server/nuxeo-drive-web-ui/src/main/resources/web/nuxeo.war/ui/nuxeo-drive
```
guirenard: same

nuxeo-csv
atoom75: @guirenard Arf, sorry, it's fixed

nuxeo-drive
atoom75: @guirenard Arf, sorry, it's fixed

guirenard:
:+1:

PR #365, opened 2018-11-14
NXP-26171: use Bulk.RunAction for csv export

nelsonsilva:
Depends on https://github.com/nuxeo/nuxeo/pull/2531 and https://github.com/nuxeo/nuxeo-elements/pull/40

### DIFF
--- a/elements/nuxeo-csv-export/nuxeo-csv-export-button.html
+++ b/elements/nuxeo-csv-export/nuxeo-csv-export-button.html
@@ -17,6 +17,7 @@ limitations under the License.
 
 <link rel="import" href="../../bower_components/polymer/polymer.html">
 <link rel="import" href="../../bower_components/paper-icon-button/paper-icon-button.html">
+<link rel="import" href="../../bower_components/nuxeo-elements/nuxeo-operation.html">
 <link rel="import" href="../../bower_components/nuxeo-elements/nuxeo-resource.html">
 <link rel="import" href="../../bower_components/nuxeo-ui-elements/actions/nuxeo-action-button-styles.html">
 <link rel="import" href="../../bower_components/nuxeo-ui-elements/nuxeo-i18n-behavior.html">
@@ -32,7 +33,9 @@ limitations under the License.
   <template>
     <style include="nuxeo-action-button-styles"></style>
 
-    <nuxeo-resource id="resource"></nuxeo-resource>
+    <nuxeo-operation id="op" op="Bulk.RunAction" input="[[provider]]"></nuxeo-operation>
+
+    <nuxeo-resource id="resource" path="/bulk/[[status.commandId]]"></nuxeo-resource>
 
     <div class="action" on-click="execute">
       <paper-icon-button icon="nuxeo:csv-export"></paper-icon-button>
@@ -66,7 +69,7 @@ limitations under the License.
           type: String,
         },
         /**
-         * A comman separated list of fields to be be exported.
+         * A comma separated list of fields to be be exported.
          */
         fields: {
           type: String,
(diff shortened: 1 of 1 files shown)


elements/nuxeo-csv-export/nuxeo-csv-export-button.html
```diff
@@ -32,7 +33,9 @@
   <template>
     <style include="nuxeo-action-button-styles"></style>
 
-    <nuxeo-resource id="resource"></nuxeo-resource>
+    <nuxeo-operation id="op" op="Bulk.RunAction" input="[[provider]]" params="[[_params(schemas, fields)]]"></nuxeo-operation>
```
Gabez0r: Since the schemas should be the same of the page provider if the `schemas` property is not defined, we can't use a computed binding here, since the provider's schemas can change and this change must be reflected.

elements/nuxeo-csv-export/nuxeo-csv-export-button.html
```diff
@@ -64,12 +67,14 @@
          */
         schemas: {
           type: String,
+          value: '',
```
Gabez0r: We shouldn't use `''` as default value. Since this property is a comma separated list, and an empty string means an empty array, which is still a valid input for the operation, .i.e., "just give me the system properties". We can use `undefined` here to state that the provider's schemas should be used instead (the default).

elements/nuxeo-csv-export/nuxeo-csv-export-button.html
```diff
@@ -78,35 +83,43 @@
...
-        this.$.resource.data = {};
-        var schemas = this.schemas != null ? this.schemas : this.provider.$.nxResource.schemas;
-        if (schemas) {
-          this.$.resource.data.schemas = schemas.split(',').map(function(schema) {
-            return schema.trim();
-          });
+      _params: function() {
+        var actionParams = {};
+        if (this.schemas) {
+          actionParams.schemas = this.schemas.split(',').map(function(s) { return s.trim(); });
```
Gabez0r: Need to set the provider's schemas instead if no schemas were passed to the operation. Something alongside these lines (and ditching the if):
```suggestion
          actionParams.schemas = this.schemas ? this.schemas.split(',').map(function(s) { return s.trim(); }) : this.provider.schemas;
```

elements/nuxeo-csv-export/nuxeo-csv-export-button.html
nelsonsilva: Done 👍

elements/nuxeo-csv-export/nuxeo-csv-export-button.html
nelsonsilva: Done 👍

elements/nuxeo-csv-export/nuxeo-csv-export-button.html
nelsonsilva: Done 👍 Moved parameter building to execution time just to be safe (as it was before).

elements/nuxeo-csv-export/nuxeo-csv-export-button.html
```diff
@@ -78,35 +81,41 @@
...
+        }
       },
 
       execute: function() {
-        this.$.resource.path = this.provider.$.nxResource.path.replace(/\/execute$/, '/bulk/csvExport');
-        this.$.resource.params = this.provider.$.nxResource.params;
-        this.$.resource.data = {};
-        var schemas = this.schemas != null ? this.schemas : this.provider.$.nxResource.schemas;
+        var actionParams = {};
+        var schemas = this.schemas ? this.schemas : this.provider.schemas;
```
Gabez0r: This won't allow the `schemas` property to be an empty string. We should check for `undefined` and `null`:
```suggestion
        var schemas = this.schemas != null ? this.schemas : this.provider.schemas;
```

nelsonsilva: Ok, fair enough, done 👍

Gabez0r:
👍

PR #378, opened 2018-11-20
NXP-26210: do not notify remove from clipboard because of document de…

guirenard:
…leted

### DIFF
--- a/elements/nuxeo-clipboard/nuxeo-clipboard.html
+++ b/elements/nuxeo-clipboard/nuxeo-clipboard.html
@@ -237,7 +237,6 @@ limitations under the License.
 
       remove: function(doc) {
         this.$.storage.remove(doc);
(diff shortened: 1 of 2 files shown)


elements/nuxeo-clipboard/nuxeo-clipboard.html
```diff
@@ -235,9 +235,11 @@
         return this.$.storage.contains(doc);
       },
 
-      remove: function(doc) {
+      remove: function(doc, notify) {
         this.$.storage.remove(doc);
-        this.fire('removed-from-clipboard', {docId : doc.uid});
+        if (notify) {
```
nelsonsilva: I'd rather move the firing of the event to the `_remove` method which I guess is the only place where we do want to fire it

nmpcunha:
👍

PR #384, opened 2018-11-22
NXP-25939: Reverse starting order on document diff

atoom75:
The specifications of this task was not really clear, so I assumed that the diff page should always display the current version on the right screen, and the previous one on the left screen. So if the current version is not the last one, the versions following the current one will not be displayed.


Gabez0r:
:+1:

guirenard:
:+1:

PR #400, opened 2018-12-03
 NXP-21179: add nuxeo-sardine

Gabez0r:


guirenard:
:+1:

elements/nuxeo-sardine.html
```diff
@@ -0,0 +1,230 @@
...
+              id: 'sardine',
+              icon: 'images/gift.png',
+              label: 'I love sardines!!! 🐟🐟🐟💝💝💝',
+            },
+            run: this._sardineOn.bind(this),
+          });
+        }.bind(this));
+      },
+
+      _sardineOn: function() {
```
nelsonsilva: nit: if this in `nuxeo-sardine` and the only thing we turn on / off is the sardine perhaps `_on`, '_off`, 
`animate` would be enough 🤷‍♂️

elements/nuxeo-sardine.html
```diff
@@ -0,0 +1,230 @@
...
+        this._onkeypress = this._onkeypress || function(e) {
```
nelsonsilva: why not move these handlers to actual methods `_onkeypress: function(e) {...}` ?

elements/nuxeo-sardine.html
```diff
@@ -0,0 +1,230 @@
...
+          window.addEventListener('mouseout', this._onpointerout, false);
+        }
+        window.addEventListener('keyup', this._onkeypress, false);
+      },
+
+      _animateSardine: function() {
+        if (this._idleTimeout) {
+          clearTimeout(this._idleTimeout);
+        }
+        this._idleTimeout = setTimeout(function() {
```
nelsonsilva: seems weird to register a timeout in the request animation callback...

elements/nuxeo-sardine.html
Gabez0r: Just to define them as lazily as possible.

nmpcunha:
👍

PR #406, opened 2018-12-06
NXP-26383: Fix actions overflow on attachment's blob

atoom75:


elements/nuxeo-document-blob/nuxeo-document-blob.html
```diff
@@ -42,13 +42,20 @@
       .row .info {
         @apply --layout-vertical;
         @apply --layout-flex;
+        min-width: 1px;
```
nmpcunha: That approach is valid but sounds a little bit hacky.
An overflow here may do the trick!

elements/nuxeo-document-blob/nuxeo-document-blob.html
```diff
@@ -42,13 +42,20 @@
...
       }
 
       .row .actions {
         @apply --layout-horizontal;
         @apply --layout-center;
       }
 
+      .row .info a {
```
nmpcunha: Since we are "truncating" the text, it may be useful to add title attribute to <a> element.

elements/nuxeo-document-blob/nuxeo-document-blob.html
atoom75: @nmpcunha Ok it's fixed

elements/nuxeo-document-blob/nuxeo-document-blob.html
atoom75: @nmpcunha Ok, title attribute added

nmpcunha:
👍

guirenard:
:+1:

PR #409, opened 2018-12-11
NXP-26450: nuxeo-results-view element documentation

gcarlin:


elements/search/nuxeo-results-view.html
```diff
@@ -226,7 +228,7 @@
           value: false
         },
         /**
-         * If `true`, automatically execute the search each time a param is changed.
+         * If true, automatically execute the operation when either `provider` or `params` changes.
```
guirenard: not sure `operation` is correct, we are performing a search here

guirenard:
:+1:

elements/search/nuxeo-results-view.html
Gabez0r: agree, this should be changed

Gabez0r:
👍

PR #411, opened 2018-12-13
NXP-25581: trigger refresh when task changes

nmpcunha:


elements/nuxeo-workflow-graph/nuxeo-workflow-graph.html
```diff
@@ -264,7 +268,7 @@
       },
 
       _fetchGraph: function() {
-        if (this.workflowId && this.workflowId.length > 0) {
+        if (this.taskId && this.workflowId && this.workflowId.length > 0) {
```
Gabez0r: I guess we acn cleanup the `.length > 0` check, since it will not be reached if `this.workflowId` is an empty string anyway:
```suggestion
        if (this.taskId && this.workflowId) {
```

nmpcunha: Nice catch, thanks!

guirenard:
This causes an extra request to be sent each time we navigate to a document which has a running workflow. We really need to think about limiting the number of request :)

I'd rather refactor a little bit to move the dialog inside the `nuxeo-workflow-graph` element and trigger this extra request only when and if we open the dialog.

guirenard:
Yes, but opening the graph dialog is unlikely to happen each time.

Issuing a request that will probably be useless is the worst pattern. Issuing the request, each time you open the dialog is the best thing to do IMO. Even if it implies that, in some infrequent scenario, a user will open twice the `view graph` in the doc info bar and in the task layout, it will result in 2 requests.

guirenard:
BTW
>please remind that we use nuxeo-workflow-graph in different places, which means that we may end up with the same requests or even more.

These are different elements and will trigger their requests whatsoever

elements/nuxeo-document-info-bar/nuxeo-document-info-bar.html
```diff
@@ -229,6 +208,13 @@
         }
       },
 
+      _getWorkflowCurrentTask: function(workflowId) {
+        var task =  this.tasks.find(function(item){
+          return item.workflowInstanceId === workflowId;
```
guirenard: I don't think this is robust enough. A workflow can have parallel tasks assigned to the same user (or group)...

elements/nuxeo-workflow-graph/nuxeo-workflow-graph.html
```diff
@@ -209,6 +231,17 @@
         this._initialize();
       },
 
+      show: function() {
+        if(this._needsRefetch) {
+          this.$.graphResource.execute().then(function() {
+            this._needsRefetch = false;
```
guirenard: You can have quite some complex workflow with parallel tasks and loops which would defeat your `_needsRefetch`  logic. Honestly, I wouldn't bother too much and just fetch the graph whenever the dialog is open..

guirenard:
:+1:

PR #422, opened 2018-12-18
NXP-25471: show message when no versions to compare with

nmpcunha:


elements/diff/nuxeo-versions-diff-button.html
```diff
@@ -71,6 +71,11 @@
 
       _doDiff: function() {
         this.$$('#opGetVersions').execute().then(function(result) {
+          // check if there is at least two versions to be compared
+          if (result.entries.length < 2) {
+            this.fire('notify', {message: this.i18n('versions.nothingToCompare')});
```
Gabez0r: We usually make the first part of the label the name of the current element in camelCase:
```suggestion
            this.fire('notify', { message: this.i18n('versionsDiffButton.nothingToCompare') });
```

i18n/messages.json
```diff
@@ -947,6 +947,7 @@
   "versions.info": "You're viewing the {0}.{1} version.",
   "versions.latest": "Latest",
   "versions.modified": "Modified {0} by {1}",
+  "versions.nothingToCompare": "There is no version to compare with.",
```
Gabez0r: "There are no other versions to compare with." would make more sense IMHO, since  "There is no version to compare with" suggested that can either be one or none.

guirenard:
:+1:

Gabez0r:
👍

PR #429, opened 2019-01-02
NXP-26493: change tooltip position to default bottom

nmpcunha:


nmpcunha:
It seems to be an issue with tooltip positioning on Edge.
We also tried to use `paper-tooltip` but the position is also not correct, probably due to have the tooltip inside a `dom-if`.
The best and easiest solution was to change the tooltip's position to bottom.

Gabez0r:
👍

PR #431, opened 2019-01-03
NXP-26580: fix EN label

manonlumeau:


nmpcunha:
Thanks for your contribution @manonlumeau!
We'll ask you to have a more precise message. Instead of `fix EN label` we would suggest something like 'fix EN graph error label'.

nmpcunha:
👍

Gabez0r:
Missing ticket id on commit message.

nmpcunha:
👍

PR #437, opened 2019-01-08
Fix nxp 26610 cannot create entries on empty vocabulary

nmpcunha:


guirenard:
:+1:

elements/directory/nuxeo-vocabulary-management.html
```diff
@@ -360,15 +356,37 @@ <h2>[[i18n('vocabularyManagement.popup.editEntry')]]</h2>
...
+        }.bind(this));
+      },
+
+      _getSchemaFields: function() {
+        if (this.entries.length > 0) {
+          return Promise.resolve(Object.keys(this.entries[0].properties));
+        } else {
+          this.$.schema.path = '/config/schemas/' + this._selectedSchema;
+          return this.$.schema.get()
+            .then(function(response) {
```
Gabez0r: I'd cache this response, similar to what we do on `nuxeo-diff`, since if we create an entry, then hit cancel, and then hit create again, we'll do an unnecessary request. In this case, the key of the cache would be the schema.

Gabez0r:
👍

PR #461, opened 2019-02-18
NXP-26819: preventing unnecessary rendering when no data is available.

richardsd:
Cannot cherry-pick the commit because would backport unwanted changes to the element.


richardsd:
Removed the hidden attribute, as it is not necessary with the dom-if.

Gabez0r:
👍

PR #478, opened 2019-03-01
add this.params test

cedric-legallo:
Add this.params test for nuxeo-search-forms
After a first start of the server, these lines throw an exception without these tests creating some other malfunctions.(Document Tree is empty in the drawer)

This PR is created by and for Keendoo context
We need it on the LTS so the PR is on the 10.10 SNAPSHOT branch, but can be done on master too.


guirenard:
Hi! We'll be glad to merge this PR. However, I don't reproduce on a bare nuxeo and I don't how and when `this.params` is undefined.

Could you provide reproduction steps?

nicolaschaillot:
Hi.
We solved the problem by disabling standard Nuxeo searches (trash & assets). There seems to be a conflict between our search management (all searches in a single DRAWER_ITEMS slot entry) and standard searches.

guirenard:
Ok so, you don't need this PR anymore?

PR #494, opened 2019-03-06
NXP-24988: Check for document path in nuxeo-document-tree

semisse:


nmpcunha:
I would suggest you to reword the commit, since the change is checking not for empty document, but for the existence of path. Despite that... 👍

nmpcunha:
👍

Gabez0r:
👍

PR #508, opened 2019-03-09
Task nxp 26965 s3 direct upload polymer 3

nelsonsilva:


addons/amazon-s3-online-storage/index.js
```diff
@@ -0,0 +1,178 @@
...
+      return operation.params(params).execute(options);
+    });
+  }
+}
+
+UploaderBehavior.registerProvider('s3', S3Provider);
+
+// if S3 direct upload is enabled set it as default upload provider
+// config values from configuration service are strings
+if (String(Nuxeo.UI.config.s3.useDirectUpload) == 'true') {
```
Gabez0r: use strict equals instead:
```suggestion
if (String(Nuxeo.UI.config.s3.useDirectUpload) === 'true') {
```

Gabez0r:
👍

addons/amazon-s3-online-storage/index.js
nelsonsilva: Since we're effectively "casting" to `String` there's no point in doing the strict equals, ie `String(1) === '1'` 🤷‍♂️

PR #545, opened 2019-03-29
NXP-25872: Stabilize Functional Tests

nmpcunha:


guirenard:
:+1:

package.json
```diff
@@ -41,6 +41,9 @@
     "postbuild": "npm-run-all fix:*",
     "fix:elements": "replace '/elements/' '${window.location.pathname}/' target/classes/web/nuxeo.war/ui/*.js",
     "fix:ui_elements": "replace '/node_modules/@nuxeo/nuxeo-ui-elements/' '${window.location.pathname}/' target/classes/web/nuxeo.war/ui/*.js",
+    "fix:viewers": "replace '\\${window.location.protocol}//?\\${window.location.host}\\${window.location.pathname}/(viewers|widgets)/[\\w\\d-]+.js' '${window.location.protocol}/${window.location.host}${window.location.pathname}/vendor/' target/classes/web/nuxeo.war/ui/vendors~elements.bundle.js",
```
nelsonsilva: protocol, host and path shouldn't be affected so we should be able to remove them from the regex

nelsonsilva: also, I'd rather go with *.js since number of bundles and their names may change

Gabez0r: > protocol, host and path shouldn't be affected so we should be able to remove them from the regex

Keep in mind that I'm still keeping them on the new string, I'm just using it to make the regex more specific and try to only cover the cases where with use importMeta. I think just blindly replacing `\/(viewers|widgets)\/[\w\d-]+\.js` with `/vendor/` as a great potential to break stuff in the future...

nelsonsilva:
👍

PR #552, opened 2019-04-15
NXP-26938: expose variables to configure the create and edit dialog sizes

semisse:
Pipeline successful: https://qa.nuxeo.org/jenkins/job/web-ui-pipeline-sfialho/12/

### DIFF
--- a/elements/document/nuxeo-document-form-layout.js
+++ b/elements/document/nuxeo-document-form-layout.js
@@ -42,9 +42,7 @@ Polymer({
 
       .scrollable {
         padding: 0 24px;
-        /* XXX: max-height is set as a fallback in case height is unset */
-        max-height: calc(var(--nuxeo-document-form-height, 80vh) - 5.9em);
-        height: calc(var(--nuxeo-document-form-height, unset) - 5.9em);
+        max-height: calc(var(--nuxeo-document-form-max-height, 60vh) - 6.3rem);
         @apply --layout-scroll;
       }
     </style>
--- a/elements/document/nuxeo-document-form-layout.js
+++ b/elements/document/nuxeo-document-form-layout.js
@@ -42,7 +42,9 @@ Polymer({
 
       .scrollable {
         padding: 0 24px;
-        max-height: 60vh;
+        /* XXX: max-height is set as a fallback in case height is unset */
+        max-height: calc(var(--nuxeo-document-form-height, 80vh) - 5.9em);
+        height: calc(var(--nuxeo-document-form-height, unset) - 5.9em);
         @apply --layout-scroll;
       }
     </style>
--- a/elements/nuxeo-document-actions/nuxeo-document-form-button.js
+++ b/elements/nuxeo-document-actions/nuxeo-document-form-button.js
@@ -41,7 +41,7 @@ Polymer({
       }
 
       nuxeo-dialog {
-        width: var(--nuxeo-document-form-width, 915px);
+        min-width: var(--nuxeo-document-form-min-width, 915px);
         margin: 0;
       }
     </style>
--- a/elements/nuxeo-document-actions/nuxeo-document-form-button.js
+++ b/elements/nuxeo-document-actions/nuxeo-document-form-button.js
@@ -18,14 +18,14 @@ import '@polymer/polymer/polymer-legacy.js';
 
 import '@polymer/paper-icon-button/paper-icon-button.js';
 import '@polymer/paper-dialog-scrollable/paper-dialog-scrollable.js';
+import { Polymer } from '@polymer/polymer/lib/legacy/polymer-fn.js';
+import { html } from '@polymer/polymer/lib/utils/html-tag.js';
 import '@nuxeo/nuxeo-ui-elements/actions/nuxeo-action-button-styles.js';
 import { I18nBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-i18n-behavior.js';
 import { FiltersBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-filters-behavior.js';
 import '@nuxeo/nuxeo-ui-elements/widgets/nuxeo-dialog.js';
 import '@nuxeo/nuxeo-ui-elements/widgets/nuxeo-tooltip.js';
 import '../document/nuxeo-document-form-layout.js';
-import { Polymer } from '@polymer/polymer/lib/legacy/polymer-fn.js';
-import { html } from '@polymer/polymer/lib/utils/html-tag.js';
 
 /**
 `nuxeo-document-form-button`
@@ -39,10 +39,10 @@ Polymer({
         margin: 0;
         padding: 24px 0 0 0;
       }
-      @media (min-width: 1024px) {
-        nuxeo-dialog {
-          min-width: 915px;
-        }
+
+      nuxeo-dialog {
+        width: var(--nuxeo-document-form-width, 915px);
+        margin: 0;
       }
     </style>
 
--- a/elements/nuxeo-document-create-popup/nuxeo-document-create-popup.js
+++ b/elements/nuxeo-document-create-popup/nuxeo-document-create-popup.js
@@ -42,6 +42,8 @@ Polymer({
       }
 
       nuxeo-dialog {
+        display: flex;
+        flex-direction: column;
         height: var(--nuxeo-document-create-popup-height, 80vh);
         width: var(--nuxeo-document-create-popup-width, 65vw);
         margin: 0;
@@ -68,8 +70,8 @@ Polymer({
       }
 
       #holder {
-        height: calc(100% - 48px);
-        width: 100%;
+        min-height: calc(100% - 48px);
+        min-width: 100%;
       }
 
       iron-pages {
--- a/elements/nuxeo-document-create-popup/nuxeo-document-create-popup.js
+++ b/elements/nuxeo-document-create-popup/nuxeo-document-create-popup.js
@@ -15,14 +15,14 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 import '@polymer/polymer/polymer-legacy.js';
-
 import '@polymer/iron-flex-layout/iron-flex-layout.js';
-import { I18nBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-i18n-behavior.js';
-import '@nuxeo/nuxeo-elements/nuxeo-document.js';
 import '@polymer/paper-dialog/paper-dialog.js';
 import '@polymer/paper-tabs/paper-tabs.js';
 import '@polymer/iron-pages/iron-pages.js';
+import { I18nBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-i18n-behavior.js';
+import '@nuxeo/nuxeo-elements/nuxeo-document.js';
 import '@nuxeo/nuxeo-ui-elements/nuxeo-slots.js';
+import '@nuxeo/nuxeo-ui-elements/widgets/nuxeo-dialog.js';
 import '../document/nuxeo-document-create.js';
 import '../document/nuxeo-document-import.js';
 import { Polymer } from '@polymer/polymer/lib/legacy/polymer-fn.js';
@@ -37,8 +37,14 @@ Polymer({
   _template: html`
     <style include="iron-flex">
       :host {
-        --paper-tabs-selection-bar-color: var(--nuxeo-primary-color);
         display: block;
+        --paper-tabs-selection-bar-color: var(--nuxeo-primary-color);
+      }
+
+      nuxeo-dialog {
+        height: var(--nuxeo-document-create-popup-height, 80vh);
+        width: var(--nuxeo-document-create-popup-width, 65vw);
+        margin: 0;
       }
 
       paper-tabs {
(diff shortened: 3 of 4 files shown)


guirenard:
:+1:

elements/nuxeo-document-actions/nuxeo-document-form-button.js
```diff
@@ -36,9 +36,24 @@ Polymer({
   _template: html`
     <style include="nuxeo-action-button-styles">
       .container {
+        width: var(--nuxeo-document-edit-popup-width);
```
Gabez0r: We should avoid using `edit` in the names of css variables on a generic element that can be used not only for edit layouts. Perhaps replacing "edit" with "form" at this level would be more adequate.

elements/document/nuxeo-document-form-layout.js
```diff
@@ -33,6 +33,10 @@ import { I18nBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-i18n-behavior.js';
 Polymer({
   _template: html`
     <style include="nuxeo-styles">
+      :host {
+        --nuxeo-document-edit-popup-height-scrollable: calc(var(--nuxeo-document-edit-popup-height) - 5.9em);
```
Gabez0r: Why define this variable here? It won't be overridable and since it's only used in one place, perhaps its value could be set inline. (se comment bellow)

elements/document/nuxeo-document-form-layout.js
```diff
@@ -42,7 +46,7 @@ Polymer({
 
       .scrollable {
         padding: 0 24px;
-        max-height: 60vh;
+        height: var(--nuxeo-document-edit-popup-height-scrollable, 60vh);
```
Gabez0r:
```suggestion
        height: calc(var(--nuxeo-document-form-popup-height, <default-here>) - 5.9em);
```

elements/nuxeo-document-actions/nuxeo-document-form-button.js
```diff
@@ -36,9 +36,24 @@ Polymer({
...
+        --paper-dialog: {
+          margin: 0;
+        }
+      }
+
+      @media (max-width: 1023px) {
+        nuxeo-dialog {
+          height: var(--nuxeo-document-edit-popup-height);
+          width: var(--nuxeo-document-edit-popup-width);
+        }
```
Gabez0r: These make the dialog jump and become really narrow when the screen size drops to 1023px, whereas it should take around 90% of the width instead. Also, keep in mind that we already have a `@media (min-width: 1024px) {` rule below this one, which perhaps doesn't much sense now?

themes/base.js
```diff
@@ -558,7 +564,7 @@ const template = html`
 
           --nuxeo-dialog: {
             min-width: 0;
-            width: 90%;
+            width: var(--nuxeo-document-create-popup-width, 90%);
```
Gabez0r: Since this rule affects all dialogs, we shouldn't be using a css variable specific to the create popup. Perhaps the 90% is still ok here? Keep in mind that this rule can be overridden on the themes.

elements/document/nuxeo-document-form-layout.js
semisse: I was trying to make it more readable, but I guess it turns out to be more complex. I will incorporate your suggestion.

themes/base.js
semisse: I will revert to 90%.

elements/document/nuxeo-document-form-layout.js
semisse: 👍

elements/nuxeo-document-actions/nuxeo-document-form-button.js
```diff
@@ -36,12 +36,21 @@ Polymer({
   _template: html`
     <style include="nuxeo-action-button-styles">
       .container {
+        width: var(--nuxeo-document-form-popup-width);
```
Gabez0r: This width should be applied to `nuxeo-dialog`. Otherwise, when you set it to 90%, the container will have 90% of the with of its parent, which is a `nuxeo-dialog`, leading to a vertical strip on the right, when screen width is <= 1024px.

elements/nuxeo-document-actions/nuxeo-document-form-button.js
```diff
@@ -36,12 +36,21 @@ Polymer({
...
       }
-      @media (min-width: 1024px) {
+
+      nuxeo-dialog {
+        --paper-dialog: {
+          margin: 0;
+        }
+      }
+
+      @media (max-width: 1023px) {
```
Gabez0r: I don't think we need any media query here since we're already setting the width on the base theme.

elements/nuxeo-document-create-popup/nuxeo-document-create-popup.js
```diff
@@ -62,8 +65,8 @@ Polymer({
       }
 
       #holder {
-        height: 80vh;
-        width: 65vw;
+        height: calc(var(--nuxeo-document-create-popup-height, 80vh) - 48px);
+        width: var(--nuxeo-document-create-popup-width, 65vh);
```
Gabez0r: This width should be applied to `nuxeo-dialog`, using the same rationale of the create pop.

Gabez0r:
```suggestion
        width: var(--nuxeo-document-create-popup-width, 65vw);
```

elements/nuxeo-document-create-popup/nuxeo-document-create-popup.js
```diff
@@ -62,8 +70,8 @@ Polymer({
       }
 
       #holder {
-        height: 80vh;
-        width: 65vw;
+        height: calc(var(--nuxeo-document-create-popup-height, 80vh) - 48px);
```
Gabez0r: Since we are already using `var(--nuxeo-document-create-popup-height, 80vh)` on this elements' parent, we should be able to just do:
```suggestion
        height: calc(100% - 48px);
```

Gabez0r:
:+1:

elements/nuxeo-document-create-popup/nuxeo-document-create-popup.js
```diff
@@ -37,8 +37,14 @@ Polymer({
   _template: html`
     <style include="iron-flex">
       :host {
-        --paper-tabs-selection-bar-color: var(--nuxeo-primary-color);
         display: block;
+        --paper-tabs-selection-bar-color: var(--nuxeo-primary-color);
+      }
+
+      nuxeo-dialog {
+        height: var(--nuxeo-document-create-popup-height, 80vh);
```
nelsonsilva: @Gabez0r can we go with dialog here too instead of popup ?

Gabez0r: @nelsonsilva it would not be consistent with the convention we've been using to name css variables, which is `--<element>-property`

elements/document/nuxeo-document-form-layout.js
```diff
@@ -42,7 +42,8 @@ Polymer({
 
       .scrollable {
         padding: 0 24px;
-        max-height: 60vh;
+        max-height: calc(var(--nuxeo-document-form-popup-height, 80vh) - 5.9em);
```
Gabez0r: Do we need this max-height here, given that we already have height defined down below?

semisse: Since we have to take into consideration that the height of this div might not be set, max-height is our fallback to the default expected size.

elements/nuxeo-document-create-popup/nuxeo-document-create-popup.js
nelsonsilva: Makes sense, but we agree to use `-dialog` in other elements, yes?

themes/base.js
```diff
@@ -375,6 +375,12 @@ const template = html`
 
         --nuxeo-document-blob-actions-menu-max-width: 160px;
 
+        --nuxeo-document-create-popup-height: unset;
```
nelsonsilva: Shouldn't we set these variables here?

semisse: No need since there is nothing overriding the height in the mediaquery.

elements/nuxeo-document-create-popup/nuxeo-document-create-popup.js
semisse: For future reference: 
- we decided to rename `--nuxeo-document-form-popup-*` to `--nuxeo-document-form-*`.
- we decided to keep `--nuxeo-document-create-popup-*` as it matches the name of the element.

elements/nuxeo-document-actions/nuxeo-document-form-button.js
```diff
@@ -39,10 +39,10 @@ Polymer({
         margin: 0;
         padding: 24px 0 0 0;
       }
-      @media (min-width: 1024px) {
-        nuxeo-dialog {
-          min-width: 915px;
-        }
+
+      nuxeo-dialog {
+        width: var(--nuxeo-document-form-popup-width, 915px);
```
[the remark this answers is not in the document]
semisse: For future reference:
- we decided to rename --nuxeo-document-form-popup-* to --nuxeo-document-form-*.
- we decided to keep --nuxeo-document-create-popup-* as it matches the name of the element.

elements/document/nuxeo-document-form-layout.js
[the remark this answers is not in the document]
semisse: For future reference:
- we decided to rename --nuxeo-document-form-popup-* to --nuxeo-document-form-*.
- we decided to keep --nuxeo-document-create-popup-* as it matches the name of the element.

guirenard:
:+1:

elements/document/nuxeo-document-form-layout.js
```diff
@@ -42,7 +42,7 @@ Polymer({
 
       .scrollable {
         padding: 0 24px;
-        max-height: 60vh;
+        max-height: calc(var(--nuxeo-document-form-max-height, 60vh) - 5.9em);
```
semisse: This value represents the height of the Actions bar. We are subtracting it to avoid being pushed out of screen.

nmpcunha: Is 5.9em the height taken by the `<div class="actions">`?
Why this value? Should we define its height too?

semisse: As discussed, I'm going to convert it to rem's.

themes/base.js
```diff
@@ -555,6 +561,8 @@ const template = html`
       @media (max-width: 1024px) {
         html {
           font-size: 14px;
+          --nuxeo-document-create-popup-width: 90%;
+          --nuxeo-document-form-width: 90%;
```
Gabez0r:
```suggestion
          --nuxeo-document-form-min-width: 90%;
```

elements/nuxeo-document-actions/nuxeo-document-form-button.js
```diff
@@ -39,9 +39,16 @@ Polymer({
...
         padding: 24px 0 0 0;
       }
-      @media (min-width: 1024px) {
+
+      nuxeo-dialog {
+        min-width: var(--nuxeo-document-form-min-width, 915px);
+        margin: 0;
+      }
+
+      @media (max-width: 1024px) {
```
Gabez0r: We need this media query since we are already defining it on the base theme.

elements/nuxeo-document-create-popup/nuxeo-document-create-popup.js
```diff
@@ -37,8 +37,16 @@ Polymer({
...
-        --paper-tabs-selection-bar-color: var(--nuxeo-primary-color);
         display: block;
+        --paper-tabs-selection-bar-color: var(--nuxeo-primary-color);
+      }
+
+      nuxeo-dialog {
+        display: flex;
+        flex-direction: column;
+        min-height: var(--nuxeo-document-create-popup-min-height, 80vh);
+        min-width: var(--nuxeo-document-create-popup-min-width, 65vw);
```
Gabez0r: I think we should stick with `width` and `height` here, since these prevent the dialog size changing abruptly when it's content changes.

semisse: done.

elements/nuxeo-document-actions/nuxeo-document-form-button.js
semisse: fixed

themes/base.js
semisse: fixed.

elements/nuxeo-document-actions/nuxeo-document-form-button.js
```diff
@@ -39,10 +39,10 @@ Polymer({
         margin: 0;
         padding: 24px 0 0 0;
       }
-      @media (min-width: 1024px) {
-        nuxeo-dialog {
-          min-width: 915px;
-        }
+
+      nuxeo-dialog {
+        min-width: var(--nuxeo-document-form-min-width, 915px);
```
nelsonsilva: This one only applies when the form is used in the dialog so perhaps it should be `nuxeo-document-form-popup-min-width` and we should probably introduce a `--nuxeo-document-form-popup-max-height`. This last could be used to  set `--nuxeo-document-form-max-height` here...  although I'd still expect it to be possible to set the `max-height` on the dialog itself and have a calc for the `max-height`  for the `.scrollable` in `nuxeo-document-form-layout` to do the ✨ and remove the `- 6.3rem`

semisse:
Since we've changed the approach to this problem, I'm closing this PR.
#596

PR #576, opened 2019-05-13
Fix nxp 27338 data table sorting

guirenard:


packages/nuxeo-web-ui-ftest/test/features/step_definitions/browser.js
```diff
@@ -78,6 +78,11 @@ Then('I can see the {string} child document is at position "{int}"', function (t
   this.ui.browser.indexOfChild(title).should.equals(pos - 1);
 });
 
+When('I order the content by title {string}', function (order) {
+  this.ui.browser.waitForVisible();
+  this.ui.browser.orderContent(0, order);
```
Gabez0r: Could we make so that this 0 is not hardcoded? Because if someone tries to use this step on another situation where there is a table with a title column that is not at index 0, it will fail. We could, for example, use the field name (.i.e, "dc:title") as a parameter o the step and then use the `field` property of `nuxeo-data-table-column` to find the column element on which to do the sorting, thus making this step reusable.

elements/nuxeo-results/nuxeo-document-content-behavior.js
```diff
@@ -123,7 +123,7 @@ export const DocumentContentBehavior = [
     },
 
     _displaySort(document, field) {
-      return this._canSort(document) ? field : undefined;
+      return this._canSort(document, this.sortOptions) ? field : undefined;
```
Gabez0r: The `sortOptions` is ignored by the data table. Also, the `_displaySort` method is used on the data table columns. So, I think the fix should be done at the _canSort level:

```
    _canSort(document, options) {
      return !(document && this.hasFacet(document, 'Orderable')) &&
              options ? Array.isArray(options) && options.length > 0 : true;
    },
```

This would preserve the original behavior of `_canSort` as it was before the refactoring, if no `options` is present.

packages/nuxeo-web-ui-ftest/test/features/step_definitions/browser.js
```diff
@@ -78,6 +78,11 @@ Then('I can see the {string} child document is at position "{int}"', function (t
   this.ui.browser.indexOfChild(title).should.equals(pos - 1);
 });
 
+When('I order the content by title {string}', function (order) {
```
nmpcunha: I can understand that it is not trivial to implement right now, but it would be interesting if we could choose the column instead of being always `title`.

nmpcunha:
Also, did you tried to order other fields by default? (e.g. "Modified")

packages/nuxeo-web-ui-ftest/test/features/step_definitions/browser.js
guirenard: well, I had that in mind in the first place but unfortunately, there's no name or whatsoever attribute propagated to the header cell sort elements. Unless significantly improve the data-table impl for only ftest purpose then anser is no.

packages/nuxeo-web-ui-ftest/test/features/step_definitions/browser.js
guirenard: see https://github.com/nuxeo/nuxeo-web-ui/pull/576#discussion_r283307382

guirenard:
> Also, did you tried to order other fields by default? (e.g. "Modified")

yes

packages/nuxeo-web-ui-ftest/test/features/step_definitions/browser.js
nmpcunha: Should we create a ticket for this?

packages/nuxeo-web-ui-ftest/test/pages/ui/browser.js
```diff
@@ -218,6 +218,34 @@ export default class Browser extends BasePage {
...
+      this.waitForChildren();
+      let columnIndex;
+      const columns = this.currentPage.elements('nuxeo-data-table nuxeo-data-table-column');
+      for (let i = 0; i < columns.value.length; i++) {
+        const colSortByField = columns.value[i].getAttribute('sortBy');
+        if (colSortByField && colSortByField.toLowerCase() === field.toLowerCase()) {
+          columnIndex = i;
+          break;
+        }
+      }
```
Gabez0r: nitpicking: we could rewrite this for with something smaller like:
```
const columns = this.currentPage.elements('nuxeo-data-table nuxeo-data-table-column');
const columnIndex = columns.value.map((col) => col.getAttribute('sortBy'))
    .findIndex((colSortByField) => colSortByField && olSortByField.toLowerCase() === field.toLowerCase()) 
```

packages/nuxeo-web-ui-ftest/test/pages/ui/browser.js
```diff
@@ -218,6 +218,34 @@ export default class Browser extends BasePage {
...
+      if (typeof columnIndex === 'undefined') {
+        throw new Error('Field not found');
+      }
+      const hc = this.currentPage.elements('nuxeo-data-table nuxeo-data-table-row[header] nuxeo-data-table-cell');
+      const c = hc.value[columnIndex];
+      const sortElt = c.element('nuxeo-data-table-column-sort');
```
nmpcunha: nit: we don't need all these variables, in fact. We could merge all on sortElt ;)

packages/nuxeo-web-ui-ftest/test/features/step_definitions/browser.js
```diff
@@ -78,6 +78,11 @@ Then('I can see the {string} child document is at position "{int}"', function (t
   this.ui.browser.indexOfChild(title).should.equals(pos - 1);
 });
 
+When('I sort the content by {string} {string}', function (field, order) {
```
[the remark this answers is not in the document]
nmpcunha: Indeed.

nmpcunha:
👍

Gabez0r:
👍

PR #585, opened 2019-05-16
NXP-26429: Improve error handling / permissions on CSV import

RSalem07:
this fix is related to https://jira.nuxeo.com/browse/NXP-26429, and mainly on https://jira.nuxeo.com/browse/NXP-26429?focusedCommentId=386657&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-386657

_Error message on the path suggester should not appear if user has write permission_ : to solve this issue i duplicate some code in `nuxeo-document-import-csv.js` from `nuxeo-document-create-popup.js` to be able to update the parent property and the whole validation is ensured by the behaviour `nuxeo-document-creation-behavior.js`.

But unfortunately we end up with a duplicated code, to solve the problem caused by the two way binding when we add an extension to exiting slot.

with @guirenard we worked on a 2nd solution (https://github.com/nuxeo/nuxeo-web-ui/tree/fix-NXP-26429-csvPopup) avoiding the duplicated code,  we simulate the two way binding by setting manually the properties   
`this.parentPath = e.detail.parentPath;`
 `this.suggesterChildren = e.detail.suggesterChildren;`

_Apply Date, Author and Dublin Core properties should not be available to non-admins
disable the option_ : waiting for https://jira.nuxeo.com/browse/ELEMENTS-946

### DIFF
--- a/addons/nuxeo-csv/elements/nuxeo-document-import-csv.js
+++ b/addons/nuxeo-csv/elements/nuxeo-document-import-csv.js
@@ -18,6 +18,7 @@ import { Polymer } from '@polymer/polymer/lib/legacy/polymer-fn.js';
 import { html } from '@polymer/polymer/lib/utils/html-tag.js';
 import { IronResizableBehavior } from '@polymer/iron-resizable-behavior';
 import { I18nBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-i18n-behavior.js';
+import { FiltersBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-filters-behavior.js';
 import { UploaderBehavior } from '@nuxeo/nuxeo-ui-elements/widgets/nuxeo-uploader-behavior.js';
 import { RoutingBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-routing-behavior.js';
 import { DocumentCreationBehavior } from '../../../elements/nuxeo-document-creation/nuxeo-document-creation-behavior.js';
@@ -173,7 +174,9 @@ Polymer({
       }
 
       .error {
-        color: var(--nuxeo-warn-text);
+        border-left: 4px solid var(--nuxeo-warn-text);
+        color: var(--primary-text-color);
+        padding-left: 8px;
       }
 
       .importing-label {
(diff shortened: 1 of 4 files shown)


addons/nuxeo-csv/elements/nuxeo-document-import-csv.js
```diff
@@ -244,18 +245,20 @@ Polymer({
       }
     </style>
 
-    <nuxeo-connection id="nx"></nuxeo-connection>
+    <nuxeo-connection id="nx" user="{{currentUser}}"></nuxeo-connection>
```
guirenard: Not really sure we need this ...

RSalem07: I added o retrieve the `currentUser` and  be able to passed it to a new method that allows the check of admin permissions (it will be added https://jira.nuxeo.com/browse/ELEMENTS-946)

RSalem07:
T&P Ok on web ui pipeline : https://qa.nuxeo.org/jenkins/job/Master/job/web-ui-pipeline/196/

nmpcunha:
Does it make sense to squash the commits? Despite that, seems good!

RSalem07:
@nmpcunha  I separated these two commits, because the 2 nd it's not really related on this ticket, i just did the refactoring to reuse the new method `hasAdministratorPermissions`. i think having  the 2nd commit let us to see more quickly what was changed. but if you think that one commit is needed to be consistent with the what you did in the front end development  i will do it :)

addons/nuxeo-csv/elements/nuxeo-document-import-csv.js
```diff
@@ -264,10 +267,10 @@ Polymer({
             disabled
             always-float-label
           ></nuxeo-path-suggestion>
-          <span class="error">[[targetLocationError]]</span>
+          <span class$="horizontal layout [[_formatErrorMessage(errorMessage)]]">[[errorMessage]]</span>
         </div>
 
-        <div id="dropzone" class="vertical layout flex step">
+        <div id="dropzone" class="vertical layout flex step" hidden="[[!_canUpload(canCreate)]]">
```
guirenard: can't you can directly use `[[!canCreate]]` ?

addons/nuxeo-csv/elements/nuxeo-document-import-csv.js
```diff
@@ -315,14 +318,16 @@ Polymer({
           </template>
         </div>
 
-        <div class="options">
+        <div class="options" hidden="[[!_canUpload(canCreate)]]">
```
guirenard: same same

addons/nuxeo-csv/elements/nuxeo-document-import-csv.js
```diff
@@ -703,4 +690,8 @@ Polymer({
       clearTimeout(this._waitProgressId);
     }
   },
+
+  _canUpload() {
```
guirenard: and this would become useless..

PR #588, opened 2019-05-17
NXP-27264: allow configuring packages, server url and proxy host

nelsonsilva:


webpack.config.js
```diff
@@ -136,14 +138,22 @@ const common = merge([
         THREE: 'three',
         jQuery: 'jquery',
       }),
+      new HtmlWebpackPlugin({
+        title: 'Nuxeo',
+        template: 'index.html',
+        nuxeo: {
+          packages: JSON.stringify((process.env.NUXEO_PACKAGES || '').split(' ').filter((_) => !!_)),
```
Gabez0r:
```suggestion
          packages: JSON.stringify((process.env.NUXEO_PACKAGES || '').split(' ').filter(Boolean)),
```

webpack.config.js
```diff
@@ -152,7 +162,7 @@ const development = merge([
       host: '0.0.0.0',
       historyApiFallback: true,
       proxy: {
-        '/nuxeo': 'http://localhost:8080/',
+        '/nuxeo': `http://${process.env.NUXEO_HOST || 'localhost'}/`,
```
Gabez0r: Missing port here. If we assume `NUXEO_HOST` also includes the port, we could do something like:
```suggestion
        '/nuxeo': `http://${process.env.NUXEO_HOST || 'localhost:8080'}/`,
```

webpack.config.js
Gabez0r: Perhaps it would be better to split this by commas, so we don't have to specify the list wrapped in quotes?

guirenard: comma sounds more natural and consistent to me too

Gabez0r: After checking with Nelson, it seems that the space-separated list was used here to be compatible with what's being already set for the docker images, so we could re-use it. I'd suggest splitting by both space and comma, so it could serve both purposes, something like:
```suggestion
          packages: JSON.stringify((process.env.NUXEO_PACKAGES || '').split(/[\s,]+/).filter(Boolean)),
```

nelsonsilva: I agree with you. I kept spaces for compat with `NUXEO_PACKAGES` used by our official Docker image so you can just reuse the same env var for both server and ui.

nelsonsilva: 👍

Gabez0r: So, no splitting by both spaces and commas?

webpack.config.js
```diff
@@ -136,14 +138,22 @@ const common = merge([
         THREE: 'three',
         jQuery: 'jquery',
       }),
+      new HtmlWebpackPlugin({
+        title: 'Nuxeo',
+        template: 'index.html',
+        nuxeo: {
+          packages: JSON.stringify((process.env.NUXEO_PACKAGES || '').split(' ').filter(Boolean),
```
Gabez0r: So, no split by both space and comma?

PR #591, opened 2019-05-20
NXP-27063: remove param filtering logic from search form

nmpcunha:
Elements' ticket: https://github.com/nuxeo/nuxeo-elements/pull/92
Pipeline: https://qa.nuxeo.org/jenkins/view/TestAndPush/job/web-ui-pipeline-ncunha/165/


Gabez0r:
I'd say that a better commit message would be "remove param filtering logic from search form" (or something similar), since all we are doing here is removing code.

PR #594, opened 2019-05-24
Fix nxp 27307 collection escape html

guirenard:


guirenard:
pipeline in progress

packages/nuxeo-web-ui-ftest/test/pages/ui/browser.js
```diff
@@ -105,19 +106,14 @@ export default class Browser extends BasePage {
...
-    selectivity.waitForVisible('.selectivity-result-item.highlight');
-    selectivity.click('.selectivity-result-item.highlight');
-    driver.waitForEnabled('#add-to-collection-dialog paper-button[name="add"]');
-    driver.click('#add-to-collection-dialog paper-button[name="add"]');
+    if (!this.el.isExisting('#add-to-collection-dialog') || !this.el.isVisible('#add-to-collection-dialog')) {
+      const button = this.el.element('nuxeo-add-to-collection-button paper-icon-button');
+      button.waitForVisible();
+      button.click();
+    }
+    const addToCollectionDialog = new CollectionDialog(`${this._selector} #add-to-collection-dialog`);
```
Gabez0r: I think it would be safer to name the `CollectionDialog` helper something more specific, like `AddToCollectionDialog`, since it wraps dialogs used only to create collections and it could collide in the future with other dialogs related to Collections.

packages/nuxeo-web-ui-ftest/test/pages/ui/browser/collection_dialog.js
```diff
@@ -0,0 +1,14 @@
+import BasePage from '../../base';
+
+export default class CollectionDialog extends BasePage {
+  addToCollection(collectionName) {
+    const collectionSelect = this.el.element('#nxSelect');
+    fixtures.layouts.setValue(collectionSelect, collectionName);
```
Gabez0r: Perhaps it would be safer to wait for collectionSelect to be visible?

packages/nuxeo-web-ui-ftest/test/features/step_definitions/support/fixtures/collections.js
```diff
@@ -1,17 +1,18 @@
-import nuxeo from '../services/client';
+import nuxeo, { SYNC_OPTIONS } from '../services/client';
```
Gabez0r: Are there any cases where we don't want this as default? If not, I'd say we should change on `../services/client` and try to do something like:
```
export default new Nuxeo({
  auth: { method: 'basic', username: 'Administrator', password: 'Administrator' },
  baseURL: process.env.NUXEO_URL,
  headers: { nx_es_sync: 'true' },
});
```

PR #601, opened 2019-05-31
NXP-27448: ensure form element is autofocused when dialog is opened

nmpcunha:


elements/document/nuxeo-document-layout.js
```diff
@@ -104,6 +111,24 @@ Polymer({
...
         layout: this.layout,
       });
+      this.applyAutoFocus();
     });
   },
+
+  _getFocusableElement(parent) {
+    if (parent && parent.shadowRoot && !parent.shadowRoot.activeElement) {
+      const nodes = Array.from(parent.shadowRoot.querySelectorAll('*')).filter((node) => node.shadowRoot);
+      let focusableElement = nodes.find((node) => !node.disabled && node.autofocus);
```
guirenard: What about `hidden` or `display:none` focusable element?

nmpcunha: Will add more checks, but I'm not sure if they will filter that much.
```
node.shadowRoot && !node.hidden && node.style.display !== 'none' && node.style.visibility !== 'none'
```

elements/document/nuxeo-document-layout.js
```diff
@@ -104,6 +111,24 @@ Polymer({
         element: this.element,
         layout: this.layout,
       });
+      this.applyAutoFocus();
     });
   },
+
+  _getFocusableElement(parent) {
+    if (parent && parent.shadowRoot && !parent.shadowRoot.activeElement) {
+      const nodes = Array.from(parent.shadowRoot.querySelectorAll('*')).filter((node) => node.shadowRoot);
```
Gabez0r: Don't think we should filter for shadowRoot here, since we'd be excluding native elements.

elements/document/nuxeo-document-layout.js
```diff
@@ -104,6 +111,24 @@ Polymer({
...
+      let focusableElement = nodes.find((node) => !node.disabled && node.autofocus);
+      if (focusableElement) {
+        return focusableElement;
+      }
+
+      for (let i = 0; i < nodes.length; i++) {
+        focusableElement = this._getFocusableElement(nodes[i]);
```
Gabez0r: But we should only run recursively if we have a shadowRoot.

elements/nuxeo-document-actions/nuxeo-document-form-button.js
```diff
@@ -126,4 +134,9 @@ Polymer({
   _closeDialog() {
     this.dialog.close();
   },
+
+  _formLayoutOpened() {
+    const { layout } = this.$.layout.$;
+    layout.applyAutoFocus(layout.element);
```
Gabez0r: I don't think `applyAutoFocus ` accepts a parameter?

elements/document/nuxeo-document-layout.js
nmpcunha: Nice point, thanks!

elements/nuxeo-document-actions/nuxeo-document-form-button.js
nmpcunha: Bad rebase and outdated code, sorry!

elements/nuxeo-document-actions/nuxeo-document-form-button.js
```diff
@@ -126,4 +134,9 @@ Polymer({
   _closeDialog() {
     this.dialog.close();
   },
+
+  _formLayoutOpened() {
```
Gabez0r: We just need to be careful with this logic being triggered by dialogs inside layouts.

elements/document/nuxeo-document-layout.js
```diff
@@ -104,6 +111,27 @@ Polymer({
...
+      this.applyAutoFocus();
     });
   },
+
+  _getFocusableElement(parent) {
+    if (parent && parent.shadowRoot && !parent.shadowRoot.activeElement) {
+      const nodes = Array.from(parent.shadowRoot.querySelectorAll('*')).filter(
+        (node) => !node.hidden && node.style.display !== 'none' && node.style.visibility !== 'none',
+      );
+      let focusableElement = nodes.find((node) => !node.disabled && node.autofocus);
```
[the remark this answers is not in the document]
nmpcunha: In theory, it would already a subset that would also be used for recursion.
I did some changes to try to filter in a more accurate way, but please let me know if you think we should tackle this in a different way.

PR #603, opened 2019-05-31
NXP-27307: add name attribute for test purpose

guirenard:


guirenard:
https://qa2.nuxeo.org/jenkins/job/WebUI/job/web-ui-pipeline-9.10/214/ OK

PR #611, opened 2019-06-06
NXP-26938: add css variables to nuxeo dialog (Backport 9.10)

semisse:


semisse:
We won't be adding this improvement to 9.10.

semisse:
We won't be adding this improvement to 9.10.

PR #612, opened 2019-06-07
NXP-27530: breadcrumb pushes actions menu

semisse:


elements/nuxeo-browser/nuxeo-breadcrumb.js
```diff
@@ -182,6 +192,13 @@ Polymer({
     return '';
   },
 
+  _ellipsisDirection() {
+    if (document.getElementsByTagName('html')[0].getAttribute('dir') !== 'rtl') {
```
nmpcunha: Can we use this, instead? https://developer.mozilla.org/en-US/docs/Web/API/Document/dir
And if needed, use that approach as fallback?

elements/nuxeo-browser/nuxeo-breadcrumb.js
```diff
@@ -118,7 +131,7 @@ Polymer({
       <div class="current-icon">
         <iron-icon src="[[_icon(document, url)]]"></iron-icon>
       </div>
-      <div class="doc-path">
+      <div class$="doc-path [[_ellipsisDirection()]]">
```
guirenard: Wouldn't it make more sense to have a `Function`  prop called `ellipsisDirection` that would give more flexibility in term of reusability?

Gabez0r: We have no one requesting for this feature and we have no specifications on how this would be reused, so I'd be ok with keeping this private for now.

nmpcunha: @guirenard, what do you have in mind? Not sure if I followed your idea...
This function is only returning a predefined CSS class. If we expose it as a property, we would need to give more flexibility (exposing also CSS vars, or so), otherwise a user won't be able to do that much... correct?

guirenard: Ok forget it then :)

PR #623, opened 2019-06-12
NXP-27451: clipboard validating destination subtypes

nmpcunha:


elements/nuxeo-clipboard/nuxeo-clipboard.js
```diff
@@ -255,7 +255,12 @@ Polymer({
   },
 
   canPaste(documents, doc) {
-    return documents && documents.length && this.hasFacet(doc, 'Folderish');
+    return (
+      documents &&
+      documents.length &&
+      this.hasFacet(doc, 'Folderish') &&
+      documents.every((entry) => doc.contextParameters.subtypes.map((e) => e.type).indexOf(entry.type) > -1)
```
Gabez0r: We should only do this logic if the document actually has a `contextParameters ` and `contextParameters.subtypes`.

PR #625, opened 2019-06-12
NXP-27073: Ensure favorite exists before fetching documents

akervern:


elements/nuxeo-home.html
```diff
@@ -211,6 +211,10 @@
         this.$.latestEditedDocs.fetch();
         this.$.recentStorage.reload();
         this._fetchFavorite().then((favorite) => {
+          if (!(favorite && favorite.uid)) {
```
semisse: Nitpick: Wondering if this alternative wouldn't be more readable:
`if (!favorite || !favorite.uid)`
WDYT? :)

akervern: 🤷🏻‍♂️
It's all about 🧠and habits... I'm more comfortable with my version, but I can understand yours could be easier somehow :)

PR #628, opened 2019-06-13
NXP-27543: Diff displays only id for vocabulary fields

RSalem07:


RSalem07:
The jenkins pipline [see](https://qa.nuxeo.org/jenkins/job/master/job/web-ui-pipeline/240/console) seems to be good, but some test failed i think mainly related to the knonws problem of Document / Note creation

elements/diff/elements/nuxeo-user-diff.html
```diff
@@ -84,12 +84,12 @@
       <template is="dom-if" if="[[_hasObjectInnerChanges(delta)]]">
         <template is="dom-if" if="[[!hideDeletions]]">
           <div class="deletion">
-            <nuxeo-user-tag user="[[delta.id.0]]"></nuxeo-user-tag>
+            <nuxeo-user-tag user="[[originalValue]]"></nuxeo-user-tag>
```
nmpcunha: Thanks for fixing this too.
I would say that it would be more intuitive to:
1. Separate in two different commits;
2. Change the commit message to don't be specific to directories.

Probably I would take the second one. ;)

RSalem07: you welecome, yes i agree to seperate these two commit, i completely forget about doing it

PR #629, opened 2019-06-13
ELEMENTS-949: avoid scollbars display on document actions with Edge - 9.10 only

vpasquier:


vpasquier:
_I will create a pipeline on it_

PR #631, opened 2019-06-18
NXP-27597: avoid scollbars display on document actions with Firefox - 9.10 only

vpasquier:


guirenard:
It seems more a paper elements / Polymer 2 issues on Windows/Firefox ESR 60 combinations. The paper-tooltip with a `bottom` position is actually not displayed right at the bottom but a bit shifted to the right.

I am definitely, not sure we should fix it by hiding the paper element issue with some padding trick.

Considering this issue is not reproducible on 10.10/2.4.x which is still on Polymer 2 stack but with upgraded minor elements, we could assume it has been fixed. The work to be done would be to find out where...

Issuing 
```m3 fox@fox-laptop:~/10.10-workspace/nuxeo-web-ui$ git diff 9.10:bower.json 10.10:bower.json```
tells me that:
 - 10.10 -> "paper-tooltip": "2.1.1"
 - 9.10 - > "paper-tooltip": "2.0.1"
Perhaps a an upgrade to  "paper-tooltip": "2.1.1" on 9.10 branch would fix the issue?

**Comment by guirenard** [comment]
BTW, you commit message is talking about Edge while the issue is about Firefox...

**Comment by vpasquier** [comment]
FYI the increase of the version didn't fix the issue. I'm investigating on the @Gabez0r point.

**Comment by Gabez0r** [comment]
It seems to me that there might be a sync problem here, in the sense that the tooltip might be displayed earlier than it should. If this is the case, than we should be able to do something at the `nuxeo-tooltip` level. I'd try to wrap the [show call on `nuxeo-tooptip`](https://github.com/nuxeo/nuxeo-elements/blob/master/ui/widgets/nuxeo-tooltip.js#L127) in an async call:
```
this.async(function() {
  this._tooltip.show();
}.bind(this));
```
If this pans out to fix the issue, than I'd say it should land on 10.10 and master as well.

vpasquier:
So @Gabez0r had the good one! Thank you! https://github.com/nuxeo/nuxeo-ui-elements/pull/424

PR #632, opened 2019-06-19
NXP-27317: fix path to mp to be archived

guirenard:


guirenard:
@Gabez0r waaaazaaaaaaa

PR #641, opened 2019-06-25
NXP-27037: ellipses and tooltip on document type create form

semisse:
This branch has 3 commits but it is the second one that I want to discuss with you.

In [NXP-27037](https://jira.nuxeo.com/browse/NXP-27037) we are asked to maintain two lines with ellipsis. I'm proposing `-webkit-line-clamp` to solve this question. The only place where it isn't supported now is Firefox 67, but it is already working on Firefox Nightly [(check it out here)](https://caniuse.com/#search=webkit-line-clamp). 

Today, the expected behaviour on Firefox 67 is simply the omission of the rest of the text. Do you think this could be a good solution?

### DIFF
--- a/elements/document/nuxeo-document-create.js
+++ b/elements/document/nuxeo-document-create.js
@@ -27,6 +27,7 @@ import '@polymer/paper-spinner/paper-spinner-lite.js';
 import '@polymer/iron-icon/iron-icon.js';
 import '@polymer/iron-a11y-keys/iron-a11y-keys.js';
 import '@nuxeo/nuxeo-ui-elements/nuxeo-path-suggestion/nuxeo-path-suggestion.js';
+import '@nuxeo/nuxeo-ui-elements/widgets/nuxeo-tooltip';
 import '@nuxeo/nuxeo-elements/nuxeo-document.js';
 import '../nuxeo-document-creation-stats/nuxeo-document-creation-stats.js';
 import './nuxeo-document-layout.js';
@@ -82,9 +83,31 @@ Polymer({
         box-shadow: 0 0 10px rgba(0, 0, 0, 0.3), 0 -3px 0 var(--nuxeo-link-hover-color) inset;
       }
 
-      .typeSelection iron-icon {
+      .typeIconWrapper {
+        display: flex;
+        flex: 1.5;
+        align-items: flex-end;
+      }
+
+      .typeIcon {
         width: var(--nuxeo-document-creation-form-icon-width, 42px);
         height: var(--nuxeo-document-creation-form-icon-height, 42px);
+      }
+
+      .typeLabelWrapper {
+        display: flex;
+        flex: 1;
+        align-items: flex-start;
+      }
+
+      .typeLabel {
+        width: 100%;
+        margin-top: 1em;
+        overflow: hidden;
+        text-overflow: ellipsis;
+        white-space: nowrap;
+        word-break: break-word;
+        line-height: normal;
       }
 
       .container {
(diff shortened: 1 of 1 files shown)


elements/document/nuxeo-document-create.js
```diff
@@ -82,11 +83,36 @@ Polymer({
...
+        display: flex;
+        flex: 1;
+        align-items: flex-start;
+        overflow: hidden;
+      }
+
+      .typeLabel {
+        width: 100%;
+        margin-top: 1em;
+        display: -webkit-box;
```
Gabez0r: We should avoid this, since this is non-standard. Also, it seems to be from an old spec superseded by flex, (https://www.w3.org/TR/2009/WD-css3-flexbox-20090723/).

elements/document/nuxeo-document-create.js
```diff
@@ -82,11 +83,36 @@ Polymer({
...
+        -webkit-box-orient: vertical;
```
Gabez0r: Same as above.

semisse: Yes, but unfortunately `-webkit-box-orient` does not work with new flex specifications.

Gabez0r: Yes, it belongs to the old spec, which was replaced by flexbox. This would be the equivalent to flex-direction, if I'm not mistaken.

semisse: That is correct, but in order for `-webkit-line-clamp` to work it requires both `display: -webkit-box` and `-webkit-box-orient: vertical` declared like the old spec:
https://codepen.io/kosei27/pen/gamywo

There are two alternatives here: 
1. [one css hack that requires an after with a gradient](https://codepen.io/nilsynils/pen/KgaRBo?editors=1100) (which doesn't serve our use cases, since the button background color might be changed);
2. Javascript.

I will commit the Javascript solution but, at first sight, it might be a little bit overkill for the result we are trying to achieve.

elements/document/nuxeo-document-create.js
```diff
@@ -82,11 +86,39 @@ Polymer({
         box-shadow: 0 0 10px rgba(0, 0, 0, 0.3), 0 -3px 0 var(--nuxeo-link-hover-color) inset;
       }
 
-      .typeSelection iron-icon {
+      .typeIconWrapper {
+        display: flex;
+        flex: 1.5;
+        align-items: flex-end;
+        @apply --nuxeo-document-create-selection-icon-wrapper;
```
nelsonsilva: Not a fan of exposing wrapper mixins.. this should be an implementation detail (overall not a fan of wrappers in general just for styling purposes) . Also we'd ideally start moving away from mixins and exposing well defined CSS variables.

Gabez0r: Agree.

elements/document/nuxeo-document-create.js
nmpcunha: Yes, sorry for getting back to this but we faced the same issue on comments.
We changed the approach which simplified this specific question (i.e. requirements changed and we used JS), but it would be interesting to think further on how we can do this kind of stuff since it will be needed in several places.

Gabez0r: Meanwhile, the requirement for line clamping was dropped in this ticket as well. Indeed, the solution to this problem needs to be with javascript afaik, since there is no equivalente using standard css.

Gabez0r: We could create a ticket to try to come up with a reusable solution for this problem.

PR #642, opened 2019-06-25
NXP-27113 simplified dataset export page

andreinechaev:


.gitignore
```diff
@@ -9,4 +9,7 @@ npm-debug.log
 .DS_Store
 .screenshots/
 *.iml
-.idea
\ No newline at end of file
+.idea
+.java-version
+.vscode
+package-lock.json
```
guirenard: We may want to commit a `package-lock.json` when releasing to be able to have reproducible builds. You can use the `--no-package-lock` option when running `npm install`

addons/nuxeo-ai-core/elements/admin-ai-advanced-export.js
```diff
@@ -0,0 +1,170 @@
...
+      <paper-textarea
+        id="queryInput"
+        type="search"
+        label="[[i18n('admin.ai.query')]]"
+        value="{{query}}"
+        placeholder="[[i18n('imaging.query.placeholder')]]"
+        autofocus
+      >
+      </paper-textarea>
+      <paper-input id="inputs" label="[[i18n('admin.ai.inputs')]]" value="{{inProps}}"></paper-input>
```
guirenard: we have `nuxeo-input` elements (we usually use for style consistency)

addons/nuxeo-ai-core/elements/nuxeo-ai-export-progress.js
```diff
@@ -0,0 +1,237 @@
...
+
+        .side {
+          @apply --layout-vertical;
+          @apply --layout-flex-2;
+          margin-bottom: var(--nuxeo-card-margin-bottom, 16px);
+          min-height: 60vh;
+        }
+
+        paper-progress {
+          --paper-progress-active - color: var(--paper-blue-600);
```
guirenard: typo?

addons/nuxeo-ai-core/elements/admin-ai-advanced-export.js
andreinechaev: this file was actually moved from the original `admin-ai-export.js` on the master branch.

.gitignore
andreinechaev: done.

andreinechaev:
sorry, I messed up with imports and i18n messages. One more time, please

addons/nuxeo-ai-core/elements/admin-ai-advanced-export.js
```diff
@@ -0,0 +1,170 @@
...
+limitations under the License.
+*/
+import { Polymer } from '@polymer/polymer/lib/legacy/polymer-fn.js';
+import { html } from '@polymer/polymer/lib/utils/html-tag.js';
+/** 
+`nuxeo-ai-admin`
+@group Nuxeo UI
+@element admin-ai-advanced-export
+*/
+Polymer({
```
Gabez0r: Please use class based syntax to created new elements. The polymer factory is legacy and we're only using it on layouts an in code that was migrated from Polymer 2.

addons/nuxeo-ai-core/elements/admin-ai-advanced-export.js
```diff
@@ -0,0 +1,170 @@
...
+      type: String,
+    },
+  },
+  _displayVal(val) {
+    if (val.numericValue != null) {
+      return val.numericValue;
+    }
+    if (val.type === 'terms') {
+      let toReturn = '';
+      for (let i = 0, len = val.value.length; i < len; i++) {
```
Gabez0r: nitpicking: would probably be easier to read with a `forEach`

addons/nuxeo-ai-core/elements/admin-ai-advanced-export.js
```diff
@@ -0,0 +1,170 @@
...
+  _recompute() {
+    this.$.aiStats.params = {
+      query: this.query,
+      inputs: this.inProps,
+      outputs: this.outProps,
+    };
+    this.$.aiStats
+      .execute()
+      .then((response) => {
+        if (response === undefined || response.length === 0) {
```
Gabez0r: Is `null` a valid response or a special case? If not, this would be shorter:
```suggestion
        if (!response || response.length === 0) {
```

addons/nuxeo-ai-core/elements/admin-ai-export.js
```diff
@@ -38,142 +62,137 @@ Polymer({
...
     },
   },
 
-  _displayVal(val) {
-    if (val.numericValue != null) {
-      return val.numericValue;
+  _visibleChanged(newVal) {
+    if (newVal) {
+      this.$.aiExportStatus.execute();
+    } else if (!newVal && clearTimeout !== undefined) {
```
Gabez0r: Don't think you need to check for `clearTimeout` to be defined, since it is supported by all browser supported by Web UI.

addons/nuxeo-ai-core/elements/admin-ai-export.js
```diff
@@ -38,142 +62,137 @@ Polymer({
...
-      .then((response) => {
-        if (response === undefined || response.length === 0) {
-          this.fire('notify', { message: this.i18n('admin.ai.stats.none') });
-        }
-      })
-      .catch(() => {
-        this.fire('notify', { message: this.i18n('admin.ai.stats.error') });
-      });
+  _updateItem(e) {
+    this.$.currentDoc = e.detail.value;
```
Gabez0r: Why do we need this method? Isn't setting the same value that is already bound to `fetchedDoc`?

addons/nuxeo-ai-core/elements/admin-ai-export.js
```diff
@@ -38,142 +62,137 @@ Polymer({
...
-    inProps: {
+    typeName: {
       type: String,
-      value: 'dc:title,file:content',
+      value: '',
     },
-    outProps: {
-      type: String,
-      value: 'dc:lastContributor',
+    fetchedDoc: {
```
Gabez0r: What is this document being used for? What's the different between this one and `currentDoc`?

addons/nuxeo-ai-core/elements/admin-ai-export.js
```diff
@@ -38,142 +62,137 @@ Polymer({
...
+        return response.value;
       })
       .catch(() => {
-        this._exportText = '';
         this.fire('notify', { message: this.i18n('admin.ai.export.error') });
       });
   },
+
+  _canExport() {
+    return this.typeName.length !== 0 && this.$.currentDoc !== undefined;
```
Gabez0r: Can we export if `this.$.currentDoc` is `null`? If not:
```suggestion
    return this.typeName.length !== 0 && this.$.currentDoc != undefined;
```

addons/nuxeo-ai-core/elements/nuxeo-ai-export-progress.js
```diff
@@ -0,0 +1,252 @@
...
+        <p class="model_title">[[_modelName]]</p>
+        <div class="page">
+          <div class="widget main">
+            <paper-progress
+              class="progress_bar"
+              value="[[_progress]]"
+              secondary-progress="[[_errorProgress]]"
+            ></paper-progress>
+            <p class="progress_counter">
+              [[_processed]]/[[_total]] documents <i style="opacity: 0.5;">[[_errorCount]] errors</i>
```
Gabez0r: If possible, avoid inline styles.

i18n/messages.json
```diff
@@ -53,6 +53,8 @@
   "admin.ai.stats.field" : "Property Name",
   "admin.ai.stats.type" : "Statistic Type",
   "admin.ai.stats.value" : "Value",
+  "admin.ai.model.selection":"Model selection",
```
Gabez0r:
```suggestion
  "admin.ai.model.selection": "Model selection",
```

i18n/messages.json
```diff
@@ -53,6 +53,8 @@
...
+  "admin.ai.model.listing":"Click here to view model list",
```
Gabez0r:
```suggestion
  "admin.ai.model.listing": "Click here to view model list",
```

addons/nuxeo-ai-core/elements/admin-ai-advanced-export.js
andreinechaev: advanced export is migrated from Polymer 2 for nuxeo-ai of 10.10 ver.

addons/nuxeo-ai-core/elements/admin-ai-export.js
andreinechaev: removed fetched doc. `value` is not updated as expected. the listener although is constantly updated with a required value

addons/nuxeo-ai-core/elements/admin-ai-export.js
andreinechaev: we cannot. I don't see how it can be null. undefined - yes

addons/nuxeo-ai-core/elements/admin-ai-advanced-export.js
Gabez0r: Sorry my comment was rather confusing. We only use polymer factory on elements that were migrated from [Polymer 2 in hybrid](https://polymer-library.polymer-project.org/2.0/docs/devguide/hybrid-elements). [Here's an example](https://github.com/nuxeo/nuxeo-ai/blob/2.1_10.10/nuxeo-ai-web-ui/src/main/resources/web/nuxeo.war/ui/nuxeo-ai/nuxeo-ai-suggestions.html).

addons/nuxeo-ai-core/elements/admin-ai-export.js
Gabez0r: Could you please open a bug ticket to fix this issue, if possible, with reproduction steps?

addons/nuxeo-ai-core/elements/admin-ai-export.js
Gabez0r: Since this is exposed as a public property, someone from outside can simply set it to `null`. Consider doing the check for `!= undefined` instead or setting it as a [read-only property](https://polymer-library.polymer-project.org/2.0/docs/devguide/properties#read-only).

addons/nuxeo-ai-core/elements/admin-ai-export.js
andreinechaev: I wouldn't call it a bug. it's rather a specific case of AI.

addons/nuxeo-ai-core/elements/admin-ai-advanced-export.js
andreinechaev: no, your comment is clear. I'm just saying that this is a migrated file. the original is [here] (https://github.com/nuxeo/nuxeo-ai/blob/2.1_10.10/nuxeo-ai-web-ui/src/main/resources/web/nuxeo.war/ui/nuxeo-ai/admin/admin-ai-export.html)

addons/nuxeo-ai-core/elements/admin-ai-export.js
Gabez0r: Not sure if I understand why we are using an event listener and not binding to value here. Could you please clarify what's being set on `value` and why it is not the *expected* value?

addons/nuxeo-ai-core/elements/admin-ai-advanced-export.js
Gabez0r: The element you linked is already migrated in this PR under:
[nuxeo-ai-core/elements/admin-ai-export.js](https://github.com/nuxeo/nuxeo-web-ui/blob/de1292759092c877a58a22c28bb8a93a44303a5e/addons/nuxeo-ai-core/elements/admin-ai-export.js).

So, [nuxeo-ai-core/elements/admin-ai-advanced-export.js](https://github.com/nuxeo/nuxeo-web-ui/blob/de1292759092c877a58a22c28bb8a93a44303a5e/addons/nuxeo-ai-core/elements/admin-ai-advanced-export.js) is indeed new element.

addons/nuxeo-ai-core/elements/admin-ai-export.js
andreinechaev: our operation returns a `JSONBlob` due to limitation of the platform. `JSONBlob` represent documents that do not exist in the current instance. As far as I understood they are simply not resolvable so the value is always null|undefined.

addons/nuxeo-ai-core/elements/admin-ai-advanced-export.js
Gabez0r: Ah, you meant https://github.com/nuxeo/nuxeo-ai/blob/2.1_10.10/nuxeo-ai-web-ui/src/main/resources/web/nuxeo.war/ui/nuxeo-ai/admin/admin-ai-advanced-export.html! I'm ok with this, then. But the "avoid using the Polymer factory" also applies for 10.10, but I didn't get the chance to review the other PR.

addons/nuxeo-ai-core/elements/admin-ai-export.js
Gabez0r: Thank you for the clarification.

addons/nuxeo-ai-core/elements/admin-ai-export.js
andreinechaev: `!=` eslint won't allow it. added explicit check

Gabez0r: It's the equivalent to != null, which is allowed.

addons/nuxeo-ai-core/elements/admin-ai-export.js
```diff
@@ -23,157 +25,194 @@ import { LayoutBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-layout-behavior.j
...
       .catch(() => {
-        this._exportText = '';
         this.fire('notify', { message: this.i18n('admin.ai.export.error') });
       });
-  },
-});
+  }
+
+  _canExport() {
+    return this.typeName.length !== 0 && this.$.currentDoc !== undefined && this.$.currentDoc !== null;
```
Gabez0r:
```suggestion
    return this.typeName.length !== 0 && this.$.currentDoc != null;
```

andreinechaev: done

nmpcunha:
Should the commit message format be: `NXP-27113: some message...`?
We are missing the `:` separator.

addons/nuxeo-ai-core/elements/admin-ai-export.js
```diff
@@ -23,157 +25,194 @@ import { LayoutBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-layout-behavior.j
...
+        .totalLabel {
+          padding-right: 20px;
+        }
+
+        .row-container {
+          @apply --layout-horizontal;
+          @apply --layout-wrap;
+        }
+
+        /* #exportDatasetBtn {
```
nmpcunha: Do we need this commented code?

.gitignore
```diff
@@ -9,4 +9,6 @@ npm-debug.log
 .DS_Store
 .screenshots/
 *.iml
-.idea
\ No newline at end of file
+.idea
```
nmpcunha: Should those be added locally? There are several opinions on that, but as far as I remember, we tried to keep gitignore more generic/simple possible.

addons/nuxeo-ai-core/elements/admin-ai-export.js
andreinechaev: we might need it later according to our design.

.gitignore
andreinechaev: idea is a widely used IDE within our company, I was surprised not seeing it

PR #648, opened 2019-07-01
NXP-27119: make AI suggestions customizable

Gabez0r:


addons/nuxeo-ai-core/elements/nuxeo-ai-suggestion.js
```diff
@@ -0,0 +1,87 @@
...
+    return 'nuxeo-ai-suggestion';
+  }
+
+  static get observers() {
+    return ['_updateContainer(suggestion, property)'];
+  }
+
+  created() {
+    Object.keys(this.constructor.properties).forEach((prop) => {
+      this._createMethodObserver(`_forwardProp("${prop}", ${prop})`);
```
nelsonsilva: Mixin PropertyEffects gives us a public `createMethodObserver`, right ?

addons/nuxeo-ai-core/elements/nuxeo-ai-suggestion.js
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  _forwardProp(prop, value) {
+    if (this._instance) {
+      this._instance[prop] = value;
+    }
+  }
+
+  _updateContainer(suggestion, property) {
+    this._instance = document.createElement(
```
nelsonsilva: I'd rather split these nested calls into some temp vars for clarity

addons/nuxeo-ai-core/index.js
```diff
@@ -134,7 +145,12 @@ document.addEventListener('document-layout-changed', (e) => {
...
   }
   const layout = e.detail.element;
-  layout._documentChanged = () => AISuggestionManager.updateWidgetSuggestions(layout);
-  layout._createMethodObserver('_documentChanged(document.*)', true);
-  AISuggestionManager.updateWidgetSuggestions(e.detail.element);
+  if (!layout) {
+    return;
+  }
+  customElements.whenDefined(layout.tagName.toLowerCase()).then(() => {
+    layout._documentChanged = () => AISuggestionManager.updateWidgetSuggestions(layout);
```
nelsonsilva: kind of hacky to inject a method to register as observer but if we're to do it I'd rather go with a name with fewer changes of conflct ie `__aiDocumentChanged` or something along those lines

addons/nuxeo-ai-core/elements/nuxeo-ai-suggestion-formatters.js
```diff
@@ -0,0 +1,52 @@
...
+    }
+    if (!formatter && _formatters.types && rules.type) {
+      formatter = _formatters.types[rules.type];
+    }
+    return formatter || _formatters.default;
+  }
+})();
+Nuxeo.AISuggestionFormatters = AISuggestionFormatters;
+
+export default AISuggestionFormatters;
```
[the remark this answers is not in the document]
Gabez0r: This would give people direct access to FORMATTERS, allowing people to circumvent the register and get methods.

addons/nuxeo-ai-core/elements/nuxeo-ai-suggestion-mixin.js
```diff
@@ -0,0 +1,72 @@
...
+        match: {
+          type: Boolean,
+          value: false,
+          reflectToAttribute: true,
+        },
+      };
+    }
+
+    _idFunction(item) {
+      const id = ['computeId', 'uid', 'id'].find((key) => Object.prototype.hasOwnProperty.call(item, key));
```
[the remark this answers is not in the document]
Gabez0r: For security reasons (linting): https://eslint.org/docs/rules/no-prototype-builtins

addons/nuxeo-ai-core/index.js
Gabez0r: We've always had it like this, but I don't foresee we'll break anything by changing this name, since this is a private method.

addons/nuxeo-ai-core/elements/nuxeo-ai-suggestion.js
Gabez0r: Yes. Would you rather extend the mixing to use the public method?

addons/nuxeo-ai-core/elements/nuxeo-ai-suggestion-formatters.js
```diff
@@ -0,0 +1,52 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+import DefaultAISuggestionFormatter from './formatters/nuxeo-default-ai-suggestion-formatter.js';
+
+const _formatters = {
```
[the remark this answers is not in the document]
Gabez0r: Agree with the rename to `FORMATTERS` and initialization. Not so sure about renaming an object containing rules for several types just as type (singular).

addons/nuxeo-ai-core/elements/nuxeo-ai-suggestion.js
nelsonsilva: If possible I'd say it's better to rely on public API so we get a heads up when they break it + a migration path 🤞

Gabez0r: It turns out that this mixin is already extended by `Polymer.Element`, and the method is already available as a static method (needs to be invoked from the constructor).

addons/nuxeo-ai-core/elements/formatters/nuxeo-document-ai-suggestion-formatter.js
```diff
@@ -0,0 +1,265 @@
...
+        #split:active {
+          background-color: #ceced4;
+        }
+        #content {
+          @apply --layout-vertical;
+          font-size: 11px;
+          line-height: normal;
+        }
+        #content p {
+          margin-block-start: 0.25em;
```
semisse: `margin-block-start` & `margin-block-end` is not currently supported on Edge. Is this a problem for us?

Gabez0r: We should find a workaround for this if possible. Do you know what styles Edge applies to the `p` tag, so we can override them?

addons/nuxeo-ai-core/index.js
```diff
@@ -19,13 +19,22 @@ import { timeOut } from '@polymer/polymer/lib/utils/async.js';
...
 const tmpl = document.createElement('template');
 tmpl.innerHTML = html;
 document.head.appendChild(tmpl.content);
 
+AISuggestionFormatters.register(DocumentAISuggestionFormatter.is, { type: 'document' });
+AISuggestionFormatters.register(DirectoryAISuggestionFormatter.is, { type: 'directoryEntry' });
+AISuggestionFormatters.register(UserGroupAISuggestionFormatter.is, { type: 'user' });
+AISuggestionFormatters.register(UserGroupAISuggestionFormatter.is, { type: 'group' });
+
 const AISuggestionManager = (() => {
```
nelsonsilva: Do we need this IIFE ?!

Gabez0r: We don't on master but we do on 10.10, so keeping it makes backporting easier.

PR #658, opened 2019-07-11
NXP-27462: handle task submission errors gracefully

nmpcunha:


i18n/messages.json
```diff
@@ -917,6 +917,9 @@
   "tasks.viewGraph": "View Graph",
   "tasks.delegate": "Delegate",
   "tasks.reassign": "Reassign",
+  "tasks.submit.error": "An error occurred while submitting the task.",
+  "tasks.submit.error.alreadyFinished": "The task you tried to submit was already finished by other user.",
+  "tasks.submit.error.noPermissions": "You don't have the permissions needed to process this task.",
```
Gabez0r: Perhaps something shorter and with the same meaning like:
```suggestion
  "tasks.submit.error.noPermissions": "You don't have permissions to process this task.",
```

nmpcunha: Thanks!

elements/workflow/nuxeo-document-task.js
```diff
@@ -220,9 +220,24 @@ Polymer({
...
-      });
+      this.$.taskRequest
+        .put()
+        .then((task) => {
+          this.fire('workflowTaskProcessed', { task });
+        })
+        .catch((error) => {
+          if (error.status === 409 || error.status === 403) {
+            this.fire('notify', {
+              message: this.i18n(`tasks.submit.error.${error.status === 409 ? 'alreadyFinished' : 'noPermissions'}`),
```
guirenard: I would'n mind a `switch` or whatever structure to clearly set the `message` accrodign to http sc. Here it is still readable until yet another sc code to handle

i18n/messages.json
```diff
@@ -917,6 +917,9 @@
   "tasks.viewGraph": "View Graph",
   "tasks.delegate": "Delegate",
   "tasks.reassign": "Reassign",
+  "tasks.submit.error": "An error occurred while submitting the task.",
+  "tasks.submit.error.alreadyFinished": "The task you tried to submit was already finished by other user.",
```
guirenard: `another user` ?

elements/workflow/nuxeo-document-task.js
nmpcunha: Since we only have two cases now, I would keep as is. Once we have more, I'll definitively say that we should change it to a `switch` or `else if` structure.

i18n/messages.json
nmpcunha: Thanks!

PR #672, opened 2019-07-23
NXP-27115: progress bar for advanced export

andreinechaev:


addons/nuxeo-ai-core/elements/admin-ai-advanced-export.js
```diff
@@ -202,6 +265,30 @@ class AdminAIAdvancedExport extends mixinBehaviors([I18nBehavior], Nuxeo.Element
...
+
+  _statusReceived(resp) {
+    this.statuses = resp.detail.response.value;
+    // defines if the tab is currently active if so, we want to poll more often
+    if (this.visible) {
+      this._timeOut = setTimeout(this._poll, 2000, this.$.aiExportStatus);
+    }
+  }
+
+  _poll(func) {
```
nmpcunha: It seems that this function is not used anywhere else despite inside `_statusReceived`.
I would suggest to integrate it in order to reduce the number of lines.

andreinechaev: passing it as a reference with explicit name makes the intention clearer IMO
second :) it's a migration of https://github.com/nuxeo/nuxeo-web-ui/pull/642 from simple to advanced view.

Gabez0r: Couldn't we write
```
setTimeout(this._poll, 2000, this.$.aiExportStatus);
```
as
```
setTimeout(() => this.$.aiExportStatus.execute(), 2000);
```
?

If so, I'd say it would be easier to read and save us a few lines of code.

Also, `_poll(func) {` is misleading since it does not receive a function but an object instead, on which it calls the `execute` method.

Indeed, we let this through #642, but I'd recommend doing this cleanup on `admin-ai-export.js` as well.

andreinechaev: ok, let me try.

andreinechaev: done.

addons/nuxeo-ai-core/elements/admin-ai-advanced-export.js
```diff
@@ -202,6 +265,26 @@ class AdminAIAdvancedExport extends mixinBehaviors([I18nBehavior], Nuxeo.Element
...
+
+  _statusReceived(resp) {
+    this.statuses = resp.detail.response.value;
+    // defines if the tab is currently active if so, we want to poll more often
+    if (this.visible) {
+      this._timeOut = setTimeout(() => this.$.aiExportStatus.execute(), 2000);
+    }
+  }
+
+  _updateItem(e) {
```
nmpcunha: Is this method being used?

andreinechaev: nope. thx.

PR #675, opened 2019-07-25
NXP-26660: add tagExpression capability

nmpcunha:


packages/nuxeo-web-ui-ftest/scripts/specs-processor.js
```diff
@@ -12,7 +13,14 @@ module.exports = (argv) => {
    */
   if (args.cucumberOpts && args.cucumberOpts.tagExpression) {
     const files = Finder.from('./test/features').findFiles('*.feature');
-    features = files.filter((file) => fs.readFileSync(file, 'utf8').includes(args.cucumberOpts.tagExpression));
+    const expression = new TagExpressionParser().parse(args.cucumberOpts.tagExpression);
+    features = files.filter((file) => {
+      const tags = fs.readFileSync(file, 'utf8').match(/@\w+/g);
```
nelsonsilva: can we match only at beginning of the line or something ? ie `/^@.*/g` 
this would also match the whole line so we can actually use operators in the expression (and, or, not, ...)

Gabez0r: This won't work if we only have a tag before a scenario, because of the indentation. We could try something like `^\s*@.+`, though.

nmpcunha: We won't be able to limit to the beginning of the line since not all tags may be there.
Also, looking for something with spaces after or so may not be that accurate according to https://cucumber.io/docs/cucumber/api/#tag-expressions
Not sure how we can improve this RegEx further.

Gabez0r: This won't work if we only have a tag before a scenario, because of the indentation. We could try something like `^\s*@.+`, though.

nmpcunha: It would imply a little bit more processing, but it can be an approach.
The only issue with the current one is that it can fetch false positives, which is not that bad at all because at the end, TagExpressionParser will process the tags accordingly.
Our custom processor is only "pre filtering"  .feature files

nmpcunha: Hopefully, an improved solution! ;)

PR #678, opened 2019-07-31
NXP-27487: adding AI_Corpus to AI_Model

andreinechaev:


addons/nuxeo-ai-core/elements/admin-ai-export.js
```diff
@@ -192,6 +192,10 @@ class AdminAIExport extends mixinBehaviors([I18nBehavior], Nuxeo.Element, Layout
       query,
       inputs: input,
       outputs: output,
+      model_id: this.$.currentDoc.uid,
```
nmpcunha: Please notice that `$` is used with a specific purpose (see https://polymer-library.polymer-project.org/1.0/docs/devguide/local-dom#work-with-local-dom).

For that reason, `this.$.currentDoc` does not map to our property `currentDoc`, which is `this.currentDoc`.

nmpcunha: Despite this small (but maybe someday big) detail, everything seems fine 👍

andreinechaev: ok, done. changed for all usages.

PR #683, opened 2019-08-06
NXP-27225: send parameters to AI suggestions operation

nmpcunha:


addons/nuxeo-ai-core/index.js
```diff
@@ -93,6 +93,10 @@ const AISuggestionManager = (() => {
       document.querySelector('nuxeo-app').appendChild(op);
     }
     op.input = doc;
```
Gabez0r: Do we still need to send the document as input, since it is already being passed as-is in the parameters?

Gabez0r: Had a look at the operation, and we're just using `updatedDocument` to carry the changed properties, right?

PR #686, opened 2019-08-09
NXP-27710: ensure document refresh after lock operations

nmpcunha:
ELEMENTS: https://github.com/nuxeo/nuxeo-elements/pull/147

### DIFF
--- a/elements/nuxeo-app.js
+++ b/elements/nuxeo-app.js
@@ -1148,7 +1148,7 @@ Polymer({
 
   _refreshAndFetchTasks() {
     // let's refresh the current document since it might have been changed (ex: state and version)
-    this.refresh();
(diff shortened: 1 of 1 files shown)


elements/nuxeo-app.js
```diff
@@ -1238,10 +1238,12 @@ Polymer({
 
   _documentLocked() {
     this._toast(this.i18n('app.document.locked'));
+    this.refresh();
```
nelsonsilva: I guess we should `this.fire('document-updated');` instead for consistency...

nmpcunha: Will also replace the remaining `this.refresh()` call on `_refreshAndFetchTasks`

PR #687, opened 2019-08-09
NXP-27710: ensure document refresh after lock operations (Backport 10.10)

nmpcunha:


elements/nuxeo-app.html
```diff
@@ -881,7 +881,7 @@
 
       _refreshAndFetchTasks: function() {
         // let's refresh the current document since it might have been changed (ex: state and version)
```
RSalem07: as you removed the call to the `refresh` method, i don't know if we should keep the comment, the name of the event is enough i think :) 

thx

nmpcunha: For coherency purposes, in all other places we are firing the event `document-updated` which will trigger the `refresh()` method.

PR #688, opened 2019-08-13
NXP-27706: suggestion actions on result list

andreinechaev:


addons/nuxeo-ai-core/elements/nuxeo-ai-bulk-add.js
```diff
@@ -0,0 +1,95 @@
...
+ */
+
+class AIBulkAddButton extends mixinBehaviors([I18nBehavior, FiltersBehavior], Nuxeo.Element) {
+  static get template() {
+    return html`
+      <nuxeo-operation-button
+        id="aiBulkAddOpBtn"
+        operation="Bulk.RunAction"
+        input="[[provider]]"
+        params="[[_params(provider)]]"
```
nelsonsilva: Seems `params` does not change with `provider`, in fact `params` are constant so you can just pass them as a JSON string here `params='{"action": "..", ...}'`

addons/nuxeo-ai-core/elements/nuxeo-ai-bulk-add.js
```diff
@@ -0,0 +1,95 @@
...
+    };
+  }
+
+  _responseReceived() {
+    this.fire('notify', { message: this.i18n('ai.bulkAddButton.action.poll') });
+  }
+}
+
+customElements.define(AIBulkAddButton.is, AIBulkAddButton);
+Nuxeo.AIBulkAddButton = AIBulkAddButton;
```
nelsonsilva: nit: perhaps having a `Nuxeo.AI` namespace would make things cleaner...

addons/nuxeo-ai-core/elements/nuxeo-ai-bulk-remove.js
```diff
@@ -0,0 +1,120 @@
...
+ */
+
+class AIBBulkRemoveButton extends mixinBehaviors([I18nBehavior, FiltersBehavior], Nuxeo.Element) {
+  static get template() {
+    return html`
+      <nuxeo-operation-button
+        id="aiBulkRemoveOpBtn"
+        operation="Bulk.RunAction"
+        input="[[provider]]"
+        params="[[_params(provider, modelId, xpaths)]]"
```
nelsonsilva: params do not change with `provider`

addons/nuxeo-ai-core/index.js
```diff
@@ -21,6 +21,8 @@ import './elements/admin-ai-export.js';
 import './elements/admin-ai-advanced-export.js';
 import './elements/nuxeo-ai-suggestions.js';
 import './elements/nuxeo-ai-export-progress.js';
+import './elements/nuxeo-ai-bulk-add.js';
```
nelsonsilva: nit: perhaps we should start organizing things into folder (ie `actions´)

addons/nuxeo-ai-core/elements/actions/nuxeo-ai-bulk-add.js
```diff
@@ -0,0 +1,96 @@
...
+ */
+
+class BulkAddButton extends mixinBehaviors([I18nBehavior, FiltersBehavior], Nuxeo.Element) {
+  static get template() {
+    return html`
+      <nuxeo-operation-button
+        id="aiBulkAddOpBtn"
+        operation="Bulk.RunAction"
+        input="[[provider]]"
+        params="[[_params()]]"
```
nmpcunha: Just for the record, other possible approach for this would be a computed property.

nelsonsilva:
> Cool.
> Just noticed that both actions are calling the same operation with different parameters. Maybe it is too soon to think further, but we may want to have something like "BulkAction" which would be something like a parent class for these kind of actions. Let's see how it evolves!

I'm personally not a fan of using `Bulk.RunAction` from the client and would rather have more "semantic" (properly named) operations relying on actions internally instead of exposing yet another concept to clients.

addons/nuxeo-ai-core/elements/actions/nuxeo-ai-bulk-add.js
nmpcunha: Just to confirm, it is expected to have this as params:
`{
      action: 'bulkEnrich',
      parameters: {},
  }`

If so, you may want to change the method itself.

addons/nuxeo-ai-core/elements/actions/nuxeo-ai-bulk-add.js
```diff
@@ -0,0 +1,96 @@
...
+       */
+      status: {
+        type: Object,
+        notify: true,
+      },
+    };
+  }
+
+  _params() {
+    const actionParams = {};
```
nmpcunha: Always empty? Why using the variable?

addons/nuxeo-ai-core/elements/actions/nuxeo-ai-bulk-remove.js
```diff
@@ -0,0 +1,121 @@
...
+ */
+
+class BulkRemoveButton extends mixinBehaviors([I18nBehavior, FiltersBehavior], Nuxeo.Element) {
+  static get template() {
+    return html`
+      <nuxeo-operation-button
+        id="aiBulkRemoveOpBtn"
+        operation="Bulk.RunAction"
+        input="[[provider]]"
+        params="[[_params()]]"
```
nmpcunha: If **modelId** or **xpaths** changes, we may want to recompute it, no?

addons/nuxeo-ai-core/elements/actions/nuxeo-ai-bulk-add.js
andreinechaev: doesn't work with no parameters. would require changes in the action itself

addons/nuxeo-ai-core/elements/actions/nuxeo-ai-bulk-remove.js
andreinechaev: don't think so. the action runs on results of a page provider on the current page

addons/nuxeo-ai-core/elements/actions/nuxeo-ai-bulk-add.js
nmpcunha: Not sure if I understood.
The following code, for example, wouldn't work?
`return {
      action: 'bulkEnrich',
      parameters: '{}',
 };`

And, since it is something static, we could have it directly on a private property, avoiding some lines of code.

Gabez0r: Agree this `actionParams` variable is redundant.

addons/nuxeo-ai-core/elements/actions/nuxeo-ai-bulk-add.js
Gabez0r: Since this method always returns the same object, you could even inline it:
```suggestion
        params='{ action: "bulkEnrich", parameters: "{}" }'
```

addons/nuxeo-ai-core/elements/actions/nuxeo-ai-bulk-remove.js
```diff
@@ -0,0 +1,121 @@
...
+        type: Object,
+        notify: true,
+      },
+    };
+  }
+
+  _params() {
+    const actionParams = {};
+    if (this.modelId) {
+      actionParams.modelId = this.modelId.split(',').map((s) => s.trim());
```
Gabez0r: If the `modelId` changes, then these params are invalid and need to be recomputed.

addons/nuxeo-ai-core/elements/actions/nuxeo-ai-bulk-remove.js
```diff
@@ -0,0 +1,121 @@
...
+    }
+    if (this.xpaths) {
+      actionParams.xpaths = this.xpaths.split(',').map((s) => s.trim());
```
Gabez0r: Same here.

addons/nuxeo-ai-core/elements/actions/nuxeo-ai-bulk-remove.js
Gabez0r: Looking at `_params`, `modelId` and `xpaths` are indeed dependencies of this computed binding (see comments above).

nmpcunha: It is related on how polymer handles the data flow (see more [here](https://polymer-library.polymer-project.org/3.0/docs/devguide/data-system)).

The suggestion was to have ` params="[[_params(modelId, xpaths)]]"`
This means that this method will be called every time `modelId` and/or `xpaths` change. By having that binding on DOM we are basically saying that we are observing those properties for that specific method.

No, we don't need to have those properties on method signature, but the reason why, you can find on the link above.

andreinechaev: ok, got it. thanks. repushed

PR #689, opened 2019-08-14
NXP-26504: spreadsheet available only on table view

nmpcunha:


addons/nuxeo-spreadsheet/nuxeo-spreadsheet.html
```diff
@@ -14,7 +14,7 @@
 See the License for the specific language governing permissions and
 limitations under the License.
 -->
-<nuxeo-slot-content name="spreadsheetButton" slot="RESULTS_ACTIONS">
+<nuxeo-slot-content name="spreadsheetButton" slot="RESULTS_ACTIONS" order="-150">
```
nelsonsilva: why the `-150` order?

nmpcunha: Since spreasheet button will appear and disappear when changing views, this order will ensure it will be to the left of csv button, which is there by default and is consistent between views.

PR #704, opened 2019-08-28
NXP-27880: multi valued suggestions

nmpcunha:


addons/nuxeo-ai-core/index.js
```diff
@@ -107,41 +107,53 @@ const AISuggestionManager = (() => {
...
+      if (widget && Array.isArray(widget.suggestions) && widget.suggestions.length > 0) {
+        widget._matchInput();
+      }
+      if (!path || isModelInput) {
+        _updateDebouncer = Debouncer.debounce(_updateDebouncer, timeOut.after(500), () => {
+          _getSuggestions(layout.document).then((response) => {
+            /*
+             * Inefficient approach that should be changed as soon as the response from server includes more information
+             * (e.g. all output fields) to only clear the elements in need.
+             * Ticket created to analyse and tackle this need: NXP-26314
```
Gabez0r: Since NXP-26314 is already closed, and we are already did optimizations to take the input fields into account, we should probably either remove this comment block or reference some other ticket.

nmpcunha: Fair enough!

PR #711, opened 2019-09-10
Feature nxp 27692 retention web ui 10.10

guirenard:


elements/directory/l10nvocabulary/nuxeo-l10nvocabulary-edit-layout.html
```diff
@@ -0,0 +1,102 @@
...
+                 label="[[i18n('vocabularyManagement.edit.label_en')]]"
+                 name="label"
+                 value="{{entry.properties.label_en::change}}">
+    </nuxeo-input>
+
+    <nuxeo-input role="widget"
+                 label="[[i18n('vocabularyManagement.edit.label_fr')]]"
+                 name="label"
+                 value="{{entry.properties.label_fr::change}}">
+    </nuxeo-input>
```
nmpcunha: Is this the expected? English and French?

elements/nuxeo-document-info-bar/nuxeo-document-info-bar.html
```diff
@@ -108,6 +113,27 @@
...
+            <span id="legalHold">[[i18n('documentPage.legalHold')]]</span>
+          </template>
+          <template is="dom-if" if="[[document.retainUntil]]">
+            <iron-icon icon="nuxeo:retain"></iron-icon>
+            <template is="dom-if" if="[[!isRetentionDateIndeterminate(document)]]">
+              <span id="retention" hidden="[[document.hasLegalHold]]">[[_computeRetentionUntiLabel(document)]]</span>
+            </template>
+            <template is="dom-if" if="[[isRetentionDateIndeterminate(document)]]">
+              <span id="indeterminateRetention" hidden="[[document.hasLegalHold]]">[[i18n('documentPage.retainIndeterminate')]]</span>
+            </template>
```
nmpcunha: If I understand this correctly, we'll always have a span and, despite the id, only the text content changes... Should we pass the conditional logic for the computation method instead?

elements/directory/l10nvocabulary/nuxeo-l10nvocabulary-edit-layout.html
Gabez0r: Yes, l10n vobacularies come with translated en and fr labels.

elements/nuxeo-document-info-bar/nuxeo-document-info-bar.html
Gabez0r: We could, but I know the distinct id's are used in the tests to distinguis both cases more easily. Alternative here is to either compute a class or check a label, with a latter being something we should avoid...

PR #724, opened 2019-09-19
NXP-26685: meaningful path in nuxeo-document-publications

semisse:


semisse:
To avoid the problem that is referenced on ELEMENTS-1002, we are using here a `&lrm;` as workaround, until we find a better solution to handle these cases.

elements/nuxeo-publication/nuxeo-document-publications.js
```diff
@@ -46,12 +46,21 @@ Polymer({
...
+        direction: rtl;
+        text-align: left;
+      }
+
+      .right-ellipsis {
+        text-align: right;
+      }
+
       .ellipsis {
+        width: calc(100% - 38px);
```
nmpcunha: any specific reason to change the order?

elements/nuxeo-publication/nuxeo-document-publications.js
```diff
@@ -117,7 +121,9 @@ Polymer({
         <nuxeo-data-table-column name="[[i18n('documentContentView.datatable.header.path')]]" flex="200">
           <template>
             <nuxeo-document-thumbnail document="[[item]]"></nuxeo-document-thumbnail>
-            <a class$="path ellipsis [[_ellipsisDirection()]]" href$="[[urlFor('browse', item.path)]]">[[item.path]]</a>
+            <a class$="path ellipsis [[_ellipsisDirection()]]" href$="[[urlFor('browse', item.path)]]"
+              >[[item.path]]&lrm;</a
+            >
```
nmpcunha: Why was it different from the first commit? Was it lint?

elements/nuxeo-publication/nuxeo-document-publications.js
semisse: This doesn't have a particular impact in the visuals, it is just regarding the discussion of having some consistency on listing the CSS properties (and using the Box Model to do it).

elements/nuxeo-publication/nuxeo-document-publications.js
semisse: Yes, probably the introduction of `&lrm;` exceeded the number of characters allowed in one line.

nmpcunha: Of course... 🤦‍♂

elements/nuxeo-publication/nuxeo-document-publications.js
```diff
@@ -33,6 +33,10 @@ import { html } from '@polymer/polymer/lib/utils/html-tag.js';
 Polymer({
   _template: html`
     <style include="nuxeo-styles">
+      [hidden] {
+        display: none !important;
```
nmpcunha: Do we need to force with `!important`?

elements/nuxeo-publication/nuxeo-document-publications.js
nmpcunha: Only impact on PR diff visual 😜 
I foresee the reason and I'm not against it, but please recall that the discussion didn't took place yet and there is still no agreement on that guideline.

elements/nuxeo-publication/nuxeo-document-publications.js
```diff
@@ -154,7 +160,7 @@ Polymer({
             </template>
           </template>
         </nuxeo-data-table-column>
-        <nuxeo-data-table-column flex="10">
+        <nuxeo-data-table-column flex="10" hidden$="[[!_canRepublish()]]">
```
nmpcunha: This method expects an input, correct?
Also, what would be the visual impact when having multiple documents, some that can be republished, some that cannot?

semisse:
We've ditched the idea of checking if the document is republishible or unpublishible because it might be tricky (in case the document has hundreds of versions, for example).

semisse:
Pipeline ok: https://qa.nuxeo.org/jenkins/job/web-ui-pipeline-sfialho/

PR #728, opened 2019-09-23
NXP-28064: add left-to-right mark to breadcrumbs

semisse:


semisse:
For the record: this solution is using a hacky way to revert ellipsis, based on language direction. It should be improved in the scope of this ticket: https://jira.nuxeo.com/browse/NXP-28073

semisse:
Pipeline ok: https://qa.nuxeo.org/jenkins/job/web-ui-pipeline-sfialho/

PR #738, opened 2019-09-30
Task nxp 27985 cent os nginx

guirenard:


nginx/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+    yum -y autoremove &&\
+    yum clean all &&\
+    wget http://nginx.org/packages/$os/$osversion/x86_64/RPMS/nginx-$nginxversion.el$elversion.ngx.x86_64.rpm &&\
+    rpm -iv nginx-$nginxversion.el$elversion.ngx.x86_64.rpm
+
+# forward request and error logs to docker log collector
+RUN ln -sf /dev/stdout /var/log/nginx/access.log \
+	&& ln -sf /dev/stderr /var/log/nginx/error.log
+
+EXPOSE 80 443
```
guirenard: There's no simple image with recent nginx and centos.
AFAIK, the most recent version of nginx available in rpm is nginx-1.16.1 http://nginx.org/packages/centos/7/x86_64/RPMS/ 
Note that redhat provides https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/rhscl/nginx-114-rhel7 but it has specific stuff for openshift and has no public availability.

If this is the way to go, perhaps we could publish in our own docker image registry this one rather than having it buiold and compose here.

PR #745, opened 2019-10-08
NXP-28068: fix default-search buttons position

semisse:


semisse:
Pipeline ok: https://qa.nuxeo.org/jenkins/job/web-ui-pipeline-sfialho/100/

elements/search/nuxeo-search-form.js
```diff
@@ -706,10 +702,23 @@ Polymer({
     return this.selectedSearchIdx > 0;
   },
 
+  _reset() {
```
nmpcunha: nit: usually we don't change method's location unless we have a strong reason to. It makes the diff more "messy"

elements/search/nuxeo-search-form.js
```diff
@@ -706,10 +702,23 @@ Polymer({
...
+        this.form.searchTerm = this.searchTerm;
+        this.dirty = false;
+      });
+    }
+  },
+
   _clear() {
     this.searchTerm = '';
-    this.isSavedSearch = false;
-    this.selectedSearch = null;
```
guirenard: Are 100% sure these 2 lines ca be removed?

semisse: Actually, after giving a closer look, I'm sure we can't remove them. 
The way I was approaching this was something along these lines: this function was running in both default-search and saved search. With this new solution, it only runs on the first one. This way we didn't need to reset the those values. Although, there is a situation when we go back and forward between default and save searches that we actually need to reset them.

PR #746, opened 2019-10-09
NXP-27803: implement IronValidatableBehavior and mark invalid when fi…

guirenard:
…le upload pending

### DIFF
--- a/elements/nuxeo-dropzone/nuxeo-dropzone.js
+++ b/elements/nuxeo-dropzone/nuxeo-dropzone.js
@@ -16,6 +16,7 @@ limitations under the License.
 */
 import '@polymer/polymer/polymer-legacy.js';
 
+import { IronValidatableBehavior } from '@polymer/iron-validatable-behavior/iron-validatable-behavior.js';
 import '@polymer/paper-icon-button/paper-icon-button.js';
 import '@polymer/paper-progress/paper-progress.js';
 import '@nuxeo/nuxeo-elements/nuxeo-connection.js';
@@ -53,8 +54,20 @@ Polymer({
         background-color: var(--paper-input-container-focus-background-color, rgba(0, 102, 255, 0.1));
       }
 
+      :host([invalid]) #container {
+        border: 2px dashed var(--paper-input-container-invalid-color, red);
+      }
+
       label {
         margin-bottom: 8px;
+      }
+
+      label[required]::after {
+        display: inline-block;
+        content: '*';
+        margin-left: 4px;
+        color: var(--paper-input-container-invalid-color, red);
+        font-size: 1.2em;
       }
 
       a,
@@ -93,6 +106,10 @@ Polymer({
         margin-bottom: 8px;
       }
 
+      :host([invalid]) #details {
+        border: 2px solid var(--paper-input-container-invalid-color, red);
+      }
+
       .file {
         @apply --layout-horizontal;
         @apply --layout-center;
@@ -124,13 +141,18 @@ Polymer({
       .actions > * {
         margin-left: 8px;
       }
+
+      .error {
+        color: var(--paper-input-container-invalid-color, red);
+        margin-top: 8px;
+      }
     </style>
 
     <nuxeo-connection id="nx"></nuxeo-connection>
     <nuxeo-document id="doc" doc-id="[[document.uid]]" enrichers="[[enrichers]]"></nuxeo-document>
 
     <template is="dom-if" if="[[label]]">
-      <label id="label">[[label]]</label>
+      <label id="label" required$="[[required]]">[[label]]</label>
     </template>
 
     <input hidden id="input" type="file" on-change="_uploadInputFiles" />
@@ -175,10 +197,12 @@ Polymer({
         </div>
       </div>
     </div>
+
+    <span class="error" hidden$="[[!invalid]]">[[_errorMessage]]</span>
   `,
 
   is: 'nuxeo-dropzone',
-  behaviors: [UploaderBehavior, FormatBehavior],
+  behaviors: [UploaderBehavior, FormatBehavior, IronValidatableBehavior],
 
   properties: {
     /**
@@ -278,6 +302,17 @@ Polymer({
       value() {
         return this._computeEnrichers();
       },
+    },
+    _errorMessage: {
+      type: String,
+    },
+    /**
+     * Required.
+     */
+    required: {
+      type: Boolean,
+      value: false,
+      reflectToAttribute: true,
     },
   },
 
@@ -383,6 +418,7 @@ Polymer({
     } else {
       this.fire('notify', { message: this.i18n(this.uploadedMessage) });
     }
+    this.invalid = false;
   },
 
   _deleteFile(e) {
@@ -490,5 +526,16 @@ Polymer({
 
   _showAbort(uploading) {
     return uploading && this.hasAbort();
+  },
+
+  _getValidity() {
+    if (this.uploading) {
+      this._errorMessage = this.i18n('dropzone.invalid.uploading');
+      return false;
+    }
+    if (!this.required) {
+      return true;
+    }
+    return this.files && this.files.length > 0;
   },
 });
--- a/i18n/messages.json
+++ b/i18n/messages.json
@@ -535,6 +535,7 @@
   "driveSyncToggleButton.unsync": "Unsynchronize",
   "dropzone.abort": "Abort file upload",
   "dropzone.add": "Upload main file",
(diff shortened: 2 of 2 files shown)


elements/nuxeo-dropzone/nuxeo-dropzone.js
```diff
@@ -93,6 +106,10 @@ Polymer({
         margin-bottom: 8px;
       }
 
+      :host([invalid]) #details {
+        border: 1.5px solid var(--paper-input-container-invalid-color, red);
```
semisse: nit: Why are we using 1.5px here?

guirenard: 1px sounded too thin and 2px too thick but I guess it is completely subjective :D

elements/nuxeo-dropzone/nuxeo-dropzone.js
```diff
@@ -306,6 +338,17 @@ Polymer({
     return this.formatPropertyXpath(xpath);
   },
 
+  _getValidity() {
+    if (this.uploading) {
+      this.errorMessage = this.i18n('dropzone.invalid.uploading');
```
nmpcunha: If this happens the `errorMessage` property is overridden and we lost the default one.

So, let's assume I've uploaded a file, tried to go further and received this message. Then I've noticed I have the wrong blob and I deleted it. If I try to go further (clicking create) without a blob (and assuming it is required), I'll get this `dropzone.invalid.uploading` message. Am i seeing this correctly?

guirenard:
> Also, when the upload finishes, should we clear the error message and the visual warning?

In the preview env, we had a file size limit set to 8MB by ingress by default. The upload was interrupted with a `413` HTTP SC and that's why the dropzone remained invalid.

I refactored a bit to be able to explicitly display such error:
 - https://github.com/nuxeo/nuxeo-web-ui/pull/746/files#diff-bfcef58b64ac39d85e6fc62124c66171R547
 - https://github.com/nuxeo/nuxeo-web-ui/pull/746/files#diff-84df3bde9e8c2ccc631ab4993c4fe7adR530

elements/nuxeo-dropzone/nuxeo-dropzone.js
guirenard: You are seeing this right. I am gonna turn **errorMessage** into a private property

elements/nuxeo-dropzone/nuxeo-dropzone.js
```diff
@@ -279,11 +303,25 @@ Polymer({
         return this._computeEnrichers();
       },
     },
+    /**
+     * Error message to show when `invalid` is true.
+     */
+    _errorMessage: String,
```
Gabez0r: We could even:
```suggestion
    errorMessage: {
      type: String,
      readOnly: true,
    },
```

elements/nuxeo-dropzone/nuxeo-dropzone.js
```diff
@@ -491,4 +530,22 @@ Polymer({
   _showAbort(uploading) {
     return uploading && this.hasAbort();
   },
+
+  _getValidity() {
+    if (this.uploading) {
+      this._errorMessage = this.i18n('dropzone.invalid.uploading');
```
Gabez0r: And then do:
```suggestion
      this._setErrorMessage(this.i18n('dropzone.invalid.uploading'));
```

elements/nuxeo-dropzone/nuxeo-dropzone.js
```diff
@@ -491,4 +530,22 @@ Polymer({
...
+    if (!this.required) {
+      return true;
+    }
+    return this.files && this.files.length > 0;
+  },
+
+  _handleError(e) {
+    if (e.detail && e.detail.error && e.detail.error.response && e.detail.error.response.status) {
+      this.invalid = true;
+      this._errorMessage = this.i18n(`file.invalid.${e.error.response.status}`);
```
Gabez0r: typo in label?
```suggestion
      this._setErrorMessage(this.i18n(`dropzone.invalid.${e.error.response.status}`));
```

elements/nuxeo-dropzone/nuxeo-dropzone.js
```diff
@@ -175,10 +197,12 @@ Polymer({
         </div>
       </div>
     </div>
+
+    <label class="error" hidden$="[[!invalid]]">[[_errorMessage]]</label>
```
Gabez0r: We should use a `span` here, to be consistent with what we already do on `nuxeo-document-create`, `nuxeo-document-import`, etc. Also, since this element uses `nuxeo-styles`, the label will receive the `--nuxeo-label` style, which has opacity and might be unwanted here?

guirenard: I was trying to be consistent with https://github.com/nuxeo/nuxeo-elements/blob/master/ui/widgets/nuxeo-selectivity.js#L7075 but OK

charts/preview/values.yaml
```diff
@@ -21,7 +21,6 @@ nuxeo:
         fabric8.io/expose: "true"
         fabric8.io/ingress.path: /nuxeo/
         fabric8.io/ingress.name: preview
-        fabric8.io/ingress.annotations: "nginx.ingress.kubernetes.io/proxy-body-size: 0"
```
Gabez0r: Please don't forget about this.

elements/nuxeo-dropzone/nuxeo-dropzone.js
nmpcunha: Any reason for not consider the setter?

i18n/messages.json
```diff
@@ -535,6 +535,8 @@
   "driveSyncToggleButton.unsync": "Unsynchronize",
   "dropzone.abort": "Abort file upload",
   "dropzone.add": "Upload main file",
+  "dropzone.invalid.httpStatusCode.413": "File size exceeded",
```
nmpcunha: I understand the reason why, but it is still a technical key IMHO.

elements/nuxeo-dropzone/nuxeo-dropzone.js
```diff
@@ -491,4 +529,22 @@ Polymer({
...
+      return false;
+    }
+    if (!this.required) {
+      return true;
+    }
+    return this.files && this.files.length > 0;
+  },
+
+  _handleError(e) {
+    if (e.detail && e.detail.error && e.detail.error.response && e.detail.error.response.status) {
```
Gabez0r: I would expect to have something like:
```suggestion
    if (this.invalid && e.detail && e.detail.error && e.detail.error.response && e.detail.error.response.status) {
```
In the sense that there shouldn't be anything preventing me from submitting a form **if** the upload failed by whatever reason, and this is not a required field.

Is this a shortcut to display an error message, which would probably be similar to what we need to achieve under NXP-27982?

guirenard: >Is this a shortcut to display an error message, which would probably be similar to what we need to achieve under NXP-27982?

Not really, here an error may occur during batch upload is hapenning. Typically:
 - using s3 direct upload, it is s3 bucket that would return a potential error
 - deploying the web ui in its own nginx with implicit `body-size` limit (see https://github.com/nuxeo/nuxeo-web-ui/commit/3c31848a8a2daa58afa83211cca802860b2993ea)

All these use cases may return errors that cannot be handled like the validation report after a POST or a PUT on the document itself.

If such a scenario occurs, we need to notify the user the blob wasn't successfully uploaded and clicking `Save` or `Create` won't attach the blob to the document. So setting the dropzone invalid is a way to make sure:
 1- submit does not happen
 2- the form focuses on the dropzone element
 3- the user sees the upload failed and eventually why
 4- the user is incited to retry (e.g. with a smaller file if size exceeded), clear and submit without attachment, etc.

Gabez0r: Sorry, I didn't mean using form hooks. What I meant was that we are lacking a system to display error messages on fields, which IIRC, was what we initially wanted to do on [NXP-27982](https://jira.nuxeo.com/browse/NXP-27982) but couldn't, because we were also trying to do it leveraging validity. The validity is used to check for value format using constraints, but in the case of an error, the value remains as is.

Anw, I can live with this as a temporary shortcut, although this does not seem to be necessary to solve [NXP-27803](https://jira.nuxeo.com/browse/NXP-27803). But if indeed we'd like to go in the direction of having a way to display error messages on fields not related to validation, then I'd wait for it or at least add a comment stating that it should be cleaned up once that feature is implemented.

PR #747, opened 2019-10-09
NXP-27803: implement IronValidatableBehavior and mark invalid when fi…

guirenard:
…le upload pending

### DIFF
--- a/i18n/messages.json
+++ b/i18n/messages.json
@@ -474,6 +474,7 @@
   "driveSyncToggleButton.sync": "Synchronize",
   "driveSyncToggleButton.unsync": "Unsynchronize",
   "dropzone.add": "Upload main file",
+  "dropzone.invalid.uploading": "File upload pending",
   "dropzone.uploaded": "File uploaded",
   "dublincoreEdit.directorySuggestion.placeholder": "Select a value.",
   "easyshare.copy.label": "Copy the link and share the document {0}.",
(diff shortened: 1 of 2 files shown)


guirenard:
Pipeline OK https://qa2.nuxeo.org/jenkins/job/WebUI/job/web-ui-pipeline-10.10/217/

i18n/messages.json
```diff
@@ -474,6 +474,8 @@
   "driveSyncToggleButton.sync": "Synchronize",
   "driveSyncToggleButton.unsync": "Unsynchronize",
   "dropzone.add": "Upload main file",
+  "dropzone.invalid.413": "File size exceeded",
```
nmpcunha: nit: despite simplifying the code, which is cool, this is not a friendly key name.
Since the labels are contributed by the community, not sure if we should have a different key name rather than the html status code

Gabez0r:
I don't think we can make this error message customizable here, since it will be set by the element itself based on the error it gets from the server. But if we find it confusing with how we handle the other `errorMessage` properties, we could event not declare it as a property (or even use a different name), WDYT?

guirenard:
This is a true question as the next step would be to also leverage `errorMessage` property for the validation report.

It seems to me that `errorMessage` should be public property of which the value must be set each time the element turns invalid.

Note that we are not the only one to wonder this kind of questions see https://github.com/PolymerElements/paper-input/issues/292

i18n/messages.json
guirenard: I could maybe use `dropzone.invalid.httpStatusCode.413` as a label key to be more explicit and friendly. But having
>a different key name rather than the html status code

means hardcoding the translation for a limited list of HTTP sc which is not cool and flexible...

PR #756, opened 2019-10-16
NXP-28161: check for empty array params

guirenard:


elements/search/nuxeo-results-view.js
```diff
@@ -315,7 +315,19 @@ Polymer({
   _getFilterCount() {
     if (this._params) {
       // subtract the number of original parameters (this._paramsCount)
-      return Object.keys(this._params).length - this._paramsCount - ('highlight' in this._params ? 1 : 0);
+      return (
+        Object.keys(this._params).filter((p) => {
+          if (this._params[p]) {
```
Gabez0r: Shouldn't we be checking for
```suggestion
          if (this._params[p] != null) {
```
Instead? Empty strings are valid params, right?

elements/search/nuxeo-results-view.js
```diff
@@ -315,7 +315,19 @@ Polymer({
...
+            if (Array.isArray(this._params[p])) {
+              return this._params[p].length > 0;
+            }
+            return true;
+          }
+          return false;
+        }).length -
```
Gabez0r: nit: we could probably do this with a smaller block, something like this:
```
        Object.keys(this._params).filter(
          (p) => this._params[p] != null && (!Array.isArray(this._params[p]) || this._params[p].length > 0),
        ).length -
```

elements/search/nuxeo-results-view.js
guirenard: I don't think so, when doing so I end up with a badge showing `1` active param once I have manually cleared an input and it is not consistent with the result set.

elements/search/nuxeo-results-view.js
Gabez0r: After the above comment:
```
        Object.keys(this._params).filter(
          (p) => this._params[p] && (!Array.isArray(this._params[p]) || this._params[p].length > 0),
        ).length -
```

PR #761, opened 2019-10-23
NXP-28160: ensure widgets are in sync with layout's doc instance

nmpcunha:


addons/nuxeo-ai-core/index.js
```diff
@@ -115,13 +115,14 @@ const AISuggestionManager = (() => {
       }
       const aiModels = layout.document.contextParameters && layout.document.contextParameters.aiModels;
       const isModelInput = path && aiModels && aiModels.inputs.includes(path.replace('document.properties.', ''));
+      const noSpecificPropertyChanged = !path || (path && path.length <= 'document.properties'.length);
```
Gabez0r: It would be safer to actually check the path matches `document.properties` and not just its length:
```suggestion
      const noSpecificPropertyChanged = !path || 'document.properties'.startsWith(path);
```
nit: I'd even name name it `noPropertyChanged ` or `noDocPropertyChanged` since the "specific" doesn't add much.

PR #764, opened 2019-10-25
Feature nxp 27982 validation

guirenard:
PR ok https://qa.nuxeo.org/jenkins/job/master/job/web-ui-pipeline/292/


guirenard:
Requires:
 - https://github.com/nuxeo/nuxeo-elements/pull/171
 - https://github.com/nuxeo/nuxeo/pull/3489
 - https://github.com/nuxeo/nuxeo-jsf-ui/pull/18

elements/document/nuxeo-document-layout.js
```diff
@@ -136,4 +154,49 @@ Polymer({
...
+        this._addValidationError(this.i18n(v.messageKey));
+      }
+    });
+  },
+
+  _addValidationError(message) {
+    const label = document.createElement('span');
+    label.setAttribute("class", "error");
+    label.innerHTML = message;
+    this.$.error.append(label);
```
Gabez0r: Couldn't we do without an extra span and bind the text to #error, thus saving this extra DOM operation?

packages/nuxeo-web-ui-ftest/features/step_definitions/document.js
```diff
@@ -300,3 +300,9 @@ Then(/^I can unpublish the document$/, function() {
   unpublishConfirm.waitForVisible();
   unpublishConfirm.click();
 });
+
+Then('I can see {int} validation error(s) in the {string} edit form', function(nbErrors, docType) {
+  const { browser } = this.ui;
+  const form = browser.editForm(docType);
+  form.errorMessages.length.should.be.equals(nbErrors);
```
Gabez0r: Ii'd be safer to put a `form.waitForVisible();` before this line.

ftest/features/validation.feature
```diff
@@ -0,0 +1,33 @@
...
+      | secondNumber | 2                 |
+    Then I can see 1 validation error in the "Validation" edit form
+
+
+  Scenario: Field constraint validation violation
+    Given I can edit the following properties in the Validation metadata:
+      | name         | value             |
+      | title        | my title          |
+      | letterOnly   | abc12             |
+    Then I can see 1 validation error in the "Validation" edit form
```
Gabez0r: We are testing for a different case than the previous scenario, yet we're doing exactly the same assertion. Both tests are testing that we're able to display validations errors. Perhaps we should have just one of them, and create a ticket to test the rest unitarily? Or, should we chose to keep both, perhaps we should try to have a proper assert for the global validation and constraint validation cases?

elements/document/nuxeo-document-layout.js
guirenard: The thing is we may have several validation error message at once, I am not sure what you suggest takes it into account.

ftest/features/validation.feature
guirenard: Ideally,  I would add assertion on the label but we don't want to make an assumption on the browser locale, right?
Alternatively, we could add the label key as the error message dom node's data?

elements/document/nuxeo-document-layout.js
Gabez0r: I was thinking that if we were to use a private string property, we could just append the different error messages there, but I understand that having spans here is convenient for denote different validation errors. I'm ok it! 👍

ftest/features/validation.feature
nmpcunha: Indeed.
I would say that all of those validations can be tested unitarily. (i.e. as soon as we move to nuxeo-elements or introduce tests on WebUI project)

i18n/messages.json
```diff
@@ -705,6 +705,43 @@
...
+  "label.schema.constraint.violation.double": "The value '{0}' must be a decimal number.",
+  "label.schema.constraint.violation.integer": "The value '{0}' must be an integer.",
+  "label.schema.constraint.violation.long": "The value '{0}' must be an integer.",
+  "label.schema.constraint.violation.string": "The value '{0}' must be a text.",
+  "label.schema.constraint.violation.userManagerResolver.any": "'{0}' is not a valid username or a valid group name.",
+  "label.schema.constraint.violation.userManagerResolver.group": "'{0}' is not a valid group name.",
+  "label.schema.constraint.violation.userManagerResolver.user": "'{0}' is not a valid username.",
+  "label.schema.constraint.violation.directoryResolver": "'{0}' is not a valid {1} id.",
+  "label.schema.constraint.violation.documentResolver.id": "'{0}' is not a valid document id.",
+  "label.schema.constraint.violation.documentResolver.path": "'{0}' is not a valid document path.",
```
nmpcunha: Can my custom validation service define custom violation labels?

packages/nuxeo-web-ui-ftest/features/step_definitions/document.js
nmpcunha: Should we add also a new line on the end of the file?

ftest/features/validation.feature
Gabez0r: We assume the local at the feature level (thing of the dates, for example, which need to be "Month day, year". So perhaps we could have a step that accepts the error message as a param? 🤔

i18n/messages.json
guirenard: You may throw a global server validation exception with a custom label key like in https://github.com/nuxeo/nuxeo/commit/34547eecce8b4e2dfa5735d895cb6ab7b0310e51#diff-60b6e86268c3c3f84241be4fa84330cdR53 in which case you must define this new label value.

Else for field constraint violation, you may define new ObjectResolver which may return new label key through `org.nuxeo.ecm.core.model.DocumentModelResolver.getConstraintErrorMessage(Object, Locale)` and you'll then have to define your label here too.

Finally, for existing schema constraints, the rule is, for a given field `validation:letterOnly` with a regex constraint like in https://github.com/nuxeo/nuxeo-web-ui/pull/764/files#diff-08a5ea697f8cc28864df989f28183a8bR35, we'll first try to find a label with key `label.schema.constraint.violation.PatternConstraint.validation.letterOnly` if not found, we fall back on `label.schema.constraint.violation.PatternConstraint`

elements/document/nuxeo-document-layout.js
```diff
@@ -140,4 +153,54 @@ Polymer({
         });
     }
   },
+
+  reportValidation(validationReport) {
+    this._resetValidationErrors();
+    validationReport.violations.reverse().forEach((v) => {
```
nmpcunha: nitpick: `v` its cool, but not the most easy name to read and contextualize :)
same for `p`

elements/document/nuxeo-document-layout.js
```diff
@@ -136,4 +153,54 @@ Polymer({
...
+      this.invalid = true;
+      if (v.path) {
+        v.path.forEach((p) => {
+          const widgets = this._getBoundElements(`document.properties.${p.field_name}`);
+          if (widgets) {
+            const msg = this.i18n(v.messageKey, v.invalid_value, p.field_name);
+            if (msg === v.messageKey && v.constraint && v.constraint.name) {
```
nmpcunha: Could this be reduced to `v.constraint && v.constraint.name`? Or do we need to confirm that the translation is not available and it returns the exact same key?

elements/document/nuxeo-document-form-layout.js
```diff
@@ -96,23 +108,15 @@ Polymer({
     headers: {
       type: Object,
     },
+
+    saving: {
+      type: Boolean,
```
nmpcunha: Should this be readonly?

elements/document/nuxeo-document-form-layout.js
```diff
@@ -128,17 +132,43 @@ Polymer({
...
+    }
+    this.set('saving', true);
+    Promise.all(validations)
+      .then(() => this._doSave())
+      .then(() => {
+        this.set('saving', false);
+        this._refresh(this);
+      })
+      .catch((err) => {
+        this.set('saving', false);
```
nmpcunha: Should the set to false be part of a `finally`?

elements/document/nuxeo-document-create.js
```diff
@@ -333,28 +333,38 @@ Polymer({
...
-  },
-
-  _create() {
-    if (!this._validate() || !this.canCreate) {
+    const elementsToValidate = innerLayout._getValidatableElements(innerLayout.element.root);
+    elementsToValidate.push(innerLayout.element);
+    let valid = true;
+    const validations = [];
+    elementsToValidate.forEach((el) => {
+      if (el.validate) {
```
nmpcunha: Why are we checking this again? The `_getValidatableElements` already checks, correct?
Is it because `innerLayout.element`?

elements/document/nuxeo-document-create.js
```diff
@@ -333,28 +333,38 @@ Polymer({
...
+      }
+    });
+    if (!valid) {
+      const invalidField = elementsToValidate.find((node) => node.invalid);
+      invalidField.scrollIntoView();
+      invalidField.focus();
       return;
     }
-    this.document.name = this.document.name || this._sanitizeName(this.document.properties['dc:title']);
     this.set('creating', true);
```
nmpcunha: Despite not being the scope of this ticket, should this property be read only?

elements/document/nuxeo-document-form-layout.js
```diff
@@ -128,13 +132,43 @@ Polymer({
...
 
   save() {
-    if (!this._validate()) {
+    const { layout } = this.$.layout.$;
+    const elementsToValidate = layout._getValidatableElements(layout.element.root);
+    elementsToValidate.push(layout.element);
+    let valid = true;
+    const validations = [];
+    elementsToValidate.forEach((el) => {
+      if (el.validate) {
```
nmpcunha: Same question as https://github.com/nuxeo/nuxeo-web-ui/pull/764/files#r345891444

packages/nuxeo-web-ui-ftest/features/step_definitions/document.js
```diff
@@ -300,3 +300,17 @@ Then(/^I can unpublish the document$/, function() {
   unpublishConfirm.waitForVisible();
   unpublishConfirm.click();
 });
+
+Then('I can see {int} validation error(s) in the {string} edit form', function(nbErrors, docType) {
+  const { browser } = this.ui;
+  const form = browser.editForm(docType);
+  form.waitForVisible();
+  form.errorMessages.length.should.be.equals(nbErrors);
```
Gabez0r: Since we are not doing a waitForVisible on the `'#error .error'` and there might be a delay showing the error, we should probably do this inside a waitUntil, or we might end up with "random" failures.

packages/nuxeo-web-ui-ftest/features/step_definitions/document.js
```diff
@@ -300,3 +300,17 @@ Then(/^I can unpublish the document$/, function() {
...
+});
+
+Then('I can see the {string} error message in the {string} edit form', function(message, docType) {
+  const { browser } = this.ui;
+  const form = browser.editForm(docType);
+  form.waitForVisible();
+  form.errorMessages.some((err) => err === message);
```
Gabez0r: Same as above, assuming step definitions are independent.

elements/document/nuxeo-document-create.js
Gabez0r: If we're dealing with native elements, they don't have a `validate` method but they do have the standard `checkValidity`, which used to be called with `form.validate()`. This means native elements won't be taken into account.

elements/document/nuxeo-document-create.js
```diff
@@ -333,28 +333,38 @@ Polymer({
...
-  _create() {
-    if (!this._validate() || !this.canCreate) {
+    const elementsToValidate = innerLayout._getValidatableElements(innerLayout.element.root);
+    elementsToValidate.push(innerLayout.element);
+    let valid = true;
+    const validations = [];
+    elementsToValidate.forEach((el) => {
+      if (el.validate) {
+        const elValidate = el.validate();
+        if (typeof elValidate.then === 'function') {
```
Gabez0r: This means that we will be checking for async validate even on elements, but we don't have such a use case, do we? We could just do a check if `innerLayout.element.validate` returns a promise or not and check if value before proceeding, right?

elements/document/nuxeo-document-create.js
```diff
@@ -333,28 +333,38 @@ Polymer({
...
+      invalidField.focus();
       return;
     }
-    this.document.name = this.document.name || this._sanitizeName(this.document.properties['dc:title']);
     this.set('creating', true);
-    this.$.docRequest
-      .post()
+    Promise.all(validations)
+      .then(() => {
+        this.document.name = this.document.name || this._sanitizeName(this.document.properties['dc:title']);
```
Gabez0r: If any of the validations returns false, this will still do the post, but it shouldn't.

elements/document/nuxeo-document-create.js
Gabez0r: `_getValidatableElements` also returns native elements, which do not expose a validate method.

elements/document/nuxeo-document-layout.js
guirenard: >Or do we need to confirm that the translation is not available and it returns the exact same key?

Yes, as explained in https://github.com/nuxeo/nuxeo-web-ui/pull/764#discussion_r339622029
>for existing schema constraints, the rule is, for a given field validation:letterOnly with a regex constraint like in https://github.com/nuxeo/nuxeo-web-ui/pull/764/files#diff-08a5ea697f8cc28864df989f28183a8bR35, we'll first try to find a label with key label.schema.constraint.violation.PatternConstraint.validation.letterOnly if not found, we fall back on label.schema.constraint.violation.PatternConstraint

elements/document/nuxeo-document-create.js
guirenard: If the `validate` returns a promise, the idea is that it is resolved -> valid, rejected -> invalid

guirenard:
Depends on https://github.com/nuxeo/nuxeo-elements/pull/183

guirenard:
https://qa.nuxeo.org/jenkins/job/Master/job/web-ui-pipeline/307/ OK

elements/document/nuxeo-document-layout.js
```diff
@@ -136,4 +154,54 @@ Polymer({
...
+      } else {
+        this._addValidationError(this.i18n(violation.messageKey));
+      }
+    });
+  },
+
+  _addValidationError(message) {
+    const label = document.createElement('span');
+    label.setAttribute('class', 'error');
+    label.innerHTML = message;
```
Gabez0r: I guess we should be using `escapeHTML` here. Perhaps we should even move it to the new `utils.js`file to be introduced in the scope of ELEMENTS-1082, to prevent having to import it from selectivity.

elements/workflow/nuxeo-document-task.js
```diff
@@ -207,38 +220,49 @@ Polymer({
     }
   },
 
-  validate() {
-    return this.$.layout.validate();
+  async validate() {
+    const layoutValidate = await this.$.layout.validate();
```
Gabez0r: We should be able to return `this.$.layout.validate()` right away.

elements/document/nuxeo-document-layout.js
guirenard: I thought about it but did not escape anything since the `message` cannot be valued dynamically (it comes from server-side java code). Anw, I replaced the code to use polymer binding to benefit xss protection.

elements/document/nuxeo-document-form-layout.js
```diff
@@ -127,14 +132,33 @@ Polymer({
     return this.$.doc.put();
   },
 
-  save() {
-    if (!this._validate()) {
+  async save() {
```
nmpcunha: Should this method be "private"? Or do we need it "public"?

nelsonsilva: method could / should be "private" meaning we could merge `_doSave()` and `save()` to end up with only the former

PR #767, opened 2019-10-30
NXP-27968: create CSV import ftest

FabioFSantos:
https://qa.nuxeo.org/jenkins/view/TestAndPush/job/web-ui-pipeline-fdsantos/59/console 🎉

### DIFF
--- a/addons/nuxeo-csv/ftest/features/csv-import.feature
+++ b/addons/nuxeo-csv/ftest/features/csv-import.feature
@@ -1,0 +1,22 @@
+Feature: CSV Import
+
+  Scenario: I can import a CSV file and navigate to the created documents
+    Given user "John" exists in group "members"
+    And I login as "John"
+    And I have a Workspace document
+    And I have permission ReadWrite for this document
+    And I browse to the document
+
+    When I click the Create Document button
+    And I go to the importCSV tab
+    And I import the csv-import-sample.csv file
+    Then I can see the "my_folder" child document is at position "1"
+    When I navigate to "my_folder" child
+    Then I can see the "file1" child document is at position "1"
+    When I navigate to "file1" child
+    Then I can see File metadata with the following properties:
+      | name         | value                                 |
+      | title        | file1                                 |
+      | description  | doc in folder                         |
+      | nature       | Letter                                |
+      | subjects     | Daily life/Gastronomy,Society/Company |
--- a/addons/nuxeo-csv/ftest/features/step_definitions/csv-import.js
+++ b/addons/nuxeo-csv/ftest/features/step_definitions/csv-import.js
@@ -1,0 +1,18 @@
+import { When } from 'cucumber';
+
+When(/^I go to the (.+) tab$/, function(name) {
+  const dialog = this.ui.createDialog;
+  dialog.waitForVisible();
+  dialog.importTab(name).click();
+});
+
+When(/^I import the (.+) file$/, function(file) {
+  const dialog = this.ui.createDialog;
+  dialog.waitForVisible();
+  dialog.setFileToImport(file);
+  dialog.selectedFileToImport.waitForVisible().should.be.true;
+  dialog.importCsvButton.click();
+  dialog.importSuccess.waitForVisible();
+  dialog.importError.isVisible().should.be.false;
+  dialog.importCloseButton.click();
+});
--- a/packages/nuxeo-web-ui-ftest/pages/ui/create_dialog.js
+++ b/packages/nuxeo-web-ui-ftest/pages/ui/create_dialog.js
@@ -1,9 +1,44 @@
+import path from 'path';
 import BasePage from '../base';
 import DocumentCreate from './browser/document_create';
 
 export default class CreateDialog extends BasePage {
   get documentCreate() {
     return new DocumentCreate('nuxeo-document-create');
+  }
+
+  importTab(name) {
+    return this.el.element(`paper-tab[name="${name}"]`);
+  }
+
+  get importCsvDialog() {
+    return this.el.element('#csvCreation');
+  }
+
+  setFileToImport(file) {
+    const field = this.importCsvDialog.element('#dropzone #uploadFiles');
+    field.waitForExist();
+    return field.chooseFile(path.resolve(fixtures.blobs.get(file)));
+  }
+
+  get importCsvButton() {
+    return this.importCsvDialog.element('div[name="upload"] paper-button.primary');
+  }
+
+  get selectedFileToImport() {
+    return this.el.element('#dropzone div.complete');
+  }
+
+  get importCloseButton() {
+    return this.el.element('div[name="progress"] paper-button.primary');
+  }
+
+  get importError() {
+    return this.el.element('div[name="progress"] #list div.item');
+  }
+
+  get importSuccess() {
+    return this.el.element('#progress div.successful');
   }
 
   get createButton() {
--- a/addons/nuxeo-csv/ftest/resources/csv-import-sample.csv
+++ b/addons/nuxeo-csv/ftest/resources/csv-import-sample.csv
@@ -1,0 +1,3 @@
+"name","type","dc:title","dc:description","dc:format","dc:source","dc:language","dc:nature","dc:subjects"
+"my_folder","Folder","my_folder","This is the description of a folder that will hold some documents",,"Nuxeo CSV import",,,
+"my_folder/file1","File","file1","doc in folder",,"Nuxeo CSV Import","English","letter","daily life/gastronomy|society/company"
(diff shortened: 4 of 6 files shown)


FabioFSantos:
https://qa.nuxeo.org/jenkins/view/TestAndPush/job/web-ui-pipeline-fdsantos/60/console 🎉

addons/nuxeo-csv/ftest/features/csv-import.feature
```diff
@@ -0,0 +1,23 @@
...
+  Background:
+    Given user "John" exists in group "members"
+    And I login as "John"
+    And I have a Workspace document
+    And I have permission ReadWrite for this document
+    And I browse to the document
+
+  Scenario: I can import a CSV file and navigate to the created documents
+    Given I click the Create Document button
+    And I go to importCSV tab and upload the csv-import-sample.csv
```
nelsonsilva: since this step is pretty specific to the import I don't think it's worth having to pass the tab name. Just something like `And I import the csv-import-sample.csv CSV file`.
Alternatively we could split it into 2 more generic steps, ie:
```
And i go to the importCSV creation tab
And I drop the csv-import-sample.csv in the .... dropzone
```

addons/nuxeo-csv/ftest/features/csv-import.feature
```diff
@@ -0,0 +1,23 @@
...
+    Then I import the CSV successfully
```
nelsonsilva: The `then` makes it sound like it's just an assertion when in fact it's the step that's clicking the button to perform the import... perhaps something like:
```
When I perform the CSV import
And I navigate to ...
```

packages/nuxeo-web-ui-ftest/pages/ui/create_dialog.js
```diff
@@ -6,6 +6,36 @@ export default class CreateDialog extends BasePage {
...
+  }
+
+  get importCsvDialog() {
+    return this.el.element('#csvCreation');
+  }
+
+  setFileToImport(value) {
+    const field = this.importCsvDialog;
+    field.waitForVisible();
+    return fixtures.layouts.setValue(field, value);
```
nelsonsilva: this is not a layout and `nuxeo-document-import-csv` is not a widget so you should move the logics you have to set / get values here. 
the `fieldRegistry`  is only relevant for widgets / elemets used in layouts so we can easily get / set property values without having to care about the widget being used.

addons/nuxeo-csv/ftest/features/csv-import.feature
FabioFSantos: I think the best option will be split since we have the import(not csv) tab, that with this logic could be used for future tests

FabioFSantos: b593c94a85b78af2dfe6663a5865adab714d31c9

addons/nuxeo-csv/ftest/features/csv-import.feature
FabioFSantos: b593c94a85b78af2dfe6663a5865adab714d31c9

packages/nuxeo-web-ui-ftest/pages/ui/create_dialog.js
FabioFSantos: b593c94a85b78af2dfe6663a5865adab714d31c9

addons/nuxeo-csv/ftest/features/csv-import.feature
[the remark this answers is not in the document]
FabioFSantos: b593c94a85b78af2dfe6663a5865adab714d31c9

addons/nuxeo-csv/ftest/features/csv-import.feature
```diff
@@ -0,0 +1,24 @@
...
+  Background:
+    Given user "John" exists in group "members"
+    And I login as "John"
+    And I have a Workspace document
+    And I have permission ReadWrite for this document
+    And I browse to the document
+
+  Scenario: I can import a CSV file and navigate to the created documents
+    Given I click the Create Document button
+    And I go to the importCSV creation tab
```
nmpcunha: nit: `I go to the X tab`
I would suggest remove the `creation` because we already have the context `Create Document button`

addons/nuxeo-csv/ftest/features/step_definitions/csv-import.js
```diff
@@ -0,0 +1,23 @@
...
+  dialog.selectedFileToImport.waitForVisible().should.be.true;
+});
+
+Then('I perform the CSV import', function() {
+  const dialog = this.ui.createDialog;
+  dialog.waitForVisible();
+  dialog.importCsvButton.click();
+  dialog.importSuccess.waitForVisible();
+  const success = dialog.importSuccess.getText();
+  success.should.be.equals('Successfully imported entries\n2 / 2');
```
nmpcunha: Not sure if this step should this validation. If yes, I would say that we may want to pass the number of entries as a parameter in order to avoid an hardcoded value. Or maybe consider to split it!

FabioFSantos:
the commits will be squashed after all the reviews are made

addons/nuxeo-csv/ftest/features/csv-import.feature
FabioFSantos: done 👍

addons/nuxeo-csv/ftest/features/step_definitions/csv-import.js
FabioFSantos: done 👍

packages/nuxeo-web-ui-ftest/pages/ui/create_dialog.js
```diff
@@ -6,6 +7,41 @@ export default class CreateDialog extends BasePage {
...
+  }
+
+  get importCsvDialog() {
+    return this.el.element('#csvCreation');
+  }
+
+  setFileToImport(file) {
+    const field = this.importCsvDialog.element('#dropzone #uploadFiles');
+    field.getValue();
+    field.waitForExist();
```
nmpcunha: Should this be executed before the `field.getValue()`?

FabioFSantos: done

addons/nuxeo-csv/ftest/features/csv-import.feature
```diff
@@ -0,0 +1,23 @@
...
+    Given user "John" exists in group "members"
+    And I login as "John"
+    And I have a Workspace document
+    And I have permission ReadWrite for this document
+    And I browse to the document
+
+  Scenario: I can import a CSV file and navigate to the created documents
+    Given I click the Create Document button
+    And I go to the importCSV tab
+    And I import the csv-import-sample.csv file
```
nmpcunha: nit: Maybe it may make sense to replace `And` by `When`

nmpcunha: Despite this small thing, it looks good! 👍 
Thanks!

addons/nuxeo-csv/ftest/features/csv-import.feature
```diff
@@ -0,0 +1,23 @@
...
+
+  Background:
+    Given user "John" exists in group "members"
+    And I login as "John"
+    And I have a Workspace document
+    And I have permission ReadWrite for this document
+    And I browse to the document
+
+  Scenario: I can import a CSV file and navigate to the created documents
+    Given I click the Create Document button
```
Gabez0r: "When" is more suitable here, since `Given` is mostly used to put the system in a known state before the user starts interacting with the it (see [here](https://cucumber.io/docs/gherkin/reference/#given)).
```suggestion
    When I click the Create Document button
```

Good job! 👏 💪

FabioFSantos: 👍

addons/nuxeo-csv/ftest/features/step_definitions/csv-import.js
```diff
@@ -0,0 +1,18 @@
+import { Then } from 'cucumber';
+
+Then(/^I go to the (.+) tab$/, function(name) {
```
nmpcunha: For consistency, we may want to turn this ones to `When`

packages/nuxeo-web-ui-ftest/pages/ui/create_dialog.js
```diff
@@ -6,6 +7,40 @@ export default class CreateDialog extends BasePage {
...
+  importTab(name) {
+    return this.el.element(`paper-tab[name="${name}"]`);
+  }
+
+  get importCsvDialog() {
+    return this.el.element('#csvCreation');
+  }
+
+  setFileToImport(file) {
+    const field = this.importCsvDialog.element('#dropzone #uploadFiles');
```
nelsonsilva: nit: ids are unique in the current global scope (in this case the element's shadow dom) so having a chain of id selectors only makes sense for readibility

PR #768, opened 2019-11-04
NXP-28263: dropzone handling tasks

nmpcunha:
Probably I'll squash both commits with the message on the title!

### DIFF
--- a/elements/nuxeo-document-attachments/nuxeo-document-attachments.js
+++ b/elements/nuxeo-document-attachments/nuxeo-document-attachments.js
@@ -16,8 +16,9 @@ limitations under the License.
 */
 import '@polymer/polymer/polymer-legacy.js';
 
+import '@nuxeo/nuxeo-elements/nuxeo-document.js';
 import { FiltersBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-filters-behavior.js';
-import { I18nBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-i18n-behavior.js';
+import { FormatBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-format-behavior.js';
 import '../nuxeo-document-blob/nuxeo-document-blob.js';
 import '../nuxeo-dropzone/nuxeo-dropzone.js';
 import { Polymer } from '@polymer/polymer/lib/legacy/polymer-fn.js';
@@ -51,6 +52,8 @@ Polymer({
       }
     </style>
 
+    <nuxeo-document id="doc" doc-id="[[document.uid]]"></nuxeo-document>
+
     <template is="dom-if" if="[[_isAvailable(document, xpath)]]">
       <h3>[[i18n('documentAttachments.heading')]]</h3>
 
@@ -67,13 +70,12 @@ Polymer({
 
       <template is="dom-if" if="[[_hasWritePermission(document)]]">
         <nuxeo-dropzone
-          document="{{document}}"
-          xpath="[[xpath]]"
+          value="{{document.properties.files:files}}"
+          multiple
+          value-key="file"
           uploaded-message="[[i18n('documentAttachments.upload.uploaded')]]"
           message="[[i18n('documentAttachments.upload.add')]]"
           drag-content-message="[[i18n('documentAttachments.upload.drop')]]"
-          blob-list
-          update-document
         >
         </nuxeo-dropzone>
       </template>
@@ -81,7 +83,7 @@ Polymer({
   `,
 
   is: 'nuxeo-document-attachments',
-  behaviors: [I18nBehavior, FiltersBehavior],
+  behaviors: [FormatBehavior, FiltersBehavior],
 
   properties: {
     document: Object,
@@ -90,6 +92,32 @@ Polymer({
       type: String,
       value: 'files:files',
     },
+  },
+
+  created() {
+    this._createMethodObserver('_valueChanged(document.properties.files:files.splices)', true);
+  },
+
+  _valueChanged(e) {
+    if (!e) {
+      return;
+    }
+    const props = {};
+    const formattedXpath = this.formatPropertyXpath(this.xpath);
+    this._createNestedObjectRecursive(props, formattedXpath.split('.'));
+    this.set(formattedXpath, this.get(formattedXpath, this.document.properties), props);
+    this.$.doc.data = {
+      'entity-type': 'document',
+      repository: this.document.repository,
+      uid: this.document.uid,
+      properties: props,
+    };
+
+    this.$.doc.put().then((response) => {
+      this.document = response;
+      this.fire('notify', { message: this.i18n(this.uploadedMessage) });
+      this.fire('document-updated');
+    });
   },
 
   _hasFiles(doc) {
@@ -118,5 +146,28 @@ Polymer({
 
   _isAvailable(document, xpath) {
     return document && xpath && this.hasSchema(document, xpath.split(':')[0]);
+  },
+
+  /**
+   * Recursive method to create nested objects when they don't exist in a parent object.
+   * It does not change any other existing objects or inner objects, only the ones referred in 'path'.
+   * @param obj Parent Object where inner nested objects should be created.
+   * @param path Array containing the inner object keys.
+   * Usage Example:
+   *
+   *  - Creating document properties using xpath:
+   *
+   *    const xpath = 'my:custom/field/subfield/x'
+   *    _createNestedObjectRecursive(this.document.properties, xpath.split('/'));
+   *
+   */
+  _createNestedObjectRecursive(obj, path) {
+    if (path.length === 0) {
+      return;
+    }
+    if ((!Object.prototype.hasOwnProperty.call(obj, path[0]) && !obj[path[0]]) || typeof obj[path[0]] !== 'object') {
+      obj[path[0]] = {};
+    }
+    return this._createNestedObjectRecursive(obj[path[0]], path.slice(1));
   },
 });
--- a/elements/nuxeo-dropzone/nuxeo-dropzone.js
+++ b/elements/nuxeo-dropzone/nuxeo-dropzone.js
@@ -384,11 +384,20 @@ Polymer({
   },
 
   _reset(value) {
-    if (value === undefined) {
-      this.cancelBatch();
+    if (
+      value == null ||
+      (Array.isArray(value) &&
+        value.filter(
+          (file) => !Object.prototype.hasOwnProperty.call(this.valueKey ? file[this.valueKey] : file, 'data'),
+        ).length === 0) ||
+      Object.prototype.hasOwnProperty.call(value, 'data')
+    ) {
+      if (this.uploading) {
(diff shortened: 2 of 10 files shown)


nmpcunha:
@guirenard , the guideline here was to minimize the number of changes on the API, despite being an improvement for master.
Anyway, not feeding the discussion about which API is better designed, `nuxeo-dropzone` has some "strings attached"... the property `updateDocument` only works when the entity-type is document, since we rely on `nuxeo-document`.

Given this, I'm not sure how we should uniformize both APIs. Mainly, because it may imply some code changes from `nuxeo-dropzpone` to `nuxeo-attachments` (e.g. `updateDocument` property).

guirenard:
>it may imply some code changes from `nuxeo-dropzpone` to `nuxeo-attachments`

True but we'd nee some backward compat anyway (and yes,  it is going to be hard to maintain, but we could log a deprecation warning).

Otherwise, I am wondering if another pair of `task` / `variable` properties (vs. `document` / `xpath`) wouldn't be cleaner and better.

In any case, setting the `document` property with a `task` entity seems quite awkward to me.

elements/nuxeo-dropzone/nuxeo-dropzone.js
```diff
@@ -183,6 +183,7 @@ Polymer({
   properties: {
     /**
      * Input Document.
+     * It's entity type can be 'document' or 'task'
```
nelsonsilva:
```suggestion
     * Its entity type can be 'document' or 'task'
```

elements/nuxeo-dropzone/nuxeo-dropzone.js
```diff
@@ -183,6 +183,7 @@ Polymer({
...
      */
     document: {
```
nelsonsilva: Introducing another `task` property would make it a bit more explicit that we support tasks and make it cleaner imho.

elements/nuxeo-dropzone/nuxeo-dropzone.js
```diff
@@ -220,6 +222,7 @@ Polymer({
     },
     /**
      * This flag determines whether the file should be immediately uploaded or not.
+     * It is only applicable for 'document' entity type.
```
nelsonsilva: This can be removed if we go with a `task` property

elements/nuxeo-dropzone/nuxeo-dropzone.js
```diff
@@ -302,29 +305,32 @@ Polymer({
     this.$$('input').click();
   },
 
-  _computeParsedXpath(xpath) {
-    return this.formatPropertyXpath(xpath);
+  _computeParsedXpath(xpath, document) {
```
nelsonsilva: this is no longer a valid xpath... it's a property path

elements/nuxeo-dropzone/nuxeo-dropzone.js
nmpcunha: But still only applicable for documents and not tasks.

nelsonsilva:
> I was hoping about having an API like [nuxeo-file widget](https://github.com/nuxeo/nuxeo-elements/blob/master/ui/widgets/nuxeo-file.js#L175) where we have a `value` property that would receive the full path i.e.
> 
> * `<nuxeo-dropzone value="{{task.variables.blobs}}" multiple></nuxeo-dropzone>`
> * `<nuxeo-dropzone value="{{document.properties.file:content}}"></nuxeo-dropzone>`
> 
> Interesting related ticket https://jira.nuxeo.com/browse/ELEMENTS-748

Although I agree it would be nice to also support a "pure widget mode" where you can simply bind to a blob property IIRC this element had to be bound to a document at some point. Perhaps since we've dropped usage of automation (see ie https://github.com/nuxeo/nuxeo-web-ui/commit/284824437e43ef02a8d41f7b839c5e00a59d62ea#diff-08050d080c2454da8a97325053ae0241) we can actually do that 🤔 Not sure it'd be in the context of this ticket but worth having a look for sure

elements/nuxeo-document-attachments/nuxeo-document-attachments.js
```diff
@@ -67,13 +69,12 @@ Polymer({
 
       <template is="dom-if" if="[[_hasWritePermission(document)]]">
         <nuxeo-dropzone
-          document="{{document}}"
+          value="{{document.properties.files:files}}"
+          multiple
           xpath="[[xpath]]"
```
guirenard: How `xpath` is still needed/used here?

nmpcunha: Regarding `xpath`: if you take a look it is only used because we have a different object structure for `files:files` property.  I'll take a look on server side, but I'm pretty sure that we may not be able to change that structure.
What can be done, is to have a flag to do that magic and consider a different object structure. (https://github.com/nuxeo/nuxeo-web-ui/blob/improvement-NXP-28263-DropzoneHandlingTasks/elements/nuxeo-dropzone/nuxeo-dropzone.js#L353)

elements/nuxeo-dropzone/nuxeo-dropzone.js
```diff
@@ -337,106 +312,89 @@ Polymer({
...
+          };
+        } else {
+          uploadedFile = {
+            'upload-batch': data.detail.batchId,
+            'upload-fileId': index.toString(),
+          };
+        }
+        files.push(
           // Handle special case when using files:files
           this.xpath === 'files:files' ? { file: uploadedFile } : uploadedFile,
```
Gabez0r: Since the only purpose of `xpath` is checking if it's `files:files` or not, we could even use a boolean property, no? We could even encapsulate this logic on a function, which could be overridden from a parent element, according to its own `xpath`. This way we wouldn't probably need to expose the xpath on `nuxeo-dropzone`.

elements/nuxeo-document-attachments/nuxeo-document-attachments.js
```diff
@@ -92,6 +93,30 @@ Polymer({
...
+  created() {
+    this._createMethodObserver('_valueChanged(document.properties.files:files.splices)', true);
+  },
+
+  _valueChanged(e) {
+    if (!e) {
+      return;
+    }
+    const props = {};
+    props['files:files'] = this.document.properties['files:files'];
```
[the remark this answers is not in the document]
nmpcunha: Yes, of course. That's also why this PR was `wip` tag.
Just want to validate the concept, API and ideas. But thanks for noticing that ;)

elements/nuxeo-dropzone/nuxeo-dropzone.js
nmpcunha: It was one of the possibilities to overcome this `files:files` question, as discussed here (https://github.com/nuxeo/nuxeo-web-ui/pull/768#discussion_r345680569)

guirenard:
Could we please also consider ```diff --git i/elements/nuxeo-dropzone/nuxeo-dropzone.js w/elements/nuxeo-dropzone/nuxeo-dropzone.js
index f970e7d2..b52a1059 100644
--- i/elements/nuxeo-dropzone/nuxeo-dropzone.js
+++ w/elements/nuxeo-dropzone/nuxeo-dropzone.js
@@ -154,7 +154,7 @@ Polymer({
       <label id="label" required$="[[required]]">[[label]]</label>
     </template>
 
-    <input hidden id="input" type="file" on-change="_uploadInputFiles" />
+    <input hidden id="input" type="file" on-change="_uploadInputFiles" multiple$="[[multiple]]"/>
 
     <div id="details" hidden$="[[!hasFiles]]">
       <template is="dom-repeat" items="[[files]]" as="file">
```

File: elements/nuxeo-document-attachments/nuxeo-document-attachments.js [hunk 8]
```diff
@@ -92,6 +94,30 @@ Polymer({
...
+  created() {
+    this._createMethodObserver('_valueChanged(document.properties.files:files.splices)', true);
+  },
+
+  _valueChanged(e) {
+    if (!e) {
+      return;
+    }
+    const props = {};
+    props[this.xpath] = this.document.properties[this.xpath];
```
Gabez0r: Assuming `xpath` can be multilevel, we should probably format it first and initialize the props object before setting it.

elements/nuxeo-dropzone/nuxeo-dropzone.js
```diff
@@ -337,106 +310,94 @@ Polymer({
...
+  _reset(value) {
+    if (
+      value == null ||
+      (Array.isArray(value) &&
+        value.filter(
+          (file) => !Object.prototype.hasOwnProperty.call(this.valueKey ? file[this.valueKey] : file, 'data'),
+        ).length === 0) ||
+      Object.prototype.hasOwnProperty.call(value, 'data')
+    ) {
+      if (this.uploading) {
```
Gabez0r: Not sure if is safe to assume that we should only cancel the batch if we're still uploading. I know we had to do it when the form was canceled in the past, but because there is no actual way to be sure when that happens, we leaned on `document === undefined` (see https://github.com/nuxeo/nuxeo-web-ui/commit/d445ffc69b1d5db3d8f1a2de86da46cd72718820). We must make sure that we don't cause a regression for [NXP-24979](https://jira.nuxeo.com/browse/NXP-24979) and [ELEMENTS-845](https://jira.nuxeo.com/browse/ELEMENTS-845) (fixed under NXP-24979).

elements/nuxeo-document-attachments/nuxeo-document-attachments.js
nmpcunha: Indeed.
Probably we can think to tackle this one https://jira.nuxeo.com/browse/ELEMENTS-1082

elements/nuxeo-dropzone/nuxeo-dropzone.js
nmpcunha: @guirenard do you have any insight about this?

guirenard: Since you are only trying to `cancelBatch` when there's an upload pending, I think it is ok. As stated in https://jira.nuxeo.com/browse/ELEMENTS-845

In any case, we can (easily) test with s3 direct upload enabled and observe the network dev tool panel that a pending upload is aborted when clicking cancel in a edit or create form

elements/nuxeo-document-attachments/nuxeo-document-attachments.js
```diff
@@ -92,6 +93,30 @@ Polymer({
...
+    props['files:files'] = this.document.properties['files:files'];
+    this.$.doc.data = {
+      'entity-type': 'document',
+      repository: this.document.repository,
+      uid: this.document.uid,
+      properties: props,
+    };
+
+    this.$.doc.put().then((response) => {
+      this.document = response;
```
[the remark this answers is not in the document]
nmpcunha: In theory it won't introduce any regression since we are firing a `document-updated` and it is fetching the document with the proper enrichers.
Not sure why it was added in first place, but at that time we may not had the global enrichers configuration containing by default `appLinks`

PR #782, opened 2019-11-25
NXP-28312 10.10 backport

michaelva:
NXP-28312: remove data list min-height in nuxeo-clipboard


nmpcunha:
Thanks for you PR.
Since this is a backport from https://github.com/nuxeo/nuxeo-web-ui/pull/779, we would expect both commits to have the same message.

michaelva:
@nmpcunha done

PR #786, opened 2019-11-26
NXP-28114: ping batch upload endpoint to avoid http timeout

guirenard:


plugin/web-ui/addon/src/main/resources/OSGI-INF/web-ui-properties.xml
```diff
@@ -1,6 +1,8 @@
 <?xml version="1.0"?>
 <component name="org.nuxeo.web.ui.properties.contrib">
   <extension target="org.nuxeo.runtime.ConfigurationService" point="configuration">
+    <property name="org.nuxeo.web.ui.sessionTimeoutMinutes">${session.timeout:=30}</property>
```
nelsonsilva: nit: this isn't really a Web UI configuration as we don't have a session timeout on our end. Perhaps we should make it explicit that this is about configuring the keep alive poll interval during S3 upload (ir `org.nuxeo.web.ui.s3.keepAlive`)

guirenard: I has such `org.nuxeo.web.ui.s3.keepAlive` at the beginning and then I thought that since it is actually the tomcat's HTTP session timeout, there's no reason the store it under the s3 namespace in config properties. As a matter of fact, we may realize it could be needed/used for something else than s3 direct upload...

addons/amazon-s3-online-storage/index.js
```diff
@@ -40,19 +40,35 @@ class S3Provider {
...
         Key: this.extraInfo.baseKey.replace(/^\/+/g, '').concat(uuid()),
         ContentType: file.type,
         Body: file,
       });
+      if (sessionTimeout > 0) {
+        this._keepAlive = setInterval(() => {
+          const resource = this._resource();
+          if (!resource.loading) {
+            resource.path = ['upload', this.batchId].join('/');
+            resource.get();
```
Gabez0r: I'd avoid using the same resource, since this `get` could be canceled or cancel the `post`. If we want to reuse the same resource, we should set `uncancelable` to `true` to prevent these requests from being aborted.

plugin/web-ui/addon/src/main/resources/OSGI-INF/web-ui-properties.xml
Gabez0r: It looks weird to me to add a configuration property to Web UI, when it's just used on an add-on. I'd say that we should only sign this contract when we actually need it, since this is something that we'd have to maintain.

PR #789, opened 2019-12-05
NXP-28375: expose iron behaviors for compat

nelsonsilva:


index.js
```diff
@@ -1,11 +1,50 @@
...
+import { IronValidatorBehavior } from '@polymer/iron-validator-behavior/iron-validator-behavior.js';
+
+// expose behaviors for compat
+import { AggregationBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-aggregation/nuxeo-aggregation-behavior.js';
+import { FiltersBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-filters-behavior.js';
+import { FormatBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-format-behavior.js';
+import { I18nBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-i18n-behavior.js';
+import { LayoutBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-layout-behavior.js';
+import { PageProviderDisplayBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-page-provider-display-behavior.js';
+import { RoutingBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-routing-behavior.js';
```
Gabez0r: Do we need to have imports for i18n and Routing behaviors on both the index.js and nuxeo-app.js?

nelsonsilva: `nuxeo-app` uses `RoutingBehavior`, `I18nBehavior` and  `FiltersBehavior` and it does not expose these. Idea is for this compat "hack" to be centralized and outside `nuxeo-app` so we need to explicitly import everything we want to export and not rely on these exported "things" in our own Polymer 3 code.

PR #804, opened 2019-12-19
NXP-28117: trigger refresh on spreadsheet close

nelsonsilva:


elements/search/nuxeo-search-form.js
```diff
@@ -601,14 +601,12 @@ Polymer({
   },
 
   refresh() {
-    if (this.visible) {
```
nmpcunha: I was trying to understand this `visible` check, but I'm not sure what was the purpose (probably not really in use by WebUI).
Anyway, since this is a method from the public API so we may need to aware for a possible behavioural change.

Gabez0r: Do we need to get rid of the `this.visible` check?

nelsonsilva: We do because if you look at https://github.com/nuxeo/nuxeo-web-ui/blob/master/elements/nuxeo-app.js#L850 you'll see that calls to `refreshSearch` actually don't do anything when the drawer is closed...

nelsonsilva:
🔵 https://qa.nuxeo.org/jenkins/job/master/job/web-ui-pipeline/310

PR #806, opened 2019-12-23
NXP-26890: make the selectivity setter helper to clear value in single mode

FabioFSantos:
Make the selectivity setter helper to clear value before setting a new one on single value mode

### DIFF
--- a/packages/nuxeo-web-ui-ftest/features/step_definitions/support/fixtures/layouts.js
+++ b/packages/nuxeo-web-ui-ftest/features/step_definitions/support/fixtures/layouts.js
@@ -24,8 +24,13 @@ const suggestionSet = (element, value) => {
         element.waitForVisible('.selectivity-multiple-input');
         element.element('.selectivity-multiple-input').setValue(values[i]);
       } else {
+        const hasSelectedValue = element.element('.selectivity-single-selected-item').isExisting();
         dropdown.waitForVisible('.selectivity-search-input');
         dropdown.element('.selectivity-search-input').setValue(values[i]);
+        if (hasSelectedValue) {
+          dropdown.element('.selectivity-result-item').waitForVisible();
+          driver.keys('Down arrow');
+        }
       }
       driver.waitUntil(() => {
         if (dropdown.isVisible('.selectivity-result-item.highlight')) {


FabioFSantos:
https://qa.nuxeo.org/jenkins/view/TestAndPush/job/web-ui-pipeline-fdsantos/71/

ftest/features/metadata.feature
```diff
@@ -37,24 +37,28 @@ Feature: Edit metadata
...
+  Scenario: Clear selectivity widget & Edit single item
     Given I have a File document
     When I browse to the document
     Then I can edit the following properties in the File metadata:
       | name         | value                |
       | nature       | Internship report    |
       | subjects     | Medicine,Video games |
-    Then I see the File page
+      | coverage     | Canada               |
+  Then I see the File page
```
nmpcunha: Can you please double check the formatting on this specific line?

ftest/features/metadata.feature
```diff
@@ -37,24 +37,28 @@ Feature: Edit metadata
     | Folder   |
     | Workspace|
 
-  Scenario: Clear selectivity widget
+  Scenario: Clear selectivity widget & Edit single item
```
nmpcunha: I understand the bug and that's the reason why we mix two different things in one scenario. Not sure if I would break it up or, maybe not even having a test for this since the bug was on our testing framework that, by itself, it is not tested. 🤷‍♂

packages/nuxeo-web-ui-ftest/features/step_definitions/support/fixtures/layouts.js
```diff
@@ -18,6 +18,11 @@ const suggestionSet = (element, value) => {
     element.scrollIntoView('#input');
     for (let i = 0; i < values.length; i++) {
       element.waitForVisible(isMulti ? 'input' : '#input');
+      const item = element.element('.selectivity-single-selected-item');
```
nmpcunha: Any specific reason for this to be done inside the for loop?
If I understood it correctly, we generally want to clear that value.

nmpcunha:
nit: we usually start our commit messages with lower case
i.e. `NXP-26890: fix ...`

ftest/features/metadata.feature
FabioFSantos: fixed 👍

packages/nuxeo-web-ui-ftest/features/step_definitions/support/fixtures/layouts.js
FabioFSantos: fixed 👍

ftest/features/metadata.feature
FabioFSantos: I needed this for the studio itests -> [buildind_type_document.feature](https://github.com/nuxeo/nos/blob/aa9381a2c5880b5db7662ec700018971ec37e972/itests/10.10/web-ui/webdriver/test/features/building_type_document.feature#L42) - and I wasn't going to change here but I realized that we don't have a validation for editing a single mode filled field here. Didn't want to add a new feature but I would like to know what's the opinion from you and the rest of the team and what option should we take:
 1 - remove the step and just keep the helper so that I can use it on itests
 2 - add a new scenario just to test this
 3 - keep it like this

nmpcunha: I would suggest the first option because we don't want to test our framework (for now...)
Anyway, I understand your point about not having any test where we edit an entry in single mode... I would suggest to keep this step with the original aim "clearing selectivity", but also with coverage entry, checking that it was also cleared. But just that.

nmpcunha: Regarding my last comment, since single value mode is already being tested by "nature" field, I would suggest to keep the test as is. But let's wait for more feedback 🙂

FabioFSantos:
https://qa.nuxeo.org/jenkins/view/TestAndPush/job/web-ui-pipeline-fdsantos/74/

PR #808, opened 2019-12-26
fix-NXP-27163-HandlingMyFavoritesCollection

nmpcunha:
Pipeline: https://qa.nuxeo.org/jenkins/job/web-ui-pipeline-ncunha/258/

### DIFF
--- a/elements/nuxeo-collections/nuxeo-collections.js
+++ b/elements/nuxeo-collections/nuxeo-collections.js
@@ -475,6 +475,10 @@ Polymer({
     window.addEventListener('added-to-collection', () => {
       if (this.visible) {
         this._refreshCollections();
+        if (this._isDisplayMembers) {
+          this.$.membersList.reset();
+          this.$.membersList.fetch();
+        }
       }
     });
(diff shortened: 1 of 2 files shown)


elements/nuxeo-collections/nuxeo-favorites.js
```diff
@@ -184,11 +184,18 @@ Polymer({
...
+    window.addEventListener('removed-from-collection', this._refresh.bind(this));
   },
 
-  _refresh() {
+  _refresh(e) {
     this._fetchFavorite().then((favorite) => {
-      if (!favorite) {
+      if (
+        !favorite ||
+        (e &&
```
nelsonsilva: nit: would rather compute a boolean flag and keep the oneliner if as it would make it easier to understand at a glance

nelsonsilva:
I'd really rather exclude the Favorites from the collections' list when adding a document to a collecting than doing this as imho the fact the Favorites is a collection should be an implementation detail

nmpcunha:
Backend part: https://github.com/nuxeo/nuxeo/pull/3665

PR #816, opened 2020-01-24
NXP-28495: lock import controls while document is being initialized

Gabez0r:


Gabez0r:
Web UI pipeline: https://qa.nuxeo.org/jenkins/view/All/job/web-ui-pipeline-gbarata/291/ 🔵

elements/nuxeo-document-creation/nuxeo-document-creation-behavior.js
```diff
@@ -19,6 +19,20 @@ import { FormatBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-format-behavior.j
 import { RoutingBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-routing-behavior.js';
 
 let schemaFetcher = null;
+const typeCache = {};
+
+function initializeDoc(doc, properties) {
+  if (doc == null) {
+    return doc;
+  }
+  const newDoc = JSON.parse(JSON.stringify(doc));
```
nmpcunha: Could we use `Object.create`?

elements/document/nuxeo-document-import.js
```diff
@@ -972,24 +982,24 @@ Polymer({
...
-      this._selectDoc(e.model.index);
-    }
+    this._selectDoc(e.model.index);
   },
 
   _tapRemoteDoc(e) {
-    if (this.canCreate) {
-      this._selectDoc(e.model.index + this.localFiles.length);
-    }
+    this._selectDoc(e.model.index + this.localFiles.length);
```
nmpcunha: Not sure if worths to have only one method since what changes is basically the index.
Same for both methods below.

elements/nuxeo-document-creation/nuxeo-document-creation-behavior.js
Gabez0r: Don't think object assign would help, since it does shallow copies.

elements/document/nuxeo-document-import.js
Gabez0r: If we were to use a single method, how would we parameterize it to do local or remote tap? Please keep in mid that these are listeners and not computed bindings.

PR #821, opened 2020-01-27
NXP-27569: add more functional tests for user creating

PGoncalvesNX:


packages/nuxeo-web-ui-ftest/features/step_definitions/user.js
```diff
@@ -24,6 +24,10 @@ Then(/^I can search for the user "([^"]*)"$/, function(username) {
   user.waitForVisible().should.be.true;
 });
 
+Then(/^I can see the user has the email "([^"]*)"$/, function(email) {
+  this.ui.user.getField(email).should.not.be.empty;
```
nmpcunha: Semantically it may not make much sense.
We are checking that the e-mail is not empty when our definition says `has the email X`, a concrete one. I would suggest to:
- Reword the step;
- Check, in fact, that we are comparing the concrete e-mail address;

FabioFSantos: I agree, we should be comparing the concrete email and not if the field is not empty

packages/nuxeo-web-ui-ftest/features/step_definitions/user.js
```diff
@@ -24,6 +24,13 @@ Then(/^I can search for the user "([^"]*)"$/, function(username) {
   user.waitForVisible().should.be.true;
 });
 
+Then(/^I can see the user has the email "([^"]*)"$/, function(email) {
+  this.ui.user
+    .getField(email)
```
guirenard: don't you mean `'email'`? here if the email is grenard@nuxeo.com, you'll be trying to retrieve something like `<input name="grenard@nuxeo.com">`  which I doubt it exists

packages/nuxeo-web-ui-ftest/features/step_definitions/user.js
```diff
@@ -24,6 +24,12 @@ Then(/^I can search for the user "([^"]*)"$/, function(username) {
   user.waitForVisible().should.be.true;
 });
 
+Then(/^I can see the user has the email "([^"]*)"$/, function(useremail) {
```
semisse: nit: shall we use camel case here ie `userEmail`?

nmpcunha: +1 on this.

PGoncalvesNX:
The pipeline has several failed runs but it is due to other tests' flakiness. The test affected by these changes is passing.

packages/nuxeo-web-ui-ftest/features/step_definitions/user.js
```diff
@@ -24,6 +24,14 @@ Then(/^I can search for the user "([^"]*)"$/, function(username) {
   user.waitForVisible().should.be.true;
 });
 
+Then(/^I can see the user has the email "([^"]*)"$/, function(userEmail) {
+  const user = this.ui.user.el.element('nuxeo-view-user [name="email"]');
+  user
```
nmpcunha: Damn formatter 😅 
It is not that bad to have it in one line...

PGoncalvesNX:
https://qa.nuxeo.org/jenkins/view/TestAndPush/job/web-ui-pipeline-pgoncalves/6/

PR #824, opened 2020-02-05
NXP-28610: ensure tasks drawer is updated when task is processed

nmpcunha:


elements/nuxeo-app.js
```diff
@@ -912,6 +912,8 @@ Polymer({
     // let's refresh the current document since it might have been changed (ex: state and version)
     this.fire('document-updated');
     this._fetchTaskCount();
+    this.currentTask = null;
+    this.currentTaskId = null;
```
Gabez0r: note: task selection will be lost here, which I think is ok, since it is in fact a different task

Gabez0r: I don't think resetting the current task is the key here. The main problem is that the drawer is not updated when open, not only when you process a task (addressed by these lines), but also when you start or end a workflow. I'd expect us to actually refetch the tasks, in order to receive any changes that might have occurred; something along these lines:
```
    const tasksDrawer = this.$$('nuxeo-tasks-drawer');
    if (tasksDrawer.visible) {
      tasksDrawer.$.tasks.fetch();
    }
```

nmpcunha: Yes, that was the aim. 👍

nmpcunha: Yes, probably it is safer to have that approach for all tasks events that follow this flow. 
Thanks!

elements/nuxeo-app.js
```diff
@@ -615,6 +615,11 @@ Polymer({
     Performance.mark('nuxeo-app.ready');
   },
 
+  _resetSelections() {
```
Gabez0r: nit: I'd name this `_resetSelection` (singular) or even `_resetTaskSelection`, since we're only changing task selection.

nmpcunha: Fair! 👍

PR #832, opened 2020-02-14
NXP-28666: check enforceCommentable property as a String

guirenard:


guirenard:
https://qa2.nuxeo.org/jenkins/job/WebUI/job/web-ui-pipeline-10.10/377/ OK

PR #842, opened 2020-02-28
NXP-24967: allow saved searches to be loaded via uid url

Gabez0r:


elements/nuxeo-app.js
```diff
@@ -666,6 +690,19 @@ Polymer({
     this.$.doc
       .get()
       .then((doc) => {
+        if (this.docId && doc.facets.includes('SavedSearch')) {
+          this._routedSearch = doc;
+          const savedSearchForm = this._getSavedSearchForm();
+          if (savedSearchForm) {
+            const name = savedSearchForm.getAttribute('search-name');
+            this.navigateTo('search', name);
+            if (!this.__searchOnLoad) {
```
nmpcunha: Should this be `!this._searchOnLoad`?

Gabez0r: Yes, because if `this._searchOnLoad` is true, than `_loadSavedSearch` is expected to run on `_updateSearch` (where we set `_searchOnLoad` to `false`).

Gabez0r: ah, sorry, it's the double `_`. Will fix it, thanks!

RSalem07:
I now this is not related to your modification :). Should we define the `searches` property on the `nuxeo-search-form` element ->  
[here](https://github.com/nuxeo/nuxeo-web-ui/blob/b6de55a7dbbc06a207cf87acbcda317aae53b33d/elements/search/nuxeo-search-form.js#L253)

Gabez0r:
@RSalem07 Yes, I agree. For clarity and consistency it should be declared, although we would have to rename it since it would most likely become a private property. Feel free to create a cleanup ticket. 👍

Gabez0r:
As for the second comment about the `this`, I'd say that it's a left over from a previous version of the element (from 10.10). We should double check and remove it. We can re-use the same clean up ticket as described above.

RSalem07:
thanks for taking into account the add of the `search` field and the remove of the bind call :)

elements/search/nuxeo-search-form.js
```diff
@@ -905,4 +911,26 @@ Polymer({
   _displayQuickFilters() {
     return this._quickFilters && this._quickFilters.length > 0;
   },
+
+  _loadSavedSearch(id) {
+    if (id) {
```
andreacornaglia: can we flatten out this function? It's a little bit hard to follow the logic of ifs/elses.  My suggestions:
if(!id) return;
const load = () => { if(this.searches){ return} ...}
if (this.results) return load();
this._loadSavedSearchListener

Gabez0r: Thank you for the suggestion. I'll make an early return to reduce indentation:
```
    if (!id) {
      return;
    }
```
But we usually do not use single-line if blocks because they can lead to errors when someone whats to add a a second statement, and they force people to add the curly-braces afterwards, making diffs more messy.

Gabez0r:
Pipeline check: https://qa.nuxeo.org/jenkins/view/All/job/web-ui-pipeline-gbarata/327/ 🔵

elements/search/nuxeo-search-form.js
andreacornaglia: yes! single line was just to save space in the comment :)

PR #850, opened 2020-03-12
NXP-28662: Trashed folder allows document creation

RSalem07:
T&P In Progress: https://qa.nuxeo.org/jenkins/job/master/job/web-ui-saouana/1/
T&P will be relaunched once the functional tests issue fixed (related on stuff to fix `pom.xml`, `assembly.xml`...

new T&P after rebasing the master [PR](https://github.com/nuxeo/nuxeo-web-ui/pull/853)  is now merged into master 
https://qa.nuxeo.org/jenkins/job/master/job/web-ui-saouana/2/ -> OK

### DIFF
--- a/elements/document/nuxeo-document-create.js
+++ b/elements/document/nuxeo-document-create.js
@@ -200,6 +200,7 @@ Polymer({
                   class="docTypeButton"
                   on-tap="_selectType"
                   data-args$="[[type]]"
+                  disabled$="[[!_canCreate(canCreate, creating)]]"
                 >
                   <iron-icon class="typeIcon" src="[[_getTypeIcon(type)]]"></iron-icon>
                   <div class="typeLabel">[[_getTypeLabel(type)]]</div>
--- a/i18n/messages.json
+++ b/i18n/messages.json
@@ -404,6 +404,7 @@
   "documentCreationBehavior.error.invalidSubtype": "You cannot create this type of document in this location.",
   "documentCreationBehavior.error.noPermission": "You do not have permissions to create under this location.",
   "documentCreationBehavior.error.noSubtypes": "Select a different container to create your content.",
+  "documentCreationBehavior.error.trashedParent": "You cannot create content under a trashed document.",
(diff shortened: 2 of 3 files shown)


elements/document/nuxeo-document-create.js
```diff
@@ -200,6 +200,7 @@ Polymer({
                   class="docTypeButton"
                   on-tap="_selectType"
                   data-args$="[[type]]"
+                  disabled$="[[!_canCreate()]]"
```
Gabez0r: _canCreate has two dependencies, `canCreate` and `creating`:
```
  _canCreate() {
    return this.canCreate && !this.creating;
  },
```
This means that we'd like it to be recomputed when they changed:
```suggestion
                  disabled$="[[!_canCreate(canCreate, creating)]]"
```

i18n/messages.json
```diff
@@ -399,6 +399,7 @@
   "documentCreationBehavior.error.invalidSubtype": "You cannot create this type of document in this location.",
   "documentCreationBehavior.error.noPermission": "You do not have permissions to create under this location.",
   "documentCreationBehavior.error.noSubtypes": "Select a different container to create your content.",
+  "documentCreationBehavior.error.trashedParent": "You cannot create this type of document under a trashed document.",
```
Gabez0r: We cannot create any type of document on a trashed folder, right? Perhaps something like this would me shorter and more appropriate?:
```suggestion
  "documentCreationBehavior.error.trashedParent": "You cannot create content under a trashed document.",
```

elements/document/nuxeo-document-create.js
RSalem07: Yes I agree, ;) I did it on my local, `disabled$="[[!_canCreate(canCreate)]]"` but I loosed it before I push, I think when I did the screenshot. Thanks

RSalem07:
Thanks for your time, I think you agree :) with the solution, what I forgot to say is with the modification of the document behaviour, we will disable the import on trashed docs too

PR #853, opened 2020-03-13
NXP-28000: add missing dependency and correct package name

nmpcunha:
Platform PR: https://github.com/nuxeo/nuxeo/pull/3832

### DIFF
--- a/plugin/itests/addon/src/main/resources/OSGI-INF/extensions.xml
+++ b/plugin/itests/addon/src/main/resources/OSGI-INF/extensions.xml
@@ -6,7 +6,7 @@
   <require>org.nuxeo.ecm.core.schema.TypeService</require>
 
   <extension target="org.nuxeo.ecm.core.schema.TypeService" point="schema">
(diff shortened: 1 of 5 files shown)


plugin/itests/addon/src/main/resources/OSGI-INF/extensions.xml
```diff
@@ -6,7 +6,7 @@
   <require>org.nuxeo.ecm.core.schema.TypeService</require>
 
   <extension target="org.nuxeo.ecm.core.schema.TypeService" point="schema">
-    <schema name="files" src="schema/files.xsd"/>
+    <schema name="files" src="data/schemas/files.xsd"/>
```
Gabez0r: Do we really need to deploy a local schema? I'm assuming `<schema name="files" enabled="true" />` won't work?

nmpcunha: Indeed. I'll create a ticket to evaluate the possibility of implementing that behaviour. I'll double check with Architecture Team.

PR #856, opened 2020-03-19
NXP-28601: Transfer document in S3 Glacier - Manual action in WebUI a…

RSalem07:
…t document level

This PR needs the https://github.com/nuxeo/nuxeo-elements/pull/217


elements/nuxeo-document-info-bar/nuxeo-document-info-bar.js
```diff
@@ -251,4 +268,8 @@ Polymer({
   _toggleGraphDialog(e) {
     this.$$(`#graph-${e.model.workflow.id}`).show();
   },
+
+  _contentStoredInColdStorage(doc) {
+    return this.hasFacet(doc, 'ColdStorage') && doc.properties['coldstorage:coldContent'];
```
nmpcunha: I would suggest something like:`return this.hasFacet(doc, 'ColdStorage') && doc.properties && doc.properties['coldstorage:coldContent']`

elements/nuxeo-web-ui-bundle.html
```diff
@@ -327,6 +327,15 @@
   </template>
 </nuxeo-slot-content>
 
+<nuxeo-slot-content name="moveToColdStorageDocumentAction" slot="DOCUMENT_ACTIONS" order="120">
```
nmpcunha: Not sure about the ordering, we just need to double check where we want to show this button by default 😉

elements/nuxeo-document-info-bar/nuxeo-document-info-bar.js
```diff
@@ -60,6 +60,11 @@ Polymer({
         color: white;
       }
 
+      .bar.coldstorage {
+        background: black;
```
nmpcunha: I didn't had a chance to follow the conversation, but did we ended up with black background instead of blue?

elements/nuxeo-document-info-bar/nuxeo-document-info-bar.js
```diff
@@ -177,6 +182,18 @@ Polymer({
     <template is="dom-if" if="[[isPublication(document)]]">
       <nuxeo-publication-info-bar document="[[document]]"></nuxeo-publication-info-bar>
     </template>
+
+    <!-- cold storage-->
+    <template is="dom-if" if="[[_contentStoredInColdStorage(document)]]">
+      <div id="coldStorageInfoBar" class="bar coldstorage">
+        <div class="layout horizontal center flex">
+          <span><iron-icon noink icon="nuxeo:coldstorage"></iron-icon></span>
```
nmpcunha: We probably don't need to wrap the icon inside a span, right? 🤔

elements/nuxeo-document-info-bar/nuxeo-document-info-bar.js
```diff
@@ -177,6 +182,18 @@ Polymer({
...
+          <span>[[i18n('documentPage.content.storedInColdStorage')]]</span>
+          <span>&nbsp;</span>
+          <span>[[i18n('documentPage.content.lowResolutionPreview')]]</span>
```
nmpcunha: Since we need the first sentence to be styled in a different way than the second one, we may want to create a class for it and it may also save us the `<span>$nbsp;</span>`

elements/nuxeo-web-ui-bundle.html
RSalem07: yeah, i just put it at the last position ;)

elements/nuxeo-document-info-bar/nuxeo-document-info-bar.js
RSalem07: really i should ask Carolina, because i think she thinks that the retention in a blue once and  i check the css of retention and it the dark color. I will ping her to confirmation

elements/nuxeo-document-info-bar/nuxeo-document-info-bar.js
RSalem07: I agree this was the copy/past of the original once where i need the fact of the wrapping it for more css control ;)

elements/nuxeo-document-info-bar/nuxeo-document-info-bar.js
carolinagouveia: I did believe that the background of retention was light blue. For consistency, let's leave it in black then.

RSalem07: yeah, [here](https://github.com/nuxeo/nuxeo-web-ui/blob/master/elements/nuxeo-document-info-bar/nuxeo-document-info-bar.js#L58) and [here](https://github.com/nuxeo/nuxeo-web-ui/blob/master/elements/nuxeo-document-info-bar/nuxeo-document-info-bar.js#L122)


thanks Carolina

addons/nuxeo-coldstorage/elements/nuxeo-move-content-to-coldstorage-button.js
```diff
@@ -0,0 +1,145 @@
...
+            <paper-icon-button icon="nuxeo:coldstorage" noink></paper-icon-button>
+            <span class="label">[[i18n('documentContent.moveToColdStorage.label')]]</span>
+            <nuxeo-tooltip>[[i18n('documentContent.moveToColdStorage.tooltip')]]</nuxeo-tooltip>
+          </div>
+        </template>
+      </dom-if>
+
+      <nuxeo-dialog id="contentToMoveDialog" with-backdrop>
+        <h2>[[i18n('documentContent.moveToColdStorage.ask.confirmation')]]</h2>
+        <p class="horizontal">
```
nmpcunha: Where is this `horizontal` class defined?

addons/nuxeo-coldstorage/elements/nuxeo-move-content-to-coldstorage-button.js
```diff
@@ -0,0 +1,145 @@
...
+  _isAvailable(document) {
+    return (
+      !this.hasFacet(document, 'ColdStorage') &&
+      (this.hasAdministrationPermissions(this.user) || this.hasPermission(document, 'WriteColdStorage')) &&
+      this.hasContent(document)
+    );
+  }
+
+  _toggle() {
+    if (this._isAvailable(this.document)) {
```
nmpcunha: Since this method is called when the content is rendered (i.e. when `_isAvailable` already ran), I'm not sure if we need to double check the permissions here 🙂

addons/nuxeo-coldstorage/elements/nuxeo-move-content-to-coldstorage-button.js
```diff
@@ -0,0 +1,145 @@
...
+
+          this.dispatchEvent(
+            new CustomEvent('notify', {
+              composed: true,
+              bubbles: true,
+              detail: { message: [[this.i18n('documentContent.moveToColdStorage.success')]] },
+            }),
+          );
+        },
+        () => {
```
nmpcunha: We usually prefer `catch` syntax, since it is more explicit (still being a little more verbose). Anyway, no problem with this, I suppose 🙂

addons/nuxeo-coldstorage/elements/nuxeo-move-content-to-coldstorage-button.js
```diff
@@ -0,0 +1,145 @@
...
+          this.dispatchEvent(
+            new CustomEvent('notify', {
+              composed: true,
+              bubbles: true,
+              detail: { message: [[this.i18n('documentContent.moveToColdStorage.error')]] },
+            }),
+          );
+        },
+      );
+    } else {
```
nmpcunha: Without the `if` we can cleanup this code branch and also remove one more message, right?

addons/nuxeo-coldstorage/elements/nuxeo-move-content-to-coldstorage-button.js
nmpcunha: We can have different error codes here: at least 401, 403 and 500, right? Should we think in different messages? 🤔

addons/nuxeo-coldstorage/nuxeo-coldstorage.html
```diff
@@ -0,0 +1,24 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<nuxeo-slot-content name="moveToColdStorageDocumentAction" slot="DOCUMENT_ACTIONS" order="120">
+  <template>
+    <nuxeo-move-content-to-coldstorage-button
+      document="[[document]]"
+      user="[[user]]"
```
nmpcunha: Not sure if we want to pass the user as argument. 🤔 
We can have user's information by using `nuxeo-connection`. WDTY @Gabez0r ?

elements/nuxeo-document-info-bar/nuxeo-document-info-bar.js
```diff
@@ -60,6 +60,15 @@ Polymer({
         color: white;
       }
 
+      .bar.coldStorage {
+        background: black;
+        color: white;
+      }
+
+      .coldStorage .storedInColdStorage {
```
nmpcunha: It is correct, but we don't need have a parent class, since `.storedInColdStorage ` is not ambiguous and it not being used anywhere else (so no risk of "clashes"). It would only be for readable purposes, right? 🙂

addons/nuxeo-coldstorage/elements/nuxeo-move-content-to-coldstorage-button.js
RSalem07: My idea here is to hide the message if any. but if the in others places is what you do, i will add the server message here to be consistent

addons/nuxeo-coldstorage/nuxeo-coldstorage.html
RSalem07: `nuxeo-connection` will send an http request to the server backend or not ?

addons/nuxeo-coldstorage/elements/nuxeo-move-content-to-coldstorage-button.js
```diff
@@ -0,0 +1,136 @@
...
+@element nuxeo-move-content-to-coldstorage-button
+*/
+class MoveToColdStorage extends mixinBehaviors([FiltersBehavior, FormatBehavior], Nuxeo.Element) {
+  static get template() {
+    return html`
+      <style include="nuxeo-action-button-styles nuxeo-styles"></style>
+      <nuxeo-operation id="opMove" op="Document.MoveToColdStorage" input="[[document.uid]]"></nuxeo-operation>
+      <nuxeo-connection id="nxcon" user="{{currentUser}}"></nuxeo-connection>
+
+      <dom-if if="[[_isAvailable(document)]]">
```
Gabez0r: _isAvailable has two dependencies: document anda currentUser, so they should be reflected here as well:
```suggestion
      <dom-if if="[[_isAvailable(document, currentUser)]]">
```

addons/nuxeo-coldstorage/elements/nuxeo-move-content-to-coldstorage-button.js
```diff
@@ -0,0 +1,136 @@
...
+       */
+      currentUser: Object,
+    };
+  }
+
+  _toggleDialog() {
+    this.$.contentToMoveDialog.toggle();
+  }
+
+  _isAvailable(document) {
```
Gabez0r: Following the comment above, for consistency, we should:
```suggestion
  _isAvailable(document, currentUser) {
```

addons/nuxeo-coldstorage/elements/nuxeo-move-content-to-coldstorage-button.js
```diff
@@ -0,0 +1,136 @@
...
+    return (
+      !this.hasFacet(document, 'ColdStorage') &&
+      (this.hasAdministrationPermissions(this.currentUser) || this.hasPermission(document, 'WriteColdStorage')) &&
```
Gabez0r:
```suggestion
      (this.hasAdministrationPermissions(currentUser) || this.hasPermission(document, 'WriteColdStorage')) &&
```

addons/nuxeo-coldstorage/elements/nuxeo-move-content-to-coldstorage-button.js
```diff
@@ -0,0 +1,136 @@
...
+`nuxeo-move-content-to-coldstorage-button`
+@group Nuxeo UI
+@element nuxeo-move-content-to-coldstorage-button
+*/
+class MoveToColdStorage extends mixinBehaviors([FiltersBehavior, FormatBehavior], Nuxeo.Element) {
+  static get template() {
+    return html`
+      <style include="nuxeo-action-button-styles nuxeo-styles"></style>
+      <nuxeo-operation id="opMove" op="Document.MoveToColdStorage" input="[[document.uid]]"></nuxeo-operation>
+      <nuxeo-connection id="nxcon" user="{{currentUser}}"></nuxeo-connection>
```
Gabez0r: Missing import.

addons/nuxeo-coldstorage/elements/nuxeo-move-content-to-coldstorage-button.js
```diff
@@ -0,0 +1,136 @@
...
+
+        this.dispatchEvent(
+          new CustomEvent('notify', {
+            composed: true,
+            bubbles: true,
+            detail: { message: [[this.i18n('documentContent.moveToColdStorage.success')]] },
+          }),
+        );
+      })
+      .catch(() => {
```
Gabez0r: We don't care about what the error is here, right? If so, we should at least re-throw it, so it doesn't get swallowed.

i18n/messages.json
```diff
@@ -392,6 +392,15 @@
   "documentContentView.datatable.header.type": "Type",
   "documentContentView.datatable.header.version": "Version",
   "documentContentView.drag.import": "Drop to import",
+  "documentContentView.moveToColdStorage.ask.confirmation": "Do you want to send the document content to cold storage?",
```
Gabez0r: Please confirm with Carolina (perhaps with a screenshot) that this change looks alright. 🙏

carolinagouveia: For consistency with the other confirmation dialog, we just need to change the message to “Send the document content to cold storage?”. Everything else looks good.

i18n/messages.json
```diff
@@ -461,6 +470,8 @@
   "documentPage.activity": "Activity",
   "documentPage.collections": "Collections",
   "documentPage.comments": "Comments",
+  "documentPage.content.storedInColdStorage": "The master file is stored in cold storage.",
+  "documentPage.content.lowResolutionPreview": "This a low resolution preview.",
```
nmpcunha: These two should be part of the second commit, but no big deal 😉

RSalem07: thanks ;), I did the modification to keep things consistent.

PR #858, opened 2020-03-24
NXP-28785: compare audit comment using regex

nmpcunha:
The second commit is, in fact, a workaround for some limitations that we have on our event system.
Ideally, we would have information provided by the server to know which is the type of event's comment. Then we could have a formatter approach as we have for Diff.

### DIFF
--- a/elements/nuxeo-document-history/nuxeo-document-history.js
+++ b/elements/nuxeo-document-history/nuxeo-document-history.js
@@ -81,7 +81,7 @@ Polymer({
         </nuxeo-data-table-column>
         <nuxeo-data-table-column name="[[i18n('documentHistory.comment')]]">
           <template>
-            <a href$="[[_parseComment(item.comment)]]">[[item.comment]]</a>
+            <a href$="[[_parseComment(item.comment)]]">[[_formatComment(item.comment)]]</a>
           </template>
         </nuxeo-data-table-column>
         <nuxeo-data-table-column name="[[i18n('documentHistory.state')]]">
@@ -163,11 +163,19 @@ Polymer({
     return this.i18n(`activity.${key}`);
   },
 
+  // XXX: Both parse and format methods shouldn't be needed after NXP-28820
   _parseComment(comment) {
-    if (comment && /^(\w+):(\S+)$/.test(comment)) {
+    if (comment && /^\w+:(?:\w+-){2,}(?:\w+)$/.test(comment)) {
       // repoName:docId
       return this.urlFor('document', comment.split(':')[1]);
     }
     return null;
+  },
+
+  _formatComment(comment) {
+    if (moment(comment, moment.ISO_8601).isValid()) {
+      return this.formatDateTime(comment);
+    }
+    return comment;
   },
 });
--- a/elements/nuxeo-document-history/nuxeo-document-history.js
+++ b/elements/nuxeo-document-history/nuxeo-document-history.js
@@ -164,13 +164,9 @@ Polymer({
   },
 
   _parseComment(comment) {
-    if (comment) {
-      const split = comment.split(':');
-      if (split.length >= 2) {
-        // split[0] is repo name, split[1] is doc id
-        return this.urlFor('document', split[1]);
-      }
-      return null;
+    if (comment && /^(\w+):(\S+)$/.test(comment)) {
+      // repoName:docId
+      return this.urlFor('document', comment.split(':')[1]);
     }
     return null;
   },


elements/nuxeo-document-history/nuxeo-document-history.js
```diff
@@ -170,4 +170,11 @@ Polymer({
     }
     return null;
   },
+
+  _formatComment(comment) {
+    if (comment && !Number.isNaN(Date.parse(comment))) {
+      return this.formatDateTime(comment);
+    }
+    return comment;
+  },
```
RSalem07: can be replaced by:

```
  if (!Number.isNaN(Date.parse(comment))) {
      return this.formatDateTime(comment);
    }
    return comment ? comment : '';
```
the check on `Number.isNaN(Date.parse(comment)`will do the job as 

`Number.isNaN(Date.parse(undefined)` -> `true`
`Number.isNaN(Date.parse(null)` -> `true`
`Number.isNaN(Date.parse('null')` -> `true`

nmpcunha: Thanks! It is a fair point!
I just did not take the `return comment ? comment : '';` because we don't need that in this context. If it returns null or undefined, it will be rendered as empty. 🙂

elements/nuxeo-document-history/nuxeo-document-history.js
```diff
@@ -164,13 +164,9 @@ Polymer({
...
-      if (split.length >= 2) {
-        // split[0] is repo name, split[1] is doc id
-        return this.urlFor('document', split[1]);
-      }
-      return null;
+    if (comment && /^(\w+):(\S+)/.test(comment)) {
+      // repoName:docId
+      return this.urlFor('document', comment.split(':')[1]);
     }
     return null;
```
RSalem07: I am wondering if in the comment we have cases where the  text is like `anyWord:ddd 555` the regex will handle it as an url doc but it isn't 

```
let str = 'myRepo:dfechfjfhf150fgj k';
console.log(/^(\w+):(\S+)/.test(str)); --> true
console.log( /^(\w+):(\w+)$/.test(str)); --> false
```

I don't wdyt if we replace it with `^(\w+):(\w+)$`, even with this one if an user add a comment like `first:thanks` it will be take as an url but I test the 2nd regex and it avoids `first:thanks any thing` to be evaluated as an repo url

nmpcunha: Indeed. Thanks for the suggestion 👍 
The be even more robust, just if we could assume that the second part is always a UUID. But I'm not sure if we want to go that way 🤷‍♂ 
WDYT?

RSalem07: you welcome ;) 

for
>To be even more robust, just if we could assume that the second part is always a UUID. But I'm not sure if we want to go that way 🤷‍♂
WDYT?

I asked myself the same question, but I am not sure, I don't know if in some case we can find ourself using a D.B sequence generator like on Oracle even Postgres

nmpcunha:
A bigger discussion happening at backend level may led the second commit to be dropped.
Further details [here](https://github.com/nuxeo/nuxeo/pull/3849)

RSalem07:
@nmpcunha thanks for your feedback, I asked yesterday Julien to answer to our interrogations on the other PR, I saw the answer and the new interrogations.  let wait to see what will be the actions to take.

elements/nuxeo-document-history/nuxeo-document-history.js
Gabez0r: I think we can assume the part after the `:` is supposed to be an id, based on [what JSF does](https://github.com/nuxeo/nuxeo/blob/master/nuxeo-features/nuxeo-platform-audit/nuxeo-platform-audit-api/src/main/java/org/nuxeo/ecm/platform/audit/api/comment/CommentProcessorHelper.java#L60-L88).

Also, if we are trying to get doc ids, then `(\w+):(\w+)` won't pick up the dashes:
```
console.log( /^(\w+):(\w+)$/.test('myRepo:dfechfjfhf150fgj k')); // false
console.log( /^(\w+):(\w+)$/.test('myRepo:dfechfjfhf150fgj')); // true
console.log( /^(\w+):(\w+)$/.test('myRepo:dfec-hfjf-hf150fgj')); // false
```
If we don't want anything too fancy, but sill want to prevent stuff like `first:thanks`, `first:tanks-` or `john:vira-te` to be accepted, I'd suggest something like: `/^\w+:(?:\w+-){2,}(?:\w+)$/`. It's not bullet proof and we can always use a UUID regex. Also, if we decide we want to allow path as well, we could try something like `/^\w+:[\w+\\\-]+$/`.

Gabez0r: See related [NXP-22062](https://jira.nuxeo.com/browse/NXP-22062).

elements/nuxeo-document-history/nuxeo-document-history.js
```diff
@@ -164,14 +164,17 @@ Polymer({
...
-      return null;
+    if (comment && /^(\w+):(\S+)$/.test(comment)) {
+      // repoName:docId
+      return this.urlFor('document', comment.split(':')[1]);
     }
     return null;
   },
+
+  _formatComment(comment) {
+    if (!Number.isNaN(Date.parse(comment))) {
```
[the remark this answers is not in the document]
RSalem07: i think as we use the ISO format date (at least for R.M I am not sure about others uses cases) we can check using a third part library like `moment.js`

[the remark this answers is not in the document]
nmpcunha: Of course! Didn't noticed that we already have moment here 😅 🤦‍♂ 
Let's not reinvent the wheel 😉

elements/nuxeo-document-history/nuxeo-document-history.js
```diff
@@ -163,15 +163,19 @@ Polymer({
...
-      return null;
+    if (comment && /^\w+:(?:\w+-){2,}(?:\w+)$/.test(comment)) {
+      // repoName:docId
+      return this.urlFor('document', comment.split(':')[1]);
     }
     return null;
   },
+
+  _formatComment(comment) {
+    if (moment(comment).isValid()) {
```
Gabez0r: Please keep in mind that `moment('123').isValid()` returns true.

nmpcunha: Check! Failed to push 🤦‍♂

RSalem07: Yes Gabriel, my idea was to use a Date Format and try to check it:

`alert(moment("2011-12-03T10:15:30+01:00[Europe/Paris]", "YYYY-MM-DD", true).isValid())` -->> `true`
`alert(moment("1253", "YYYY-MM-DD", true).isValid())` --> `false`

perhaps just ensure that we have at least: year, month, day

nmpcunha: I've used `moment.ISO_8601` since, in theory, the dates from backend are in ISO format. 🙂

PR #861, opened 2020-03-26
NXP-27969: automate video dam tests

PGoncalvesNX:
https://qa.nuxeo.org/jenkins/job/web-ui-pipeline-pgoncalves/11/


addons/nuxeo-dam/ftest/features/video.feature
```diff
@@ -0,0 +1,47 @@
...
+      | description  | my description    |
+      | content      | Park.mp4        |
+      | nature       | Booklet       |
+      | subjects     | Ecology,Gardening |
+    Then I see the Video page
+    And I can see Video metadata with the following properties:
+      | name         | value                            |
+      | title        | my title                         |
+      | description  | my description                   |
+      | nature       | Booklet                      |
```
nmpcunha: It looks better if we align the pipes `|` in each step.

addons/nuxeo-dam/ftest/features/step_definitions/video.js
```diff
@@ -0,0 +1,13 @@
+import { Then } from 'cucumber';
+
+Then('I can see the video conversions panel', function() {
+    const page = this.ui.browser.documentPage(this.doc.type);
+    page.waitForVisible();
+    page.el.element('div.additional').waitForVisible('nuxeo-video-conversions').should.be.true;
```
nmpcunha: Despite this kind of approach being used in other places, I believe it may turn this test flaky.
Would it be sensible to check if `div.additional` is stamped in order to then wait for `nuxeo-video-conversions`? @Gabez0r is doing some stabilisation work and he may have further input 🙂

Gabez0r: Yes, I agree. We should do a `page.waitForVisible('div.additional')` first.

addons/nuxeo-dam/ftest/features/step_definitions/video.js
```diff
@@ -0,0 +1,13 @@
...
+Then('I can see the video conversions panel', function() {
+  const page = this.ui.browser.documentPage(this.doc.type);
+  page.waitForVisible();
+  page.el.element('nuxeo-video-conversions').waitForVisible().should.be.true;
+});
+
+Then('I can see the video storyboard', function() {
+  const page = this.ui.browser.documentPage(this.doc.type);
+  page.waitForVisible();
+  page.el.element('nuxeo-video-viewer#video .hasStoryboard').waitForVisible().should.be.true;
```
nmpcunha: Since the focus is on storyboard, and despite we are adding the `hasStoryboard` class dynamically, I would suggest a different selector, pointing directly to storyboard's `div` (https://github.com/nuxeo/nuxeo-elements/blob/master/ui/viewers/nuxeo-video-viewer.js#L96)

Do you think something like this would make sense?
```
const videoViewer = page.el.element('nuxeo-video-viewer')
videoViewer.waitForVisible();
videoViewer.el.element('#storyboard').waitForVisible().should.be.true;
```

I believe that this would be more robust (i.e. less prone to flakiness 🤷‍♂ )

PGoncalvesNX:
https://github.com/nuxeo/nuxeo-web-ui/pull/861#pullrequestreview-381912823
By copying the structure of picture.feature, I would have to use `I have a Video document` and that creates a Video document without a proper video-preview

PR #865, opened 2020-03-30
NXP-28827: bind multiple attribute to blob-list in dropzone (proposal)

mnixo:
The problem: 
NXP-28827
All the info on the [ticket](https://jira.nuxeo.com/browse/NXP-28827).


nmpcunha:
It is a backport of a small part of [this commit](https://github.com/nuxeo/nuxeo-web-ui/commit/8a343a83d70e405e7f7f8a16aa7ec9ff51b410c9#diff-bfcef58b64ac39d85e6fc62124c66171R157)
I would say that it makes sense, besides not being reported by any customer.

Thanks @mnixo ! 👍

PR #866, opened 2020-04-01
Feature nxp 28602 transfer document in s3 glacier manual action in web ui for a resultset of documents

RSalem07:


andreacornaglia:
At a first glance looks good, would be nice to have a small demo to see the visual/interaction part :)

nmpcunha:
> At a first glance looks good, would be nice to have a small demo to see the visual/interaction part :)

Someday we'll have the preview environment working. In the meanwhile, we can checkout the branch and take a look locally 😉

i18n/messages.json
```diff
@@ -809,6 +809,13 @@
   "mimetype.text/plain": "Text",
   "mimetype.text/x-web-markdown": "Markdown",
   "mimetype.text/xml": "XML",
+  "moveDocumentsContentsToColdStorage.ask.confirmation": "Send the documents contents to cold storage?",
+  "moveDocumentsContentsToColdStorage.description": "This will result in increased file retrieval times and a low resolution preview.",
+  "moveDocumentsContentsToColdStorage.cancel": "Cancel",
+  "moveDocumentsContentsToColdStorage.confirm": "Confirm",
+  "moveDocumentsContentsToColdStorage.error": "Error moving the documents contents to cold storage.",
+  "moveDocumentsContentsToColdStorage.success": "Documents contents moved to cold storage successfully.",
+  "moveDocumentsContentsToColdStorage.tooltip": "Send to cold storage the selected items",
```
semisse, line 818: nit: 
```suggestion
  "moveDocumentsContentsToColdStorage.tooltip": "Send the selected items to cold storage",
```

semisse:
Looks good! Just left a small suggestion on the labels.

elements/nuxeo-document-attachments/nuxeo-document-attachments.js
```diff
@@ -160,6 +160,11 @@ Polymer({
   },
 
   _isAvailable(document, xpath) {
-    return document && xpath && this.hasSchema(document, xpath.split(':')[0]);
+    return (
+      document &&
+      xpath &&
+      this.hasSchema(document, xpath.split(':')[0]) &&
+      !(this.hasFacet(document, 'ColdStorage') && this.hasContent(document, 'coldstorage:coldContent'))
```
nmpcunha, line 167: If I'm seeing this correctly, I can't find any reason why we should "hide" the attachments when we have a document with its content on Cold Storage 🤔

addons/nuxeo-coldstorage/elements/nuxeo-move-contents-to-coldstorage-button.js
```diff
@@ -0,0 +1,137 @@
...
+
+  _toggleDialog() {
+    this.$.contentsToMoveDialog.toggle();
+  }
+
+  _isAvailable(documents, currentUser) {
+    if (this.hasAdministrationPermissions(currentUser) || this.hasPermission(document, 'WriteColdStorage')) {
+      return documents.every((doc) => !this.hasFacet(doc, 'ColdStorage') && this.hasContent(doc));
+    }
+
```
nmpcunha, line 104: nit: blank line

addons/nuxeo-coldstorage/nuxeo-coldstorage.html
```diff
@@ -14,8 +14,16 @@
...
 limitations under the License.
 -->
-<nuxeo-slot-content name="moveToColdStorageDocumentAction" slot="DOCUMENT_ACTIONS" order="120">
+<nuxeo-slot-content name="moveToColdStorageDocumentContentAction" slot="DOCUMENT_ACTIONS" order="120">
   <template>
     <nuxeo-move-content-to-coldstorage-button document="[[document]]"></nuxeo-move-content-to-coldstorage-button>
   </template>
 </nuxeo-slot-content>
+
+<nuxeo-slot-content name="moveToColdStorageDocumentsContentsAction" slot="RESULTS_SELECTION_ACTIONS" order="-50">
```
nmpcunha, line 23: What about: `moveDocumentContentToColdStorageAction` and `moveSelectionContentToColdStorageAction` ? 🙂

i18n/messages.json
```diff
@@ -809,6 +809,13 @@
   "mimetype.text/plain": "Text",
   "mimetype.text/x-web-markdown": "Markdown",
   "mimetype.text/xml": "XML",
+  "moveDocumentsContentsToColdStorage.ask.confirmation": "Send the documents contents to cold storage?",
+  "moveDocumentsContentsToColdStorage.description": "This will result in increased file retrieval times and a low resolution preview.",
+  "moveDocumentsContentsToColdStorage.cancel": "Cancel",
+  "moveDocumentsContentsToColdStorage.confirm": "Confirm",
+  "moveDocumentsContentsToColdStorage.error": "Error moving the documents contents to cold storage.",
```
nmpcunha, line 816: Despite the approach we decide on server side, it may be a good option to say that some of the documents couldn't be moved. Since we may end up with the action performed to part of the selection 🤷‍♂

elements/nuxeo-document-attachments/nuxeo-document-attachments.js
RSalem07: my idea is to bloc any blob(s) modification once our document is in ColdStorage, even we transfer the main content into S3 and not the attachments.

i18n/messages.json
RSalem07: yes why not to go more further, currently to be consistent I did what we did on Trash Bulk. for example we don't move any of the selection if one of them is already move (the check using the method `every` see  
```
if (this.hasAdministrationPermissions(currentUser) || this.hasPermission(document, 'WriteColdStorage')) {
      return documents.every((doc) => !this.hasFacet(doc, 'ColdStorage') && this.hasContent(doc));
    }
```

and i thought to:

- even the user select all docs, we should only filter and try to move those that are not in cold storage, because can be knows about is as we don"t have any infos in the table to say that the doc is in cold storage.

- and yes we can improve messages to include the fact some docs are not moved because of error...

I think in these two cases PO point of view is welcome 

@jaubenque @lisamcintyre

nmpcunha: > we don't move any of the selection if one of them is already move

Indeed, just a small clarification: we don't **show** the action.
This approach it to try to minimize the probability of having errors on the UI.

addons/nuxeo-coldstorage/elements/nuxeo-move-contents-to-coldstorage-button.js
RSalem07: I introduced the ability to just move the documents on which he has the right permission and the same method is used when we call the server. 

- avoid any server erreur if the doc is already in Cold Storage.
- don't punish :)  the user to search doc by doc which one is in cold storage and which one is not, he can selects all docs and is up to us to filter which once should be send to the server

i18n/messages.json
RSalem07: yes, plz see [comment](https://github.com/nuxeo/nuxeo-web-ui/pull/866#discussion_r402234600)

elements/nuxeo-document-attachments/nuxeo-document-attachments.js
RSalem07: Julien asks to allow adding attachment even on document who's main content is in cold storage

i18n/messages.json
lisamcintyre: In the big picture, in fact, I would not point this out the to user as an "error" as that has a bad connotation and an idea that something went "wrong" when it fact, it just didn't do what we request because it had already been done (nothing "wrong" just work that didn't need to be done). indeed, some sort of filter to remove those items already flagged as being in glacier BEFORE the action is performed would be good.  Then, we could eventually have messaging that states:
- move of x docs has started.  
- y docs are not being moved as they are already in glacier (or y docs were removed from the request as they are already stored in glacier)

and then, eventually, some cue in the list (like i've seen in retention demos for hold items) would be interesting (we should establish a pattern similar to what we need to do for ribbons/banners).

addons/nuxeo-coldstorage/elements/nuxeo-move-contents-to-coldstorage-button.js
nmpcunha: I understand and point and I agree it improves UX.
Anyway, there are two small concerns:
- Not being totally consistent with the other "bulk" actions;
- The logic do not seem robust enough. 

What happens when the user is not admin and only have "WriteColdStorage" permission for 1 out of 5 documents? We allow him to "move" 5 documents, but in fact he/she can only move 1. Not sure if we want this kind of pattern... especially when we have "silent exceptions" on server side. It may generate more issues than "punishing" the user for selecting the correct docs (e.g. support issues because it is not working). 
Anyway none of the solutions being perfect 🤷‍♂ 

cc @lisamcintyre

RSalem07: I agree see https://github.com/nuxeo/nuxeo-web-ui/pull/866#discussion_r402578133

For the user without the right permission the doc will be filtered and we don't move it and perhaps showing a message that we can improve, see the Lisa suggestions below

i18n/messages.json
nmpcunha: > and then, eventually, some cue in the list (like i've seen in retention demos for hold items) would be interesting (we should establish a pattern similar to what we need to do for ribbons/banners).

We don't have this on Web UI by default and I'm not sure if it is a pattern to introduce in this context, since we would need to think about other scenarios and have also design's input for that.

> Then, we could eventually have messaging that states:

As referred [above](https://github.com/nuxeo/nuxeo-web-ui/pull/866#discussion_r402913642) it would improve the UX but it won't be consistent with other bulk action approaches (if it is for better I don't mind to start an inconsistency!). 
Anyway, it would be cool if we could have only one message stating that, otherwise we'll also introduce a different pattern of notifications which may not be a good option at this moment, IMHO.

RSalem07: IMHO we should add the filtering in web-ui to avoid blocking things or sending things to the server and we will find ourselves with a log errors for some reasons (no permission, already moved....)

on other hand for the message as it's more technical we can just keep things generic without a lot of details ....

addons/nuxeo-coldstorage/elements/nuxeo-move-contents-to-coldstorage-button.js
lisamcintyre: ok, i think an approach Nuno talks about is acceptable, in essence, based on what is selected you only see what actions are possible for the selected (basically, you filter the action based on the selected versus filtering the docs once the action is clicked.  basically, you should only see an action that can be applied to all selected.

addons/nuxeo-coldstorage/elements/nuxeo-move-contents-to-coldstorage-button.js
```diff
@@ -0,0 +1,144 @@
...
+  _canMoveDocument(doc) {
+    return (
+      (this.hasAdministrationPermissions(this.currentUser) || this.hasPermission(doc, 'WriteColdStorage')) &&
+      !this.hasFacet(doc, 'ColdStorage') &&
+      this.hasContent(doc)
+    );
+  }
+
+  _isAvailable(documents) {
+    return documents.some((doc) => this._canMoveDocument(doc));
```
nmpcunha, line 109: According to @lisamcintyre last comment, we should use `every`, right? 🙂

RSalem07: Yep, I will update it asap :)

PR #867, opened 2020-04-03
NXP-28858: Adding bulk Add and Remove enrichment icons

andreinechaev:


addons/nuxeo-ai-core/elements/actions/nuxeo-ai-bulk-add.js
```diff
@@ -39,7 +39,7 @@ class BulkAddButton extends mixinBehaviors([I18nBehavior, FiltersBehavior], Nuxe
         operation="Bulk.RunAction"
         input="[[provider]]"
         params='{ action: "bulkEnrich", parameters: "{}" }'
-        icon="icons:speaker-notes"
+        icon="nuxeo-ai:nuxeo-ai-bots-on"
```
nmpcunha, line 42: If it makes sense on AI context, a suggestion would be to drop `nuxeo-ai` from icon's ID since it is already present on [icon set name](https://github.com/nuxeo/nuxeo-web-ui/blob/829440c8a6e7033b4b5e01934ba8850314eb83a5/addons/nuxeo-ai-core/elements/nuxeo-ai-icons.js#L21).
So, every icon will have `nuxeo-ai:` as prefix 🙂

andreinechaev: sure, sounds good. thanks

PR #871, opened 2020-04-09
task-NXP-28044-UnitTestsStructure

nmpcunha:
Requires https://github.com/nuxeo/nuxeo-elements/pull/226


.eslintrc.js
```diff
@@ -78,5 +78,19 @@ module.exports = {
         'no-unused-expressions': 'off',
       },
     },
+    {
+      files: ['addons/**/test/**/*.test.js', 'elements/test/**/*.test.js'],
```
nelsonsilva, line 82: would rather have tests for WebUI at 'test/**/*.test.js' as I'd like at somve point to `mv elements src`.
If we don't have any yet perhaps we can delay adding this until then...

addons/nuxeo-coldstorage/test/nuxeo-move-content-to-coldstorage-button.test.js
```diff
@@ -0,0 +1,270 @@
...
+  flush,
+  html,
+  isElementVisible,
+  tap,
+  timePasses,
+  waitForEvent,
+} from '@nuxeo/testing-helpers';
+import '../elements/nuxeo-move-content-to-coldstorage-button.js';
+
+suite('Addons', () => {
```
nelsonsilva, line 29: Will this group all tests from the "Addons" suite ? Perhaps we can drop one of these suites and go with a convention for now, like "[Addon] Cold Storage" or something.

addons/nuxeo-coldstorage/test/nuxeo-move-content-to-coldstorage-button.test.js
```diff
@@ -0,0 +1,270 @@
...
+
+        test('Should not be visible when user is null', async () => {
+          button = await fixture(
+            html`
+              <nuxeo-move-content-to-coldstorage-button
+                .document=${document}
+              ></nuxeo-move-content-to-coldstorage-button>
+            `,
+          );
+          button.currentUser = null;
```
nelsonsilva, line 76: if we're doing lit-html and setting properties we should be able to do `.currentUser=${null}`, no ?

addons/nuxeo-coldstorage/test/nuxeo-move-content-to-coldstorage-button.test.js
```diff
@@ -0,0 +1,270 @@
...
+            ' permission on the document',
+          async () => {
+            button = await fixture(
+              html`
+                <nuxeo-move-content-to-coldstorage-button
+                  .document=${document}
+                ></nuxeo-move-content-to-coldstorage-button>
+              `,
+            );
+            button.currentUser = {
```
nelsonsilva, line 93: same, `const user = {..}; await fixture(... .currentUser=${user}... )`

addons/nuxeo-coldstorage/test/nuxeo-move-content-to-coldstorage-button.test.js
```diff
@@ -0,0 +1,270 @@
...
+      suite('Interactions', () => {
+        test('Should display a confirmation dialog when the button is pressed', async () => {
+          button = await fixture(
+            html`
+              <nuxeo-move-content-to-coldstorage-button
+                .document=${document}
+              ></nuxeo-move-content-to-coldstorage-button>
+            `,
+          );
+          await flush();
```
nelsonsilva, line 193: looking at all these `flush()` calls makes me thing we should have our own fixture in our test helpers calling the open-wc fixture followed by a flush()

karma.conf.js
```diff
@@ -0,0 +1,143 @@
...
+    middleware: ['static'],
+    static: {
+      path: path.join(process.cwd(), ''),
+    },
+    files: [
+      {
+        pattern: 'node_modules/@webcomponents/webcomponentsjs/webcomponents-bundle.js',
+        watched: false,
+      },
+      {
```
nelsonsilva, line 75: same, perhaps we can drop this for now

package.json
```diff
@@ -68,9 +86,12 @@
     "format": "npm run format:polymer && npm run format:prettier && npm run format:eslint",
     "win-format": "npm run win-format:polymer && npm run format:prettier && npm run format:eslint",
     "preinstall": "cd packages/nuxeo-web-ui-ftest && npm install",
-    "test": "node scripts/test-runner.js --report --screenshots --headless --tags='not @ignore'",
-    "test:dev": "node scripts/test-runner.js --nuxeoUrl=http://localhost:8080/nuxeo --url=http://localhost:5000/",
-    "test:watch": "node scripts/test-runner.js --nuxeoUrl=http://localhost:8080/nuxeo  --url=http://localhost:5000/ --tags='@watch' --debug"
+    "test": "npm-run-all -s unit-test functional-test",
```
nelsonsilva, line 89: I'd have this script running unit tests only and use `ftest` as prefix for scripts running functional tests for consistency with folder names, mvn profiles, etc... of course you'd have to update ftest/itests.xml to do `npm run ftest` and probably the README too

nelsonsilva: So: test, test:watch and ftest, ftest:dev, ftest:watch

package.json
```diff
@@ -164,8 +185,7 @@
   "lint-staged": {
     "*.{js,html}": [
       "prettier --write",
-      "eslint --fix",
-      "git add"
+      "eslint --fix"
```
nelsonsilva, line 188: we don't need to add after the `eslint --fix` to stage the fixes ?

addons/nuxeo-coldstorage/elements/nuxeo-move-contents-to-coldstorage-button.js
```diff
@@ -106,7 +106,7 @@ class BulkMoveToColdStorage extends mixinBehaviors([FiltersBehavior, FormatBehav
   }
 
   _isAvailable(documents) {
-    return documents.every((doc) => this._canMoveDocument(doc));
+    return documents.length > 0 && documents.every((doc) => this._canMoveDocument(doc));
```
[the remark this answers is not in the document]
nmpcunha: Thanks to unit tests 😉

.eslintrc.js
nmpcunha: Indeed.
I created a dummy unit test on elements and then left it there. Anyway, your suggestion seems more generic. Thanks! 👍

addons/nuxeo-coldstorage/test/nuxeo-move-content-to-coldstorage-button.test.js
nmpcunha: Yes, the idea was to group tests by suite on our reports.
```
Addons
    Cold Storage
      nuxeo-move-content-to-coldstorage-button
        Visibility
          ✔ Should ...
          REST Calls
            ✔ Should ...
      nuxeo-move-contents-to-coldstorage-button
        Visibility
          ✔ Should ...
        Interactions
          ✔ Should ...
          REST Calls
            ✔ Should ...
    Retention Management
    ....
    Live Connect
    ....
```

I'm ok to go with both approaches. 🤷‍♂️ 
Let's wait for more feedback to see if anyone has a different proposal for the report's structure.

addons/nuxeo-coldstorage/test/nuxeo-move-content-to-coldstorage-button.test.js
nmpcunha: More or less, i.e., generically yes, but... the `currentUser` is returned by `nuxeo-connection` (on connected). 
We could mock it on our `nuxeo-mock-client`, but it would imply to create different instances in each test and I found easier to use this approach.

addons/nuxeo-coldstorage/test/nuxeo-move-content-to-coldstorage-button.test.js
nmpcunha: Indeed, we'll need to have this approach frequently.

package.json
nmpcunha: Hmm... 🤔 
And our pipelines to run both, right?

package.json
nmpcunha: No, with the new version of this library it is done automatically. 😉

addons/nuxeo-coldstorage/test/nuxeo-move-content-to-coldstorage-button.test.js
nelsonsilva: But with tests in separate files do we get that nesting or to we end up with multiple Addons suites?

addons/nuxeo-coldstorage/test/nuxeo-move-content-to-coldstorage-button.test.js
nelsonsilva: Ok, I see 👍

package.json
nelsonsilva: Yes, ideally on different stages to have different PR checks

addons/nuxeo-coldstorage/test/nuxeo-move-content-to-coldstorage-button.test.js
nmpcunha: The reporter merges them. 😉

addons/nuxeo-coldstorage/test/nuxeo-move-content-to-coldstorage-button.test.js
nmpcunha: Ticket created: https://jira.nuxeo.com/browse/ELEMENTS-1165

PR #872, opened 2020-04-10
NXP-28847: modify styles and add classes for button redesign

andreacornaglia:


themes/base.js
```diff
@@ -424,18 +424,29 @@ const template = html`
           color: var(--nuxeo-link-hover-color);
         }
 
+        --nx-button-base: {
+          font-weight: 600;
+          font-size: 1em;
```
semisse, line 429: Are we really sure we want to use a relative unit such em here? I would say it would be safer to user rem.

themes/base.js
```diff
@@ -424,18 +424,29 @@ const template = html`
           color: var(--nuxeo-link-hover-color);
         }
 
+        --nx-button-base: {
```
semisse, line 427: We should set a min-width (96px)

themes/base.js
```diff
@@ -424,18 +424,29 @@ const template = html`
...
+          font-weight: 600;
+          font-size: 1em;
+          text-transform: capitalize;
+          height: 40px;
```
semisse, line 431: Can we consider to use a min-height here instead of the hard coded height?

themes/base.js
```diff
@@ -424,18 +424,29 @@ const template = html`
...
+        }
+
         --nx-button-primary: {
+          @apply --nx-button-base;
           color: var(--nuxeo-button-primary-text);
```
semisse, line 436: every time we use a var we should set a default value as a safe measure

themes/base.js
```diff
@@ -424,18 +424,29 @@ const template = html`
...
-          font-weight: 700;
           background-color: var(--nuxeo-button-primary);
+          border-color: transparent;
```
semisse, line 438: Curious: Why do we need this?

elements/search/nuxeo-saved-search-actions.js
```diff
@@ -43,16 +43,27 @@ Polymer({
         cursor: pointer;
       }
       iron-icon {
-        margin-right: 16px;
+        margin-right: 4px;
```
semisse, line 46: We should start to use either margin-inline-start/end or margin-block-start/end to support different text directions

elements/search/nuxeo-saved-search-actions.js
```diff
@@ -43,16 +43,27 @@ Polymer({
...
       }
       paper-menu-button[hidden] {
         display: none !important;
       }
+
+      paper-button.primary {
+        @apply --nx-button-primary;
+        height: 32px;
```
semisse, line 54: As far as I could understand, we have two different button sizes: default and small. Should we consider to add another class to achieve that style?

themes/base.js
nmpcunha: I guess it was just migrated from `paper-button` style, but not sure if we need it at all...

themes/base.js
nmpcunha: It is a good question, I guess it would depend on the factor rational.
We already have that `min-width` applied directly to `paper-button`. 🤷‍♂️

themes/base.js
andreacornaglia: we can, but the padding/line height is making the height be greater than it. Will need to understand why but sure

themes/base.js
andreacornaglia: yeah I will check out to make sure that's relevant or I'll remove it.

elements/search/nuxeo-saved-search-actions.js
andreacornaglia: great point! I never used them before, will do some researching

elements/search/nuxeo-saved-search-actions.js
andreacornaglia: yes, I thought about that. maybe --nx-button-small-height

themes/base.js
andreacornaglia: good point, changed it to rem

themes/base.js
andreacornaglia: yes, the min-width is already added to paper-button so no need to add it here

themes/base.js
andreacornaglia: changed

themes/base.js
nmpcunha: The conversation is marked as resolved, but it may be helpful to discuss the factoring rational, since it can be done in several different ways. @semisse point is still valid

elements/search/nuxeo-saved-search-actions.js
```diff
@@ -43,16 +43,27 @@ Polymer({
...
+      paper-button.primary {
+        @apply --nx-button-primary;
+        @apply --nx-button-small;
+      }
+
+      paper-button.primary:hover,
+      paper-button.primary:focus {
+        @apply --nx-button-primary-hover;
+        @apply --nx-button-small;
+      }
```
nmpcunha, line 61: Assuming this is our first use case for the `small` button, we may want to think if we want this approach of redefining the `paper-button`. 
This one is valid, a little bit verbose when we have to apply this in several places. And since we are not applying any other styles, just applying the mixins, should we try to have other class for `paper-button` in order to define its size to small?

elements/search/nuxeo-saved-search-actions.js
```diff
@@ -43,16 +43,27 @@ Polymer({
         cursor: pointer;
       }
       iron-icon {
-        margin-right: 16px;
+        margin-inline-end: 4px;
```
nmpcunha, line 46: If I understood correctly, this should be only applied to icons inside buttons, right?
If so, probably we may want to migrate this style to other place.

elements/search/nuxeo-saved-search-actions.js
andreacornaglia: good idea, added a new class for small button

elements/search/nuxeo-saved-search-actions.js
andreacornaglia: added a variable for the spacing.

elements/document/nuxeo-document-import.js
```diff
@@ -424,7 +424,7 @@ Polymer({
         </div>
         <div class="buttons horizontal end-justified layout">
           <div class="flex start-justified">
-            <paper-button noink dialog-dismiss on-tap="_cancel" hidden$="[[_creating]]"
+            <paper-button noink dialog-dismiss on-tap="_cancel" hidden$="[[_creating]]" class="secondary"
```
nmpcunha, line 427: It seems we have other cancel button on this element. (see [here](https://github.com/nuxeo/nuxeo-web-ui/blob/master/elements/document/nuxeo-document-import.js#L565))

themes/base.js
```diff
@@ -85,31 +85,53 @@ const template = html`
...
           min-width: 96px;
         }
+
         paper-button:hover {
           @apply --nx-button-hover;
         }
+
         paper-button.primary {
           @apply --nx-button-primary;
-          border-color: transparent;
```
nmpcunha, line 93: It seems that the default is `rgba(0, 0, 0, 0.15)`, so I guess that was the reason why `transparent` was there.
Indeed that it is almost not visible, but not sure if we want to remove them.
WDYT?

themes/base.js
```diff
@@ -85,31 +85,53 @@ const template = html`
...
+        paper-button.small:focus {
+          @apply --nx-button-small;
+        }
+
+        paper-button.primary iron-icon,
+        paper-button.secondary iron-icon {
+          margin-inline-end: var(--nuxeo-button-icon-margin-end, 4px);
+          width: 16px;
+          height: 16px;
+        }
```
nmpcunha, line 129: Should we assume `paper-button iron-icon`?
Looking to https://nuxeowiki.atlassian.net/wiki/spaces/DT/pages/1097924848/Buttons it is still not clear for me.
@carolinagouveia was the button with icon guidelines thought for every button?

themes/base.js
```diff
@@ -424,43 +446,61 @@ const template = html`
           color: var(--nuxeo-link-hover-color);
         }
 
+        --nx-button-small: {
+          min-height: 32px;
```
nmpcunha, line 450: Given we are extracting a `--nx-button-base`, should we apply it here too?

themes/base.js
```diff
@@ -85,31 +85,53 @@ const template = html`
...
         }
+
+        paper-button.small {
+          @apply --nx-button-small;
+        }
+
+        paper-button.small:hover,
+        paper-button.small:focus {
+          @apply --nx-button-small;
+        }
```
nmpcunha, line 122: Do we need this? The behaviour on `hover` and/or `focus` is the same as the other buttons, right? Depending on its "main class", i.e. `primary, secondary`

themes/base.js
```diff
@@ -424,43 +446,61 @@ const template = html`
...
+          color: var(--nuxeo-button-primary-text, #ffffff);
+          background-color: var(--nuxeo-button-primary, #0066ff);
         }
 
         --nx-button-primary-hover: {
-          color: var(--nuxeo-button-primary-text);
-          background-color: var(--nuxeo-button-primary-focus);
+          @apply --nx-button-base;
+          color: var(--nuxeo-button-primary-text, #ffffff);
+          background-color: var(--nuxeo-button-primary-focus, #1f28bf);
```
nmpcunha, line 472: Shouldn't we define only what is different from `--nx-button-primary`. 
If I am seeing this correctly, it is only applied on hover state which means that we should only the different parts, no? Guess we don't need to "reapply" all the `--nx-button-base` and also `--nx-button-primary` context 🤷‍♂️

themes/base.js
carolinagouveia: Yes, since the label of the button uses a 14px font-size, then it should be paired with a 16px icon ([as here](https://nuxeowiki.atlassian.net/wiki/spaces/DT/pages/1098809364/Icons)) for balance. What is missing on the guidelines is the menu icon, which is also a button but I believe it is a paper-icon-button. I'll add that but if you see any place where these sizes would be troubling I'll review it. Not sure if I answered the question.

nmpcunha: Sorry, maybe I didn’t explain myself well.
The question was more: every button (paper-button) can have icon, right? Despite being primary or secondary.

nmpcunha:
Still believe that we are missing some addons (e.g. easyshare, csv, ...)
It may worth to look at them.

themes/base.js
andreacornaglia: good point! i'll add it back :)

themes/base.js
andreacornaglia: so to keep it easier i'm having small as an additional class to be used with primary or secondary

themes/base.js
andreacornaglia: yes, otherwise on hover/focus the element gets the default height (which is 40px)

themes/base.js
andreacornaglia: good point, i had the impression otherwise we were getting paper styles, but seems that they are really not needed - removed.

themes/base.js
andreacornaglia: yes but sometimes we're using paper-button for things that shouldn't get this styles, for example when creating a document, each tile with doc type is a paper-button but there we don;t want the icon to be that size.

addons/nuxeo-csv/elements/nuxeo-document-import-csv.js
```diff
@@ -46,7 +46,7 @@ Polymer({
       }
 
       .suggester {
-        background-color: var(--nuxeo-dialog-buttons-bar);
+        background-color: var(--nuxeo-other-buttons-background);
```
semisse, line 49: not sure to feel comfortable with the naming of it... which buttons are the other buttons?

semisse: besides that, don't forget to add a default color to the variable.

elements/nuxeo-cloud-services/nuxeo-oauth2-consumed-tokens.js
```diff
@@ -154,7 +154,9 @@ class OAuth2ConsumedTokens extends mixinBehaviors([TokenBehavior, FormatBehavior
           </form>
         </iron-form>
         <div class="buttons">
-          <paper-button id="cancel" name="cancel" noink dialog-dismiss>[[i18n('command.cancel')]]</paper-button>
+          <paper-button id="cancel" name="cancel" noink class="secondary" dialog-dismiss
+            >[[i18n('command.cancel')]]</paper-button
+          >
```
semisse, line 159: nit: why did it add a break to a new line before closing the paper-button tag?

elements/nuxeo-document-versions/nuxeo-document-create-version.js
```diff
@@ -50,10 +50,12 @@ Polymer({
       nuxeo-tag {
         cursor: pointer;
         @apply --nx-button-primary;
+        min-height: auto;
```
semisse, line 53: for curiosity: why do we need to reset min-height here?

themes/base.js
```diff
@@ -424,43 +450,59 @@ const template = html`
...
+          min-height: 32px;
+          padding: 5px 16px;
+        }
+
+        --nx-button-base: {
+          font-weight: 600;
+          font-size: 1rem;
+          text-transform: capitalize;
+          min-height: 40px;
+          padding: 9px 16px;
```
semisse, line 463: I'm seeing that some paddings are not respecting the defined grid of multiples of 4. Are we ok with this @carolinagouveia ?

themes/base.js
```diff
@@ -424,43 +450,59 @@ const template = html`
...
+        }
+
         --nx-button-primary: {
-          color: var(--nuxeo-button-primary-text);
-          font-weight: 700;
-          background-color: var(--nuxeo-button-primary);
+          @apply --nx-button-base;
+          color: var(--nuxeo-button-primary-text, #ffffff);
+          background-color: var(--nuxeo-button-primary, #0066ff);
+          border-color: transparent;
```
semisse, line 471: In order to align pixel perfect with the secondary buttons (which are only outlined by a stroke) I think we should include a border here as well, otherwise we will have this button two pixels smaller in height compared with the secondary ones.

addons/nuxeo-csv/elements/nuxeo-document-import-csv.js
andreacornaglia: hehe yeah I agree with the naming, i will think of something better :D

themes/base.js
carolinagouveia: Yes you're right, instead of 9px we should have 8px

elements/nuxeo-cloud-services/nuxeo-oauth2-consumed-tokens.js
andreacornaglia: not sure but i will adjust the format.

elements/nuxeo-document-versions/nuxeo-document-create-version.js
andreacornaglia: because otherwise it will inherit --nx-button-primary min-height which is set to 40px;

themes/base.js
andreacornaglia: the padding is calculated in relation to line-height and border to make the btn be the 40px

andreacornaglia: ok, changed it to 8px;

elements/document/nuxeo-document-import.js
```diff
@@ -74,7 +74,7 @@ Polymer({
       }
 
       .suggester {
-        background-color: var(--nuxeo-dialog-buttons-bar);
+        background-color: var(, rgba(0, 0, 0, 0.05));
```
nelsonsilva, line 77: missing `--nuxeo-input-background` here ?

elements/nuxeo-app.js
```diff
@@ -273,6 +273,48 @@ Polymer({
         padding: 0 24px;
         justify-content: space-between;
       }
+
+      #dialog paper-button.primary {
```
nelsonsilva, line 277: aren't these already in theme/base ? Why do we need it here too ? 🤔

themes/default/theme.html
```diff
@@ -65,14 +65,18 @@
       --nuxeo-button-primary: var(--nuxeo-primary-color);
       --nuxeo-button-primary-focus: var(--nuxeo-secondary-color);
       --nuxeo-button-primary-text: #ffffff;
+      --nuxeo-button-secondary-text: #0066ff;
       --nuxeo-button-disabled: rgba(0, 0, 0, 0.05);
       --nuxeo-button-disabled-text: rgba(0, 0, 0, 0.1);
       --nuxeo-link-color: var(--nuxeo-text-default);
       --nuxeo-link-hover-color: var(--nuxeo-primary-color);
 
+      --nuxeo-button-icon-margin-end: 4px;
```
nelsonsilva, line 74: this seems super specific, is the margin 4px everywhere? perhaps we could have `--nuxeo-button-icon-margin`

themes/default/theme.html
```diff
@@ -65,14 +65,18 @@
...
+
       /* Custom Backgrounds */
       --nuxeo-container-hover: var(--nuxeo-page-background);
-      --nuxeo-dialog-buttons-bar: rgba(0, 0, 0, 0.05);
+      --nuxeo-dialog-buttons-bar: transparent;
+      --nuxeo-input-background: rgba(0, 0, 0, 0.05);
```
nelsonsilva, line 79: nit: since we do have a `nuxeo-input` element this makes it sound like it only defines the background color of nuxeo-input which is not the case as it's meant to be use by any HTML input control.

elements/document/nuxeo-document-import.js
andreacornaglia: ops :)

elements/nuxeo-app.js
nmpcunha: @andreacornaglia , can you provide us more info about this one?
Which was the use case not covered that led to the need of redefining these styles here?

themes/base.js
nmpcunha: Indeed, but the styles are being redefined there, right?
Would you be able to try without `primary` and `secondary` to double check if it works properly?

themes/base.js
nmpcunha: But I guess we may want to have "small" buttons despite being `primary` or `secondary`

themes/base.js
nmpcunha: Can you please point to an use case in order to be easier to understand why it is happening? Thanks!

themes/base.js
```diff
@@ -85,31 +85,52 @@ const template = html`
         paper-button {
           min-width: 96px;
         }
+
         paper-button:hover {
           @apply --nx-button-hover;
         }
+
+        dialog-confirm,
```
nmpcunha, line 93: Do we have this kind of element? `dialog-confirm`?
I can only find a `paper-button` property/attribute

elements/nuxeo-document-versions/nuxeo-document-create-version.js
nmpcunha: It may worth to do the exercise of extracting what we really need for having the `nuxeo-tag` looking we want here.
Maybe we don't need most of the `--nx-button-primary` styles and we could drop the mixin usage. WDYT?

themes/default/theme.html
```diff
@@ -80,8 +80,8 @@
 
       /* Warn, Info, Error */
       --nuxeo-action-color-activated: #00aded;
-      --nuxeo-warn-text: #fb6107;
-      --nuxeo-validated: #99d749;
+      --nuxeo-warn-text: #de350b;
```
nmpcunha, line 83: Should we apply on the other themes, right? `kawaii`, `light`

themes/base.js
andreacornaglia: inded, seems like it's redundant, removing it

themes/base.js
andreacornaglia: yes but we're making a compound class, so the btn will get the base form the primary/secondary class, and the height from the small one. does that make sense?

elements/nuxeo-document-versions/nuxeo-document-create-version.js
andreacornaglia: ok, i need to check with Carolina what is the design guideline for this

themes/base.js
andreacornaglia: yeah not sure why it's there, will remove it.

themes/default/theme.html
andreacornaglia: not sure about those, we have guidelines for dark and default.

elements/nuxeo-app.js
andreacornaglia: all the nuxeo-dialogs auto-generated are not getting the styles.

andreacornaglia: there are a lot of styles applied directly at this place, which makes me wonder why it's not getting the styles otherwise.

themes/base.js
nmpcunha: It makes sense. But other approaches may also make sense here... 😅 
It really depends on how we want to tackle this.

I'm still thinking on what is our drive 🤔 
If I'm seeing this correctly:
- `primary` and `secondary` are driving the button style, since are only them applying `--nx-button-base`. 
Do we want to assume this? Or do we want every `paper-button` on Web UI to have the base style (font sizes, min dimensions, etc.)?
-

themes/default/theme.html
andreacornaglia: as discussed with Carolina, applying same as default.

elements/nuxeo-document-versions/nuxeo-document-create-version.js
andreacornaglia: as discussed, we'll have another ticket to discuss substituting this nuxeo-tag for a text btn

themes/base.js
```diff
@@ -85,31 +85,47 @@ const template = html`
...
         paper-button.secondary[disabled] {
           @apply --nx-button-secondary-disabled;
         }
+
+        paper-button.small {
+          @apply --nx-button-small;
+        }
+
+        paper-button iron-icon {
+          margin-inline-end: var(--nuxeo-button-icon-margin, 4px);
```
nmpcunha, line 120: About this `margin-inline-end`, two small questions:
- I assume we are confident that it will [become standard](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline-end) in order to start using it;
- The variable `--nuxeo-button-icon-margin` is a little more broad than the css property, right? Also because it can be combined with `writing-mode`, `direction`, etc.

What did you had in mind, @nelsonsilva ?

themes/base.js
andreacornaglia: we have paper-button play different roles. one of them is to be on dialogs and other places as primary or secondary, but we also have btns that have a different style - that's why we are adding those classes sepparately.

themes/base.js
```diff
@@ -84,6 +84,7 @@ const template = html`
         /* paper-button */
         paper-button {
           min-width: 96px;
+          @apply --nx-button-base;
```
nmpcunha, line 87: With this kind of approach we may want to drop `--nx-button-base`. Or do we want to expose it so people can override it?

Also, the `min-width: 96px;` seems to be on both: `paper-button` and `--nx-button-base`

themes/base.js
andreacornaglia: as discussed, we are applying now btn base to all paper-btns.

nelsonsilva: guess this is a common pattern when you also have a `btn` class (ie `class="btn primary small"`). In this case since we don't we apply the base to all <paper-button>.
If we were to apply the base to small we would have to it also for primary and secondary.

themes/base.js
nelsonsilva: I'm ok with a variable to define the margin, previously it was just about the end thus my remark iirc

themes/base.js
nelsonsilva: @andreacornaglia @nmpcunha shouldn't we get rid of `--nx-button-base` as just update the existing `--paper-button` https://github.com/nuxeo/nuxeo-web-ui/blob/master/themes/base.js#L557 ?

themes/base.js
andreacornaglia: fixed.

themes/base.js
andreacornaglia: @nmpcunha the support looks pretty well, except for ie. About the other icon properties like weight and height we can add them in that variable if we can add them I guess

andreacornaglia: added a class for paper-button-icon

themes/base.js
```diff
@@ -556,15 +532,22 @@ const template = html`
...
+          text-transform: capitalize;
+          min-height: 40px;
+          padding: 8px 16px;
+          min-width: 96px;
           border-radius: 0.1em;
-          padding: 0.8em 2em;
           margin: 0;
         }
 
+        --paper-button-icon: {
```
nmpcunha, line 545: Some concerns about this one (see more on [elements PR](https://github.com/nuxeo/nuxeo-elements/pull/227/files#r416524191))

nmpcunha: And, as far as I can understand, it is not being used anywhere in Web UI 🤔 
Can you please point to some of its usages?

elements/nuxeo-document-bulk-actions/nuxeo-add-to-collection-documents-button.js
```diff
@@ -38,7 +39,7 @@ import { html } from '@polymer/polymer/lib/utils/html-tag.js';
 */
 Polymer({
   _template: html`
-    <style include="nuxeo-action-button-styles">
+    <style include="nuxeo-action-button-styles nuxeo-button-styles">
```
nmpcunha, line 42: Should we still refer `nuxeo-button-styles` here? Or it is safe to consider `nuxeo-styles`?

elements/directory/nuxeo-vocabulary-management.js
```diff
@@ -42,7 +43,7 @@ const schemaDataCache = {};
 */
 Polymer({
   _template: html`
-    <style>
+    <style include="nuxeo-button-styles">
```
nmpcunha, line 46: Same question about using `nuxeo-button-styles` _vs_ `nuxeo-styles`.

themes/base.js
andreacornaglia: we are using it on nuxeo-elements, which at the same time is used on web-ui.

elements/nuxeo-document-bulk-actions/nuxeo-add-to-collection-documents-button.js
andreacornaglia: yes, good point, added nuxeo-styles and it works fine.

themes/base.js
nmpcunha, line 545: Still have concerns about this name (as referred [here](https://github.com/nuxeo/nuxeo-elements/pull/227/files#r416528534)).
Maybe something along `--nx-button-icon` would avoid misunderstandings with Polymer Elements mixins.
Anyway, let's wait for more opinions on this 🙂

andreacornaglia: renamed it

themes/default/theme.html
```diff
@@ -65,19 +65,23 @@
...
       --nuxeo-link-color: var(--nuxeo-text-default);
       --nuxeo-link-hover-color: var(--nuxeo-primary-color);
 
+      --nuxeo-button-icon-margin: 4px;
+
       /* Custom Backgrounds */
       --nuxeo-container-hover: var(--nuxeo-page-background);
-      --nuxeo-dialog-buttons-bar: rgba(0, 0, 0, 0.05);
+      --nuxeo-dialog-buttons-bar: transparent;
+      --input-background: rgba(0, 0, 0, 0.05);
```
semisse, line 79: some as above

semisse:
Made some more suggestions, but overall it looks good. This is a huge step to ensure consistency across our entire interface, though we will still to design and implement some states in a future iteration (focus, activated).

themes/base.js
```diff
@@ -424,43 +396,47 @@ const template = html`
           color: var(--nuxeo-link-hover-color);
         }
 
+        --nx-button-small: {
+          min-height: 32px;
+          padding: 5px 16px;
```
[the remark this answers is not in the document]
andreacornaglia: changed :)

themes/dark/theme.html
```diff
@@ -65,19 +65,21 @@
...
       --nuxeo-button-disabled: rgba(255, 255, 255, 0.05);
       --nuxeo-button-disabled-text: rgba(255, 255, 255, 0.1);
       --nuxeo-link-color: var(--nuxeo-text-default);
       --nuxeo-link-hover-color: var(--nuxeo-primary-color);
 
       /* Custom Backgrounds */
       --nuxeo-container-hover: var(--nuxeo-page-background);
-      --nuxeo-dialog-buttons-bar: var(--nuxeo-quicksearch-background);
+      --nuxeo-dialog-buttons-bar: transparent;
+      --input-background: var(--nuxeo-quicksearch-background);
```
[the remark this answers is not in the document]
andreacornaglia: i did have it before but Nelson had a comment around the background being able to be used for any input element, so we decided not to have the prefix nuxeo for that reason.

themes/base.js
```diff
@@ -424,43 +396,47 @@ const template = html`
...
+          background-color: transparent;
         }
 
         --nx-button-disabled: {
-          color: var(--nuxeo-button-disabled-text);
-          font-weight: normal;
-          background-color: var(--nuxeo-button-disabled);
+          color: var(--nuxeo-button-disabled-text, rgba(0, 0, 0, 0.1));
+          background-color: var(--nuxeo-button-disabled, rgba(0, 0, 0, 0.05));
+          border: none;
```
nmpcunha, line 436: Maybe a cool next step would be to have consistency on how we define the css properties (e.g. order) 🙂 
But it is not important for now. Anyway, I think @semisse will have a input about that 😉

PR #882, opened 2020-04-16
NXP-28806: add missing space in retention duration

gcarlin:


addons/nuxeo-retention/document/retentionrule/nuxeo-retentionrule-view-layout.html
```diff
@@ -80,11 +80,11 @@
...
-        <div role="widget">
+        <div role="widget" style="margin-right:10px">
           <label>[[i18n("retention.rule.label.duration.years")]]</label>
           <div name="years">[[document.properties.retention_def:durationYears]]</div>
         </div>
-        <div role="widget">
+        <div role="widget" style="margin-right:10px">
           <label>[[i18n("retention.rule.label.duration.months")]]</label>
           <div name="months">[[document.properties.retention_def:durationMonths]]</div>
         </div>
```
RSalem07, line 90: I am wondering if we can add the same thing that you did on `10.10`, [here](https://github.com/nuxeo/nuxeo/pull/3925/commits/eff3048992b4bf467f06cbc8037040b73a3c12c7) I think we can even add the style part:
```
  *[role=widget] {
        margin-bottom: 16px;
        margin-right: 10px;
      }
```

gcarlin: That was my original idea but I really worry about side effects. In the 10.10 branch, this style is applies specifically on the layout so I guess there must be a good reason

nmpcunha: For master we've extracted that `margin-bottom` that is being applied to all widgets to our style base (more details [here](https://github.com/nuxeo/nuxeo-web-ui/commit/0bfc3943cb5f907efde4f2a7c64a958a8cb8eb97)).

No side effect by using @RSalem07 approach.
Anyway, you can simplify it to:
```
*[role=widget] {
     margin-right: 10px;
}
```
But please keep in mind that this will apply the margin to all widgets.

You could also give an id/class to months widget div and apply `margin: 0 10px;`. It would generate the same end result without extra margins on other divs.

PR #884, opened 2020-04-17
NXP-28787: add missing translation label for Nuxeo Drive

BoboTiG:
The `driveSyncRootsManagement.root.disable` label was lost at some point.
Reintroducing it to solve the tooltip display on the delete button of a synchronized root.

It is used here: https://github.com/nuxeo/nuxeo-web-ui/blob/c7e67299e9c35803d8c25104324ba37e54787f4d/addons/nuxeo-drive/elements/nuxeo-drive-sync-roots-management.js#L103-L107

Signed-off-by: Mickaël Schoentgen <mschoentgen@nuxeo.com>


BoboTiG:
@nmpcunha thanks for the review. I applied changes, this is better now :)

nmpcunha:
> @nmpcunha thanks for the review. I applied changes, this is better now :)

Thank you! 👍

PR #886, opened 2020-04-17
fix-NXP-28895-retentionRuleDuration

nmpcunha:


addons/nuxeo-retention/document/retentionrule/nuxeo-retentionrule-create-layout.html
```diff
@@ -220,6 +220,19 @@
          */
         document: Object,
       },
+
+      validate() {
+        const years = this.shadowRoot.querySelector('nuxeo-input[name="years"]');
+        const months = this.shadowRoot.querySelector('nuxeo-input[name="months"]');
+        const days = this.shadowRoot.querySelector('nuxeo-input[name="days"]');
+        if (parseInt(years.value, 10) + parseInt(months.value, 10) + parseInt(days.value, 10) === 0) {
```
RSalem07, line 228: I am wondering if we can use a third library like `moment.js` (if available on 10.10) and we also ensure that the date will be in the future:

thing like:

`console.log(moment().day(0).month(0).year(2022).diff(new Date()));`
https://momentjs.com/docs/#/displaying/difference/

nmpcunha: See your point, but not sure if worths the effort.
These inputs represent the _number of years_ and not the _year_ (same for months and days).
This means that we would need to do something like: 
```
moment().add(years.value, 'years').add(months.value, 'months').add(days.value, 'days').diff(moment()) // or "new Date()"
```

Not sure if easier to read or not, but it is possible and does the work.
Conceptually, we are not validating dates, but three integer inputs and, at least one, must be greater than zero. I would be ok with any of the approaches 🤷‍♂️

andreacornaglia: is there a possibility that there is no value or that the user inputs something that is not a number and makes the parseInt fail?

addons/nuxeo-retention/document/retentionrule/nuxeo-retentionrule-edit-layout.html
```diff
@@ -220,6 +220,19 @@
...
+      validate() {
+        const years = this.shadowRoot.querySelector('nuxeo-input[name="years"]');
+        const months = this.shadowRoot.querySelector('nuxeo-input[name="months"]');
+        const days = this.shadowRoot.querySelector('nuxeo-input[name="days"]');
+        if (parseInt(years.value, 10) + parseInt(months.value, 10) + parseInt(days.value, 10) === 0) {
+          years.invalid = true;
+          months.invalid = true;
+          days.invalid = true;
+          return false;
+        }
```
andreacornaglia, line 233: should we possibly have a validation module so we make sure the checks are all consistent?

addons/nuxeo-retention/document/retentionrule/nuxeo-retentionrule-create-layout.html
nelsonsilva: we should just check if they're all equal to zero and drop the date math ;)

nmpcunha: Fair enough, @nelsonsilva! 😂 

@andreacornaglia , there is always chance to break stuff, specially in JS 🙂 
But if you look to the layout, we are using `nuxeo-input` with well defined properties that should help us on that. 😉

addons/nuxeo-retention/document/retentionrule/nuxeo-retentionrule-edit-layout.html
nmpcunha: Can you please elaborate a little bit more? 🤔

PR #892, opened 2020-04-18
NXP-28953: add isrecord column in document listing

gcarlin:


gcarlin:
This property is present independently of the addon installation, hence why Ive added it

gcarlin:
To follow up:

- This property in on the system schema, so not dependent of the addon installation
- This property is fundamental for people using the Retention addon (this is the first point I've noticed)
- The column is hidden, so it is only about giving a way to the users to display it (too tricky to implement)
- There are plenty of properties in the default document listing which are never used by our clients (`dc:nature`, `dc:subjects`, `dc:coverage`)

elements/nuxeo-results/nuxeo-default-results.js
```diff
@@ -217,6 +217,16 @@ Polymer({
...
           </template>
         </nuxeo-data-table-column>
+        <nuxeo-data-table-column
+          name="[[i18n('documentContentView.datatable.header.isRecord')]]"
+          field="ecm:isRecord"
+          hidden
+          flex="60"
+        >
+          <template>
+            <paper-checkbox disabled checked="[[item.properties.ecm:isRecord]]"></paper-checkbox>
```
andreacornaglia, line 227: note that this is not working because ecm:isRecord is not available on the API response. It is actually returned as item.isRecord

PR #897, opened 2020-04-21
NXP-28846: replace doctype icons

carolinagouveia:


RSalem07:
cool  stuff :). Just wondering about adding a new line at the end of each file. I am not sure about the rule on web-ui repo (on platform what we do is adding a new separate commit to clean things and add the missing break line at the end of each file)

PR #901, opened 2020-04-28
NXP-28446: Use Platform Helm Chart for Preview

doulba:
Update WebUI helm chart dependencies

 - use https://github.com/nuxeo/nuxeo-helm-chart

### DIFF
--- a/charts/preview/values.yaml
+++ b/charts/preview/values.yaml
@@ -3,6 +3,7 @@ expose:
     helm.sh/hook: post-install,post-upgrade
     helm.sh/hook-delete-policy: hook-succeeded
   config:
+    # add tls configuration
     domain: webui.dev.nuxeo.com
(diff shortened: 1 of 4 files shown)


charts/preview/values.yaml
```diff
@@ -18,28 +19,35 @@ cleanup:
...
+      labels:
+        branch: "$BRANCH_NAME"
+        team: webui
+        resource: service
+        usage: preview
+    image:
+      repository:
+      tag:
+      pullPolicy: Always
+    packages: nuxeo-dam nuxeo-web-ui-$PREVIEW_VERSION
```
nmpcunha, line 44: Not sure if we need `nuxeo-dam` since it is already part of Platform and Web UI.

doulba: will be removed certainly

PR #902, opened 2020-04-28
Nxp 28939 improvement split retention starting point prop

RSalem07:
PR open manually, related to the platform part: https://github.com/nuxeo/nuxeo/pull/3978

T&P In Progress: https://qa.nuxeo.org/jenkins/job/master/job/web-ui-saouana/9/


RSalem07:
I think we can squash the two commit, I left them separated because on `10.10` it was the case but on `10.10` the two repo (web-ui/platform) are separated. @nmpcunha wdyt

PR #906, opened 2020-04-30
NXP-28862: modify date format to show time for nuxeo document history and audit

andreacornaglia:


elements/nuxeo-document-history/nuxeo-document-history.js
```diff
@@ -71,7 +71,7 @@ Polymer({
           <template>[[_formatActivity(item.eventId)]]</template>
         </nuxeo-data-table-column>
         <nuxeo-data-table-column name="[[i18n('documentHistory.date')]]" sort-by="eventDate">
-          <template><nuxeo-date datetime="[[item.eventDate]]"></nuxeo-date></template>
+          <template><nuxeo-date datetime="[[item.eventDate]]" format="LLL"></nuxeo-date></template>
```
nmpcunha, line 74: With this change the tooltip will show exactly the same as the span.
Is this a pattern we want to promote from an UX perspective? (cc @carolinagouveia , @semisse )

andreacornaglia: by span you mean tooltip? I think so.

nmpcunha: I was just stating that we will have the same content information on both: span and tooltip.

nmpcunha: Screenshot 2020-05-04 at 10 49 42

andreacornaglia: hiding tooltip when values are same for date: https://github.com/nuxeo/nuxeo-elements/pull/233/files

PR #907, opened 2020-05-05
NXP-23346 multirepo support

nelsonsilva:


nelsonsilva:
> LGTM!
> Should we introduce some unit tests for the new element?

I will have to try and setup funcional tests for the new multi repo route because it's likely to be backported. Not sure yet how because I only want to run these when deployed it on a Nuxeo server, not when running against the dev server.

As for unit tests, given this is just a menu of hardcoded values injected in the index page I don't see huge value tbh as what I really need is to test the integration, meaning we have the right values and I can switch between repos.

elements/nuxeo-browser/nuxeo-repositories.js
```diff
@@ -0,0 +1,89 @@
...
+  static get properties() {
+    return {
+      hidden: {
+        type: Boolean,
+        reflectToAttribute: true,
+        computed: '_isHidden(repositories)',
+      },
+      repositories: {
+        type: Array,
+        value: Nuxeo.UI.repositories || [],
```
Gabez0r, line 67: Since we're dealing with an object here (or a reference to it), if you change the `repositories` property it will actually mutate the settings object (`Nuxeo.UI.repositories`) as well, which is unwanted. Perhaps we could either make this property private or/and do a deep clone using `JSON.parse(JSON.stringify(...))` (because this is an array of objects)?

nelsonsilva: Good point, guess we already have the issue with `Nuxeo.UI.config.enrichers` for instance...

PR #910, opened 2020-05-12
include advanced audit filtering on doc history

RSalem07:
PR opened Manually,

T&P OK: https://qa.nuxeo.org/jenkins/job/master/job/web-ui-saouana/12/
This PR is also related to the element one: https://github.com/nuxeo/nuxeo-elements/pull/235
JIRA is: [NXP-28863](https://jira.nuxeo.com/browse/NXP-28863)

--- 

T&P In Progress (after refactoring and reworking things):
https://qa.nuxeo.org/jenkins/job/master/job/web-ui-saouana/13/


RSalem07:
Yes, I thought the same thing, and I asked myself the same interrogation, but there a part which is mainly the same, the html part. But for the JS part these a few differences, mainly in the build of params:

- audit history sends directly the date where I tried to reproduce the check that we had on history document (suppression of `_observeEndDate`....)
- also in the audit we send the events categories separated by a `,`, where it is not the same on the document history (backend part expect an array)

and in other hand we have two different messages (those ones can be modified).

this is way I kept the thing separated here.

elements/nuxeo-document-history/nuxeo-document-history.js
```diff
@@ -38,31 +38,61 @@ import moment from '@nuxeo/moment';
...
         @apply --layout-horizontal;
-        @apply --layout-end-justified;
+        @apply --layout-wrap;
       }
 
-      #table {
-        height: calc(100vh - 370px);
+      .row-container * {
+        flex: 1 0 0;
+        margin-right: 8px;
```
semisse, line 48: Maybe we could replace this hardcoded margin with something like `margin-block-start` or `margin-block-end` so we can display it correctly on RTL languages?

elements/nuxeo-document-history/nuxeo-document-history.js
```diff
@@ -101,60 +131,70 @@ Polymer({
       value: false,
       observer: '_refresh',
     },
+    principalName: {
```
semisse, line 134: Just a question for the sake of curiosity: what does it mean, in this case, principalName?

semisse:
Hey Salem! just left a nitpick comment and a question really. Looks great to me!

elements/nuxeo-document-history/nuxeo-document-history.js
RSalem07: thanks for the review :),

the principal will be user of the action in the audit. The improvement has for purpose to add the ability to filter by the user.

RSalem07:
I add two new commits to propose two solution to refactor things between the two elements, `nuxeo-audit` and `nuxeo-document-history`:

- 1) [NXP-28863: add audit behavior ](https://github.com/nuxeo/nuxeo-web-ui/pull/910/commits/7cacf42e71564c76ea45c70ac3c50ed29df10743): This one keep the two elements with there own UI part. But the logical things are moved to the behaviour. The two elements then add the behaviour.

- 2) [NXP-28863: WIP: reuse one audit elemen](https://github.com/nuxeo/nuxeo-web-ui/pull/910/commits/a03f6bda812a8ba767ef32285ae5e58e1789951c). This is mainly  a WIP. My idea here is to introduce new element (`nuxeo-audit-search`) where we put the whole logic (if a document is available then we use the REST API to get the history doc, otherwise an operation to the admin audit). To simplify things I introduce a new elements but we can simply use the `nuxeo-audit` with the improvement that I did on `nuxeo-audit-search`. and then up to us to see if the `nuxeo-document-history` use the `nuxeo-audit` or to call it directly in `nuxeo-web-ui-bundle.html`. 


As discussed with Nuno, for the second point even I didn't terminate the whole stuff I push it to get more comments/suggestions.

elements/nuxeo-admin/nuxeo-audit.js
```diff
@@ -64,7 +64,9 @@ Polymer({
       }
     </style>
 
-    <nuxeo-audit-page-provider id="provider" page-size="40" provider-name="EVENTS_VIEW"></nuxeo-audit-page-provider>
+    <nuxeo-audit-page-provider id="provider" page-size="40" provider-name="[[providerName]]">
```
nmpcunha, line 67: It seems that we don't map `provider-name` on `nuxeo-audit-page-provider`. Guess it was just obsolete code there... But it seems that someone is asking for that 😄  ([ELEMENTS-1078](https://jira.nuxeo.com/browse/ELEMENTS-1078))

elements/nuxeo-web-ui-bundle.html
```diff
@@ -66,7 +66,8 @@
   <template>
     <nuxeo-filter user="[[user]]" expression="user.isAdministrator">
       <template>
-        <nuxeo-audit name="audit" id="audit"></nuxeo-audit>
+        <!--<nuxeo-audit name="audit" id="audit"></nuxeo-audit>-->
+        <nuxeo-audit-search name="audit" id="audit" provider-name="EVENTS_VIEW"></nuxeo-audit-search>
```
nmpcunha, line 70: Given the requirements of having the same stuff on both screens, I guess this is a sensible solution.
In fact, the elements are pretty the same thing and they have a lot of duplicated code.
(And, personally, I prefer this instead of the `auditBehavior` option you've provided)

@nelsonsilva , @Gabez0r , WDYT?
Do you foresee any issue by merging both elements?

elements/nuxeo-audit/nuxeo-audit-search.js
```diff
@@ -0,0 +1,238 @@
...
+    return this.document ? this.document.uid : '';
+  }
+
+  getTable() {
+    return this.$.table;
+  }
+
+  formati18n(path, key) {
+    return key ? this.i18n(path + key) : '';
+  }
```
nmpcunha, line 176: This should be a private method and we can probably reuse the `nuxeo-document-history` one since it is already scopped to `activity`. Also, the fallback for missing translations is already handled by `i18n` method which will return the key if no value is found.

elements/nuxeo-audit/nuxeo-audit-search.js
```diff
@@ -0,0 +1,238 @@
...
+    };
+  }
+
+  static get observers() {
+    return ['_refresh(startDate, endDate, selectedEventTypes.*, selectedEventCategory, principalName)'];
+  }
+
+  getDocumentId() {
+    return this.document ? this.document.uid : '';
+  }
```
Gabez0r, line 168: Consider using a getter:
```suggestion
  get documentId() {
    return this.document ? this.document.uid : '';
  }
```
usage: `this.documentId`

elements/nuxeo-audit/nuxeo-audit-search.js
```diff
@@ -0,0 +1,238 @@
...
+
+  getTable() {
+    return this.$.table;
+  }
```
Gabez0r, line 172: A getter here would provide a cleaner API, and we're just returning another property of an object anyway:
```suggestion
  get table() {
    return this.$.table;
  }
```

elements/nuxeo-web-ui-bundle.html
Gabez0r: We will still need an element to wrap `nuxeo-audit-search` in a `nuxeo-page`, otherwise the element will be broken when contributed to page slots ([ADMINISTRATION_PAGES](url), `PAGES`, etc.). I'd suggest:
- `nuxeo-audit-search` becomes `nuxeo-document-history` (not sure we can just kill the element like that)
- `nuxeo-audit` remains, but this time it just as a wrapper of `nuxeo-document-history`, while exposing the same API (similar to what we did with `nuxeo-results-view` and `nuxeo-search-page`, which is a wrapper of the former).
- this means we wouldn't need to have a behavior

Gabez0r: I'd also be ok with not renaming `nuxeo-audit-search` to `nuxeo-document-history` if we agree on killing the `nuxeo-document-history` element. But we'll still need the wrapper.

elements/nuxeo-audit/nuxeo-audit-search.js
```diff
@@ -0,0 +1,238 @@
...
+          </div>
+        </template>
+      </nuxeo-card>
+      <nuxeo-card>
+        <nuxeo-data-table id="table" paginable nx-provider="provider" empty-label="[[i18n('documentHistory.empty')]]">
+          <nuxeo-data-table-column name="[[i18n('documentHistory.performedAction')]]" sort-by="eventId">
+            <template>[[formati18n('activity.', item.eventId)]]</template>
+          </nuxeo-data-table-column>
+          <nuxeo-data-table-column name="[[i18n('documentHistory.date')]]" sort-by="eventDate">
+            <template><nuxeo-date datetime="[[item.eventDate]]"></nuxeo-date></template>
```
nmpcunha, line 98: Please make sure you rebase on master, since these elements were changed recently (https://github.com/nuxeo/nuxeo-web-ui/pull/906)

elements/nuxeo-web-ui-bundle.html
nelsonsilva: Yes, please, no `AuditBehavior`, prefer composition over inheritance.

nelsonsilva: Just so I understand, the plan would be to drop `nuxeo-document-history` and rely solely on `nuxeo-audit-search`  (or the other way around), correct?

Gabez0r: I'd say so, and then make `nuxeo-audit` use it as well, just as a wrapper with a `nuxeo-page`.

elements/nuxeo-document-history/nuxeo-document-history.js
```diff
@@ -38,31 +38,61 @@ import moment from '@nuxeo/moment';
...
+      .row-container {
         @apply --layout-horizontal;
-        @apply --layout-end-justified;
+        @apply --layout-wrap;
       }
 
-      #table {
-        height: calc(100vh - 370px);
+      .row-container * {
+        flex: 1 0 0;
```
nmpcunha: Guess `flex: 1` would be enough 🤔

elements/nuxeo-document-history/nuxeo-document-history.js
nmpcunha: Sorry for getting back to this, @semisse. Could you explain a little bit more?
If I understood correctly we should go with `margin-inline-end` (and just for the first child) in order to get the following look:
Screenshot 2020-05-27 at 15 15 12

Am I seeing this correctly?

elements/nuxeo-admin/nuxeo-audit.js
```diff
@@ -36,182 +28,17 @@ import { FormatBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-format-behavior.j
...
-            <nuxeo-data-table-column name="[[i18n('audit.comment')]]">
-              <template>[[item.comment]]</template>
-            </nuxeo-data-table-column>
-          </nuxeo-data-table>
-        </nuxeo-card>
-      </div>
-    </nuxeo-page>
+    <nuxeo-card>
+      <nuxeo-audit-search name="audit" id="audit" provider-name="[[providerName]]"> </nuxeo-audit-search>
+    </nuxeo-card>
```
nmpcunha, line 33: If I understood @Gabez0r [comment](https://github.com/nuxeo/nuxeo-web-ui/pull/910#discussion_r430368606), we still need to wrap this on a `nuxeo-page`. Otherwise we'll break how the page looks.

Gabez0r: Exactly.

elements/nuxeo-document-history/nuxeo-document-history.js
RSalem07: yes, we should use `margin-inline-end` have a look at https://css-tricks.com/css-logical-properties/  `margin-right` -> `margin-inline-end`. But when I added it I didn't see a big difference, it's why I removed it. but Samuel can have more details about it

elements/nuxeo-admin/nuxeo-audit.js
RSalem07: Yes, I saw the Gabriel comment, but I didn't see things related to audit in `nuxeo-page` perhaps I miss something

elements/nuxeo-audit/nuxeo-audit-search.js
```diff
@@ -0,0 +1,258 @@
...
+    return this.$.table;
+  }
+
+  _isAvailable(doc) {
+    return doc && doc.uid;
+  }
+
+  _formati18n(path, key) {
+    return key ? this.i18n(path + key) : '';
+  }
```
nmpcunha, line 186: Please make sure we use this method (from `nuxeo-document-history`) https://github.com/nuxeo/nuxeo-web-ui/blob/master/elements/nuxeo-document-history/nuxeo-document-history.js#L162-L164

elements/nuxeo-audit/nuxeo-audit-search.js
```diff
@@ -0,0 +1,258 @@
...
+
+      principalName: {
+        type: String,
+        value: '',
+      },
+
+      startDate: {
+        type: String,
+        notify: true,
+      },
```
nmpcunha, line 152: line break missing

elements/nuxeo-audit/nuxeo-audit-search.js
```diff
@@ -0,0 +1,258 @@
...
+      endDate: {
+        type: String,
+        notify: true,
+      },
+      events: {
+        type: Array,
+        value: [],
+      },
```
nmpcunha, line 160: Some line breaks missing. I was expecting our format command to perform this kind of changes for us 😞

elements/nuxeo-audit/nuxeo-audit-search.js
```diff
@@ -0,0 +1,258 @@
...
+            <template>[[_formati18n('eventType.', item.eventId)]]</template>
+          </nuxeo-data-table-column>
+          <nuxeo-data-table-column name="[[i18n('audit.date')]]" sort-by="eventDate">
+            <template><nuxeo-date datetime="[[item.eventDate]]"></nuxeo-date></template>
+          </nuxeo-data-table-column>
+          <nuxeo-data-table-column name="[[i18n('audit.username')]]" sort-by="principalName">
+            <template><nuxeo-user-tag user="[[item.principalName]]"></nuxeo-user-tag></template>
+          </nuxeo-data-table-column>
+          <nuxeo-data-table-column name="[[i18n('audit.category')]]" sort-by="category">
+            <template>[[_formati18n('eventCategory.', item.category)]]</template>
```
nmpcunha, line 98: All the `_formati18n` should be replaced when addressing this https://github.com/nuxeo/nuxeo-web-ui/pull/910/commits/4e599f352d818d93e9265fa5ab9ecd09f8208c91#r431175499

elements/nuxeo-audit/nuxeo-audit-search.js
RSalem07: Its what I did initially, I took into account one of you suggestion. But it seems that in some cases (mainly on the audit one) the key can be `null` and we ended with `something.null` it's why I used this version.

and as the audit can have things than the document history I think its better to use mainly the audit version

elements/nuxeo-audit/nuxeo-audit-search.js
RSalem07: I will add them. I don't think the formatter do them, I was facing same pb when a line exceed 120 character, I was expected that the formater do break line .... but not

elements/nuxeo-document-history/nuxeo-document-history.js
semisse: we should avoid as much as possible to hard code the margin position and leave it to the browser to decide where to apply it. 

Ideally it would be the parent's responsibility to manage this kind of spacing, but we don't have a mechanism to do it yet.

So in this case, we should apply the margin-line-end to the first nuxeo-directory-suggestion.

elements/nuxeo-audit/nuxeo-audit-search.js
RSalem07: see my answer https://github.com/nuxeo/nuxeo-web-ui/pull/910#discussion_r431178703. If I recall correctly `Activity.?` can end with `Activity.null` and I reuse the version of the audit one instead of the doc history because the audit one contains the doc history. is why I replace with `_formati18n('eventType.', item.eventId)` which is more global and avoid `something.null`

elements/nuxeo-audit/nuxeo-audit-search.js
RSalem07: to give more details, I tried a second time using the 
`
_formatActivity(key) {
    return key ? this.i18n(`activity.${key}`) : '';
  }
` 
 and some entries are missing for example: 

`activity.search`
`activity.loginSuccess`
`activity.documentRemoved`

these three ones that I saw, but perhaps we have more and even I can add them on `messages.json` we are not sure to cover all entries. We can create a separate ticket to clean and merge them ...

nmpcunha: Fair point!
If we go even deeper, we can see that some events are translated on `nuxeo-document-activity` context where they reference `activity.EVENT` and not on `nuxeo-audit` where they expect `eventType.EVENT`. And vice-versa, as you said!

Agree about the separate ticket to clean this. 👍

RSalem07: To try to keep thing more consistent I ended by adding the missing ones from `eventType.?` and `categoryType.?` from `audit` to `activitity.` I kept the same key and the same value in the order way. That's way we can use the _formatActivity method.

i18n/messages.json
```diff
@@ -2,34 +2,62 @@
...
   "activity.eventWorkflowCategory": "Workflow",
   "activity.lifecycle_transition_event": "updated the document lifecycle state",
+  "activity.loginFailed": "Login failed",
+  "activity.loginSuccess": "Login successful",
+  "activity.logout": "Logout",
+  "activity.NuxeoAuthentication": "Authentication",
+  "activity.relationNotificationCategory": "Waiting publication",
   "activity.removedFromCollection": "removed from collection",
-  "activity.retentionRuleAttached": "Retention rule applied",
+  "activity.retentionRuleAttached": "Retention rule applied",
```
nmpcunha, line 46: Extra space here?

Thanks for your efforts on trying to have consistent labels. Since there are some duplicated and mixed labels (activity and event type, sometimes also mixed with event category), we should create a ticket to clean this up. Specially now that we are sharing the same element 👍

elements/nuxeo-admin/nuxeo-audit.js
```diff
@@ -36,182 +29,22 @@ import { FormatBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-format-behavior.j
...
-              <template>[[_formatDocument(item)]]</template>
-            </nuxeo-data-table-column>
-            <nuxeo-data-table-column name="[[i18n('audit.comment')]]">
-              <template>[[item.comment]]</template>
-            </nuxeo-data-table-column>
-          </nuxeo-data-table>
-        </nuxeo-card>
-      </div>
+      <nuxeo-card>
+        <nuxeo-audit-search name="audit" id="audit" provider-name="[[providerName]]"> </nuxeo-audit-search>
```
nmpcunha, line 37: `provider-name="[[providerName]]` this is the first step for getting this working. But [ELEMENTS-1078](https://jira.nuxeo.com/browse/ELEMENTS-1078) would still be needed.
Anyway, this PR is not depending on it.

i18n/messages.json
RSalem07: You welcome,

As discussed, I am in favour of the new ticket, which make more consistency for the labels, remove the unused ones....

RSalem07:
As discussed the new ticket for cleanup the entries is [NXP-29175](https://jira.nuxeo.com/browse/NXP-29175)

elements/nuxeo-admin/nuxeo-audit.js
```diff
@@ -36,182 +29,22 @@ import { FormatBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-format-behavior.j
...
+      </nuxeo-card>
     </nuxeo-page>
   `,
 
   is: 'nuxeo-audit',
-  behaviors: [FormatBehavior],
```
Gabez0r, line 147: By removing this, the `i18n` method won't be available. At least we need to extend the `I18nBehavior`.

elements/nuxeo-admin/nuxeo-audit.js
```diff
@@ -36,182 +29,22 @@ import { FormatBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-format-behavior.j
...
-            <nuxeo-data-table-column name="[[i18n('audit.document')]]">
-              <template>[[_formatDocument(item)]]</template>
-            </nuxeo-data-table-column>
-            <nuxeo-data-table-column name="[[i18n('audit.comment')]]">
-              <template>[[item.comment]]</template>
-            </nuxeo-data-table-column>
-          </nuxeo-data-table>
-        </nuxeo-card>
-      </div>
+      <nuxeo-card>
```
Gabez0r, line 36: By keeping a card here, we get cards inside cards, which is not what we want:
broken-audit-page

elements/nuxeo-audit/nuxeo-audit-search.js
```diff
@@ -0,0 +1,263 @@
...
+        .row-container :first-child {
+          margin-inline-end: 8px;
+        }
+
+        #table {
+          height: calc(100vh - 370px);
+        }
+      </style>
+
+      <nuxeo-audit-page-provider id="provider" page-size="40" provider-name="[[providerName]]">
```
Gabez0r, line 53: The `providerName`  property does not exist in `nuxeo-audit-page-provider`. We could clean this up.

RSalem07: Agree with you, I just kept it as it was before my refactoring where it was hardcode [here](https://github.com/nuxeo/nuxeo-web-ui/blob/master/elements/nuxeo-admin/nuxeo-audit.js#L66). Perhaps we can keep it and it will be available once the 
[ELEMENTS-1078](https://jira.nuxeo.com/browse/ELEMENTS-1078) and 
https://github.com/nuxeo/nuxeo-web-ui/pull/910#discussion_r428163653.

Gabez0r: I'm ok with it as long as someone can confirm that [ELEMENTS-1078](https://jira.nuxeo.com/browse/ELEMENTS-1078) will get done soon (before 11.1 release). Otherwise, we will be keeping this code that doesn't do anything on a maintenance branch, and I wouldn't expect [ELEMENTS-1078](https://jira.nuxeo.com/browse/ELEMENTS-1078) to be backported to Web UI 3.0.x (matching 11.x).

nmpcunha: TBH, I don't believe it would be done before the release.

elements/nuxeo-admin/nuxeo-audit.js
nmpcunha: Nice catch, thanks!

elements/nuxeo-audit/nuxeo-audit-search.js
RSalem07: Ok, then I will remove it. 
Thx

RSalem07:
@nmpcunha 
Back to our discussion earlier on https://github.com/nuxeo/nuxeo-web-ui/pull/910#discussion_r431698764 and after tackling [NXP-28634](https://jira.nuxeo.com/browse/NXP-28634). It seems that the label used in the original audit are more accurate then the document history ones (prefix by `Activity`: first letter mainly of time are capitalised.... 

and in other hand when we sorting for example using the `eventId` the sort that the backend part do is not really reflect to the labels of the doc history we cas discuss more deeply into one to one.

To summerise and after comparing the result (with my fix on [NXP-28634](https://jira.nuxeo.com/browse/NXP-28634)) I am going to rollback the part where I added the new entries with `Activity.` and I will use the audit ones as I did before taking into account this of your original review (replacing by i18n on activity)

elements/nuxeo-document-history/nuxeo-document-history.js
```diff
@@ -37,185 +29,17 @@ import moment from '@nuxeo/moment';
...
-            value="{{selectedEventCategory}}"
-            placeholder="[[i18n('documentHistory.filter.selectEventCategory')]]"
-            min-chars="0"
-          >
-          </nuxeo-directory-suggestion>
-        </div>
-      </template>
-    </nuxeo-card>
-
     <nuxeo-card>
```
nmpcunha: Assuming that we may want to backport this commit individually, we won't need `nuxeo-card` here, right? (Despite we've cleaned that up in the following commits)

elements/nuxeo-document-history/nuxeo-document-history.js
```diff
@@ -37,185 +29,17 @@ import moment from '@nuxeo/moment';
...
 
   properties: {
-    document: Object,
-    visible: {
-      type: Boolean,
-      value: false,
-      observer: '_refresh',
+    document: {
+      type: Object,
+      value: {},
```
nmpcunha: Not sure if we need to want to have the empty `document` object as default.

elements/nuxeo-web-ui-bundle.html
```diff
@@ -66,7 +66,7 @@
   <template>
     <nuxeo-filter user="[[user]]" expression="user.isAdministrator">
       <template>
-        <nuxeo-audit name="audit" id="audit"></nuxeo-audit>
+        <nuxeo-audit name="audit" id="audit" provider-name="EVENTS_VIEW"></nuxeo-audit>
```
nmpcunha, line 69: Since we have that value by default on `nuxeo-audit`, I would avoid this change.
The behavior will be the same and we avoid one more diff/change

elements/nuxeo-audit/nuxeo-audit-search.js
```diff
@@ -0,0 +1,262 @@
...
+        value: false,
+        observer: '_refresh',
+      },
+
+      document: {
+        type: Object,
+        value: {},
+      },
+
+      providerName: {
```
Gabez0r, line 140: Don't think is is being used anywhere.

elements/nuxeo-document-history/nuxeo-document-history.js
```diff
@@ -37,145 +28,12 @@ import moment from '@nuxeo/moment';
...
-          <template>
-            <a href$="[[_parseComment(item.comment)]]">[[_formatComment(item.comment)]]</a>
-          </template>
-        </nuxeo-data-table-column>
-        <nuxeo-data-table-column name="[[i18n('documentHistory.state')]]">
-          <template><nuxeo-tag uppercase>[[formatLifecycleState(item.docLifeCycle)]]</nuxeo-tag></template>
-        </nuxeo-data-table-column>
-      </nuxeo-data-table>
-    </nuxeo-card>
+    <nuxeo-audit-search name="document-history" id="document-history" document="[[document]]"></nuxeo-audit-search>
```
Gabez0r, line 31: We are keeping `nuxeo-document-history` for the sake of backwards compatibility, right? If so, we should:
- keep the import on the `elements.js` file, otherwise the element won't be available
- mark this element as deprecated

As for deprecation, we could annotate with a "@deprecated since 3.0.0" on the comment block above and add the following to the `ready` callback:
```
console.warn(`${this.is} is deprecated. Please use nuxeo-audit-search instead!`);
```
This is what the [Polymer team does](https://github.com/PolymerElements/paper-drawer-panel/blob/6f9e6998674790abd438a5bc043616199858d161/paper-drawer-panel.js#L468).

WDYT? CC @nmpcunha @nelsonsilva

nmpcunha: > We are keeping nuxeo-document-history for the sake of backwards compatibility, right?
More or less. Not sure if we can just remove the element and make it explicit on the "Upgrade Notes". It may be easier and cleaner, no?

@RSalem07 split the commits in order to be easier to perform a backport, since there are some request to do so. (only the first two commits would be backported)

Anyway, if we think it is better to keep `nuxeo-document-history`, I agree with your suggestion!
The annotation makes sense. Mixed feelings with the warning, but I understand the purpose 🙂 

If we don't want to keep it, on the last commit we can just remove every reference to it.

RSalem07: Yep, the remove of this element was the part of the last commit, but with the rework of this morning I repush it. then let wait for the Nelson opinion

Thx

nelsonsilva: I agree we need to do a much better job at deprecating things. Let's move forward with Gabriel's suggestion as the DX will be much better 👌

elements/nuxeo-audit/nuxeo-audit-search.js
Gabez0r: Sorry, what I meant by my comment was that this property is not currently used by this element for anything (right?). If so, whether an element using this one sets it or not, won't make a different, and therefore we can remove it. We should avoid shipping dead code, and we can always add it later when we implement [ELEMENTS-1078](https://jira.nuxeo.com/browse/ELEMENTS-1078).

RSalem07: Yep, done :) thx

PR #915, opened 2020-05-18
NXP-27987: fix label consistency DAM

andreacornaglia:


nmpcunha:
Since the ticket may be missing some details, this comment may not make sense.
But, we may want to think about consistency in a different way, I guess, because the margin, by itself, will ensure some more spacing, but it is pretty easy to have items not aligned.
Example:
Screenshot 2020-05-18 at 17 41 37

As soon as we have bigger labels we'll end up with something like this. And, by default, i.e. without  changing the original labels, the `Bit Depth per Channel` is not aligned.

We may want input from @semisse to clarify the scope. But probably we may have room to improve the layout a little bit 🙂

themes/base.js
```diff
@@ -126,6 +126,27 @@ const template = html`
...
+          justify-content: space-between;
+          margin-left: -5px;
+        }
+        .damProperties .item:hover {
+          @apply --nuxeo-block-hover;
+        }
+
+        .additional --nuxeo-card .header {
+          margin-left: 5px;
+        }
```
nmpcunha, line 149: Despite understand the purpose of not repeating this CSS part, I guess we would like to avoid bringing "so specific stuff" for the theme configuration.
@Gabez0r @semisse what is the approach you believe should be taken here?

If there is no good idea, I personally believe that the style can be in each element, since that's not so odd on Web Components context.

andreacornaglia: I understand. For now I reverted back the css to each component - ideally we should have 1 template for dam property info that we could use for all these cases, since the code is very similar in all of them - pass the list of properties labels and values, and let the component make the layout.

elements/document/nuxeo-picture-exif.js
```diff
@@ -27,20 +27,25 @@ import { FormatBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-format-behavior.j
 Polymer({
   _template: html`
     <style>
-      .properties label {
+      .damProperties label {
```
nmpcunha, line 30: Are we taking the same direction as https://github.com/nuxeo/nuxeo-elements/pull/241?
i.e. should we revert to the original `properties` name?

andreacornaglia: yes makes sense, I reverted back the name.

elements/document/nuxeo-picture-formats.js
```diff
@@ -29,31 +29,26 @@ Polymer({
       :host {
         display: block;
       }
-
-      label {
-        opacity: 0.5;
-        min-width: 30%;
-        display: inline-block;
+      .properties label {
+        min-width: 10em;
```
semisse, line 33: are we sure we want to use em here?

elements/document/nuxeo-picture-exif.js
```diff
@@ -29,14 +29,19 @@ Polymer({
...
-        text-overflow: ellipsis;
+        margin-inline-end: 12px;
       }
 
-      .item {
+      .properties .item {
         @apply --layout-horizontal;
         @apply --layout-flex;
         line-height: 2.2rem;
+        padding: 0 5px;
```
semisse, line 39: what do you think of starting to apply our grid of multiples of 4 to our calculations?

elements/document/nuxeo-picture-formats.js
andreacornaglia: we were already using it on nuxeo-picture-exif.js so I wanted to keep it consistent, any alternative?

elements/document/nuxeo-picture-exif.js
andreacornaglia: done

elements/document/nuxeo-picture-formats.js
andreacornaglia: changed to rem

elements/document/nuxeo-picture-iptc.js
```diff
@@ -27,20 +27,24 @@ import { FormatBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-format-behavior.j
 Polymer({
   _template: html`
     <style>
+      .text-area {
+        max-width: 60%;
+      }
```
nmpcunha, line 32: nitpick: no need to reorder since it only makes the diff harder to follow. And it seems to be missing a line break 😉

elements/document/nuxeo-picture-document-page.js
```diff
@@ -34,15 +34,18 @@ Polymer({
...
         @apply --layout-flex;
-        padding-right: 1.3rem;
-        padding-bottom: 1.3rem;
+        padding: 16px;
         min-width: 384px;
-        margin: 5px;
+        margin: 4px;
+      }
+
+      .additional --nuxeo-card .header {
```
Gabez0r, line 47: What's this `--nuxeo-card` doing here?

elements/document/nuxeo-picture-exif.js
```diff
@@ -28,15 +28,20 @@ Polymer({
...
       }
 
-      .item {
+      .properties .item {
         @apply --layout-horizontal;
         @apply --layout-flex;
         line-height: 2.2rem;
+        padding: 0 4px;
+        justify-content: space-between;
+        margin-left: -4px;
```
Gabez0r, line 41: Please avoid using negative margins.

elements/document/nuxeo-picture-document-page.js
andreacornaglia: i want to target the header element inside the nuxeo-card, this is the way I found to target it.

elements/document/nuxeo-picture-exif.js
andreacornaglia: I know it's not best practice, but it's the only way I found to make this change possible.

semisse: Margins are used to create space around the elements and what we are trying to accomplish here is just manipulating the position of the hover effect. 

There's a couple of ways (or more) to do this. Consider using a pseudo element to represent the hover state and consider using css translate property. It will for sure create more lines of code but we won't be surprised with side effects that negative margins can bring.

andreacornaglia: using css translate seems more dangerous to me and i don't want to target only hover state - any other options?

andreacornaglia: @Gabez0r @semisse .1. we are already using negative margins in our code. 2. in this case the negative margin neutralizes a positive padding. 3. Any other way will be more hacky and have more side-effects. I honestly don't think we should change this.

elements/document/nuxeo-picture-document-page.js
```diff
@@ -34,15 +34,14 @@ Polymer({
...
       }
 
       nuxeo-card {
         @apply --layout-flex;
-        padding-right: 1.3rem;
-        padding-bottom: 1.3rem;
+        padding: 16px;
         min-width: 384px;
-        margin: 5px;
+        margin: 4px;
```
Gabez0r, line 44: I understand these aim at making a gap of 8px between the cards, and that other alternatives are a title bit shady. So I can live with this. We should consider a cleanup afterwards.

elements/document/nuxeo-picture-exif.js
```diff
@@ -28,15 +28,19 @@ Polymer({
...
+        min-width: 10rem;
+        margin-inline-end: 12px;
       }
 
-      .item {
+      .properties .item {
         @apply --layout-horizontal;
         @apply --layout-flex;
         line-height: 2.2rem;
+        padding: 0 4px;
```
Gabez0r, line 39: These items do not seem aligned with the card title. I guess removing this padding would help with that?

andreacornaglia: it is not aligned because i removed the negative margin :) based on the above comment I'll bring it back

Gabez0r: Do we need the negative margin? If we remove this padding, we should have it aligned, right?

elements/document/nuxeo-picture-exif.js
```diff
@@ -28,15 +28,20 @@ Polymer({
...
+        margin-left: -4px;
```
Gabez0r, line 40: If we add this at the element level, this means that this element will only work with the nuxeo-card (or with elements with a padding >= 4px). Otherwise, this element will overflow the parent. We don't want that. If we must do this hack, we should preferably do it on the parent, which is the only one here knowing what it's paddings and boundaries are. I would expect to reach the same effect by adding the following to the parent (`nuxeo-picture-document-page`) (instead of adding the negative paddings on the children themselves):
```
      nuxeo-card > * {
        margin: -4px;
      }
```

andreacornaglia: this doesn't work because i only want that margin on the item but not on the card header.

Gabez0r: Have you tested it? It aligns the elements in the way that the items are aligned with the cards. You can even just use `margin-left`.

PR #917, opened 2020-05-19
NXP-29129: refresh aws s3 token

nelsonsilva:


addons/amazon-s3-online-storage/index.js
```diff
@@ -95,14 +95,28 @@ class S3Provider {
...
+    const credentials = new AWS.Credentials(
+      options.awsSecretKeyId,
+      options.awsSecretAccessKey,
+      options.awsSessionToken,
+    );
+    credentials.expireTime = new Date(options.expiration);
+    credentials.refresh = (cb) =>
+      this._resource(`upload/${this.batchId}/refreshToken`)
+        .post()
+        .then((response) => {
```
nelsonsilva, line 107: I'll add a `catch(cb)` before merging this

addons/amazon-s3-online-storage/index.js
```diff
@@ -95,14 +95,28 @@ class S3Provider {
   }
 
   _initCredentials(options) {
+    const credentials = new AWS.Credentials(
+      options.awsSecretKeyId,
+      options.awsSecretAccessKey,
+      options.awsSessionToken,
+    );
+    credentials.expireTime = new Date(options.expiration);
+    credentials.refresh = (cb) =>
```
Gabez0r, line 104: When will this `refresh` be called? Do we need to implement `needsRefresh` as well?

nelsonsilva: No, default `needsRefresh` implementaton works with the `expiration` info, see https://github.com/aws/aws-sdk-js/blob/master/lib/credentials.js#L95

Gabez0r: :ok_hand:

PR #919, opened 2020-05-21
NXP-29139: allow removing missing blobs

nelsonsilva:


elements/document/nuxeo-document-import.js
```diff
@@ -1188,16 +1188,35 @@ Polymer({
     return this.$.docRequest.post();
   },
 
+  _removeFile(index) {
+    this.splice('localFiles', index, 1);
+    this.hasLocalFiles = this.localFiles && this.localFiles.length > 0;
+    // clear error message if no more files with error
+    if (!this.localFiles.some((f) => f.error)) {
+      this.set('errorMessage', undefined);
+    }
```
Gabez0r, line 1197: This is not safe, because the error message might not be related to the upload. It can be an "invalid location", in which case we should leave it be. I'd say that we need to reconsider how we displaying upload errors and location errors (in the scope of [ELEMENTS-1123](https://jira.nuxeo.com/browse/ELEMENTS-1123) for example).

PR #927, opened 2020-05-26
NXP-27416: add nuxeo-spreadsheet app

nelsonsilva:
Moving Spreadsheet code into Web UI.

Basically the plan here is:
- Align on oldest possible npm dependency to replace even olders bower deps
- Stick with the iframe in a dialog approach since we still rely on nuxeo-select2
- Use a webpack config to define a new entrypoint (see https://github.com/nuxeo/nuxeo-web-ui/pull/927/files#diff-db455a2559f213458984c55d2f0cbf24)


webpack.config.js
```diff
@@ -201,4 +208,6 @@ const production = merge([
   },
 ]);
 
-module.exports = (mode) => merge(common, mode === 'production' ? production : development, { mode });
+const spreadsheet = require('./addons/nuxeo-spreadsheet/webpack.config');
+
+module.exports = (mode) => merge(common, spreadsheet, mode === 'production' ? production : development, { mode });
```
Gabez0r, line 213: We should only be doing this (including the spreadsheet config) if `process.env.NUXEO_PACKAGES` includes "nuxeo-spreadsheet", right?

nelsonsilva: We have to do the build regardless of `NUXEO_PACKAGES` since we need to be able to active addons at runtime.

Gabez0r: Well noted.

index.html
```diff
@@ -115,6 +115,7 @@
             dateFormat: 'LL',
             dateTimeFormat: 'LLL',
             firstDayOfWeek: '',
+            // spreadsheet: { enabled: 'true'}
```
mnixo, line 118: Is this actually needed?

nelsonsilva: no it's not but since the index.html is used only for local development I thought it could be handy to have it there so you can easilly enable the spreadsheet locally 🤷‍♂️

addons/nuxeo-spreadsheet/app/lib/select2-editor.js
```diff
@@ -0,0 +1,168 @@
...
+  this.textareaParentStyle.left = 0;
+  this.textareaParentStyle.display = 'none';
+
+  this.TEXTAREA_PARENT.appendChild(this.TEXTAREA);
+
+  this.instance.rootElement.appendChild(this.TEXTAREA_PARENT);
+
+  const that = this;
+  Handsontable.hooks.add('afterRender', () => {
+    // TODO(nfgs) - was that.instance.registerTimeout
```
richardsd, line 42: What do you mean to do here?

addons/nuxeo-spreadsheet/app/lib/select2-editor.js
```diff
@@ -0,0 +1,168 @@
...
+
+  const keyCodes = Handsontable.helper.keyCode;
+  const ctrlDown = (event.ctrlKey || event.metaKey) && !event.altKey; // catch CTRL but not right ALT (which in some systems triggers ALT+CTRL)
+
+  // Process only events that have been fired in the editor
+  if (event.target !== that.TEXTAREA || event.isImmediatePropagationStopped()) {
+    return;
+  }
+
+  if (event.keyCode === 17 || event.keyCode === 224 || event.keyCode === 91 || event.keyCode === 93) {
```
richardsd, line 73: No readable keycodes from this right? https://github.com/handsontable/handsontable/blob/ca1ab950de4522bdb899f9e442cef978d860685e/handsontable.d.ts#L2121

index.js
```diff
@@ -58,16 +58,20 @@ const tmpl = document.createElement('template');
 tmpl.innerHTML = html;
 document.head.appendChild(tmpl.content);
 
+const isEnabled = (flag) => Nuxeo.UI.config[flag] && String(Nuxeo.UI.config[flag].enabled) === 'true';
+
+const bundles = [...Nuxeo.UI.bundles, ...(isEnabled('spreadsheet') ? ['nuxeo-spreadsheet'] : [])];
```
richardsd, line 63: Why are you adding the `nuxeo-spreadsheet` explicitly here?

addons/nuxeo-spreadsheet/ftest/features/step_definitions/spreadsheet.js
```diff
@@ -0,0 +1,68 @@
+import { Then, When } from 'cucumber';
+import Spreadsheet from '../../pages/spreadsheet';
+
+When('I open the spreadsheet', function() {
+  const button = this.ui.results.actions.element('nuxeo-spreadsheet-button').click();
+
+  const dialog = button.element('#dialog');
+  dialog.waitForVisible();
```
Gabez0r, line 8: nit: Since we don't use the dialog, we could just do:
```suggestion
  button.waitForVisible('#dialog');
```

addons/nuxeo-spreadsheet/ftest/features/step_definitions/spreadsheet.js
```diff
@@ -0,0 +1,68 @@
...
+
+  const iframe = button.element('#iframe');
+  driver.frame(iframe.value);
+
+  this.spreadsheet = new Spreadsheet();
+});
+
+When('I see the spreadsheet dialog', function() {
+  const button = this.ui.browser.results.actions.element('nuxeo-spreadsheet-button');
+  button.element('#dialog').waitForVisible();
```
Gabez0r, line 18: Same here:
```suggestion
  button.waitForVisible('#dialog');
```

addons/nuxeo-spreadsheet/ftest/features/step_definitions/spreadsheet.js
```diff
@@ -0,0 +1,68 @@
...
+
+Then('I can see the spreadsheet results actions button', function() {
+  const { results } = this.ui;
+
+  if (results.displayMode !== 'table') {
+    results.toggleTableView.click();
+  }
+
+  const button = results.actions.element('nuxeo-spreadsheet-button');
+  button.waitForVisible();
```
Gabez0r, line 29: Same here:
```suggestion
  results.actions.waitForVisible('nuxeo-spreadsheet-button');
```

addons/nuxeo-spreadsheet/app/lib/select2-editor.js
nelsonsilva: idk 😂

index.js
nelsonsilva: because unlike other addons this one is not part of `NUXEO_PACKAGES`so it'll be loaded based on a new `Nuxeo.UI.config.spreadsheet.enabled` flag

PR #929, opened 2020-05-27
NXP-28338: show translated taskName instead of action in workflow analytics

mnixo:
Note: the `_labels` method is only used for this chart in particular, so we won't be translating labels elsewhere.
Will post a pipeline status ASAP.


mnixo:
🔵 [Pipeline run.](https://qa.nuxeo.org/jenkins/job/web-ui-pipeline-mnixo/1/)

PR #934, opened 2020-06-04
NXP-29213: retry failed cleanup calls after ftest

nelsonsilva:


packages/nuxeo-web-ui-ftest/features/step_definitions/support/fixtures/documents.js
```diff
@@ -128,16 +128,26 @@ Before(function() {
...
-      nuxeo
-        .repository()
-        .delete(docUid)
-        .catch(() => {}),
-    ),
-  ) // eslint-disable-line arrow-body-style
-    .then(() => {
-      liveDocuments = [];
-    }),
+  Promise.all(liveDocuments.map((docUid) => retry(() => nuxeo.repository().delete(docUid)))).then(() => {
```
Gabez0r, line 150: We should leave a comment stating why we need to do a retry here. It is a little bit shady.

packages/nuxeo-web-ui-ftest/features/step_definitions/support/fixtures/documents.js
```diff
@@ -128,16 +128,26 @@ Before(function() {
...
+            console.error(error);
+            if (retries === 0) {
+              reject(error);
+              return;
+            }
+            retry(fn, interval, --retries).then(resolve, reject);
+          }),
+        interval,
+      ),
+  );
```
Gabez0r, line 147: nit: It would be useful to extract this into a separate file, a folder level above, that we could import. It would be useful on [group.js, where we already to 1 retry](https://github.com/nuxeo/nuxeo-web-ui/blob/master/packages/nuxeo-web-ui-ftest/features/step_definitions/support/fixtures/groups.js#L24-L30).

PR #918, opened 2020-05-20
Fix nxp 28980 show upload errors

Gabez0r:
Depends on:
- nuxeo/nuxeo-elements#240
- nuxeo/nuxeo-elements#242
- nuxeo/nuxeo-web-ui#920
- nuxeo/nuxeo-web-ui#919
- nuxeo/nuxeo-web-ui#917
- nuxeo/nuxeo-web-ui#916

PR #935, opened 2020-06-04
NXP-29129: fix Web UI upload error handling

Gabez0r:
Aggregates:
#916
#917
#918
#919
#920

### DIFF
--- a/addons/amazon-s3-online-storage/index.js
+++ b/addons/amazon-s3-online-storage/index.js
@@ -118,6 +118,10 @@ class S3Provider {
           },
           endpoint: this.extraInfo.endpoint || null,
           computeChecksums: true,
+          httpOptions: {
+            timeout:
+              Number(Nuxeo && Nuxeo.UI && Nuxeo.UI.config && Nuxeo.UI.config.s3 && Nuxeo.UI.config.s3.timeout) || 0,
+          },
         });
       });
   }
--- a/addons/amazon-s3-online-storage/index.js
+++ b/addons/amazon-s3-online-storage/index.js
@@ -91,6 +91,12 @@ class S3Provider {
                   callback({ type: 'uploadCompleted', fileIdx: file.index });
                 }
                 resolve();
+              })
+              .catch((err) => {
+                if (typeof callback === 'function') {
+                  callback({ type: 'uploadInterrupted', file, error: err.message || err });
+                }
+                reject(err);
               });
           } else {
             callback({ type: 'uploadInterrupted', file, error });
--- a/addons/amazon-s3-online-storage/index.js
+++ b/addons/amazon-s3-online-storage/index.js
@@ -95,14 +95,28 @@ class S3Provider {
   }
 
   _initCredentials(options) {
+    const credentials = new AWS.Credentials(
+      options.awsSecretKeyId,
+      options.awsSecretAccessKey,
+      options.awsSessionToken,
+    );
+    credentials.expireTime = new Date(options.expiration);
+    credentials.refresh = (cb) =>
+      this._resource(`upload/${this.batchId}/refreshToken`)
+        .post()
+        .then((response) => {
+          credentials.accessKeyId = response.awsSecretKeyId;
+          credentials.secretAccessKey = response.awsSecretAccessKey;
+          credentials.sessionToken = response.awsSessionToken;
+          credentials.expireTime = new Date(response.expiration);
+          cb();
+        });
     AWS.config.update({
-      credentials: new AWS.Credentials(options.awsSecretKeyId, options.awsSecretAccessKey, options.awsSessionToken),
+      credentials,
       region: options.region,
       s3ForcePathStyle: options.usePathStyleAccess || false,
       useAccelerateEndpoint: options.useS3Accelerate || false,
     });
-
-    AWS.config.credentials.expireTime = new Date(options.expiration);
   }
 
   _newBatch() {
--- a/addons/amazon-s3-online-storage/index.js
+++ b/addons/amazon-s3-online-storage/index.js
@@ -232,5 +232,11 @@ if (
   Nuxeo.UI.config.s3 &&
   String(Nuxeo.UI.config.s3.useDirectUpload) === 'true'
 ) {
+  // https://github.com/aws/aws-sdk-js/issues/1895
+  AWS.util.update(AWS.S3.prototype, {
+    reqRegionForNetworkingError(req, done) {
+      return done();
+    },
+  });
   UploaderBehavior.defaultProvider = 's3';
 }
--- a/addons/amazon-s3-online-storage/index.js
+++ b/addons/amazon-s3-online-storage/index.js
@@ -154,32 +154,37 @@ class S3Provider {
       });
   }
 
+  _handleUploadError(error, callback) {
+    if (this.files.every((file) => file.error)) {
+      callback({ type: 'batchFailed', error, batchId: this.batchId });
+    } else if (this.files.every((file) => file.error || file.complete)) {
+      callback({ type: 'batchFinished', batchId: this.batchId });
+    }
+    throw error;
+  }
+
   get accepts() {
     return UploaderBehavior.getProviders().default.prototype.accepts;
   }
 
   upload(files, callback) {
-    this._ensureBatch()
-      .then(() => {
-        callback({ type: 'batchStart', batchId: this.batchId });
-        if (new Date().getTime() >= this.extraInfo.expiration) {
-          this._refreshBatchInfo();
-        }
+    this._ensureBatch().then(() => {
+      callback({ type: 'batchStart', batchId: this.batchId });
+      if (new Date().getTime() >= this.extraInfo.expiration) {
+        this._refreshBatchInfo();
+      }
 
-        const promises = [];
-        for (let i = 0; i < files.length; ++i) {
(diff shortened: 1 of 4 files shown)


addons/amazon-s3-online-storage/index.js
```diff
@@ -118,35 +146,45 @@ class S3Provider {
...
 
+  _handUploadError(error, callback) {
+    if (this.files.every((file) => file.error || file.complete)) {
+      callback({ type: 'batchFinished', batchId: this.batchId });
+    } else if (this.files.every((file) => file.error)) {
+      callback({ type: 'batchFailed', error, batchId: this.batchId });
+    }
+    throw error;
+  }
+
```
RSalem07, line 165: Perhaps I am missing something, but regarding to [here](https://github.com/nuxeo/nuxeo-web-ui/blob/c8dfc4cedcca1c0b4ea16a7a5bdcc817f768b79c/addons/amazon-s3-online-storage/index.js#L178) and the call of `file, callback` each file will be uploaded independently. If i's the case should we pass the processed `file`  [here](https://github.com/nuxeo/nuxeo-web-ui/blob/c8dfc4cedcca1c0b4ea16a7a5bdcc817f768b79c/addons/amazon-s3-online-storage/index.js#L182) into this method `_handUploadError` and because the method is expect to handle only error regarding to its name.

Gabez0r: I don't think we need to pass it. The `_handUploadError` method only runs the `callback` with the events `batchFinished` and `batchFailed`, which are non file specific (note that we do not refer to any particular file in the `callback`). The file specific error handling is done in the [_upload](https://github.com/nuxeo/nuxeo-web-ui/blob/c8dfc4cedcca1c0b4ea16a7a5bdcc817f768b79c/addons/amazon-s3-online-storage/index.js#L96) method directly, which actually receives the file as a parameter.

FYI: the `_handUploadError` method was added because we were not consistently firing either `batchFailed` or `batchFinished` when a batch finished with some failed uploads. I think this concept feels a little bit weird in the context of S3, but these events are part of our public API for upload handlers, so...

elements/document/nuxeo-document-import.js
```diff
@@ -757,8 +828,17 @@ Polymer({
...
-            this.hasLocalFilesUploaded = this.hasLocalFiles && this.localFiles.every((file) => !file || file.complete);
+          if (match[2] === 'complete' || match[2] === 'error') {
+            this.hasLocalFilesUploaded =
+              this.hasLocalFiles && this.localFiles.every((file) => !file || file.complete || file.error);
+            // if we are already editing file properties, we need to uncheck those whose upload failed
+            if (this.hasLocalFilesUploaded) {
+              this._getAllFiles().forEach((f, index) => {
+                if (f.error) {
+                  this._setFileProp(index, 'checked', false);
+                }
```
RSalem07, line 839: I am wondering, if we should add the check on:
```
this.hasLocalFilesUploaded =
              this.hasLocalFiles && this.localFiles.every((file) => !file || file.complete || file.error);
```
as the end we re-check 
```
   this._getAllFiles().forEach((f, index) => {
                if (f.error) {
                  this._setFileProp(index, 'checked', false);
                }
```
I mean we control if there is any `f.error`  

and for further perhaps we can replace to 
```
  this._getAllFiles().filter((f) => f.error ).forEach(....)
                
```

addons/amazon-s3-online-storage/index.js
```diff
@@ -118,35 +146,45 @@ class S3Provider {
...
+              Number(Nuxeo && Nuxeo.UI && Nuxeo.UI.config && Nuxeo.UI.config.s3 && Nuxeo.UI.config.s3.timeout) || 0,
+          },
         });
       });
   }
 
+  _handUploadError(error, callback) {
+    if (this.files.every((file) => file.error || file.complete)) {
+      callback({ type: 'batchFinished', batchId: this.batchId });
+    } else if (this.files.every((file) => file.error)) {
```
andreacornaglia, line 160: shouldn't this.files.every((file) => file.error) come before the other one? else, you could still send batchFinished if all files are errors...

elements/document/nuxeo-document-import.js
```diff
@@ -284,6 +311,28 @@ Polymer({
...
+        margin: 1em 2em;
+        padding-left: 8px;
+      }
+
+      .upload-error:not(:empty) {
+        border-left: 4px solid var(--nuxeo-warn-text);
+      }
+
+      .upload-error:empty::before {
+        content: '\\200b';
```
andreacornaglia, line 325: what does \\200b mean?

elements/document/nuxeo-document-import.js
```diff
@@ -1222,6 +1330,37 @@ Polymer({
...
 
+  _computedCheckItem(e) {
+    if (e.base && e.base.checked) {
+      return 'icons:check-circle';
+    }
+    return 'icons:radio-button-unchecked';
+  },
+
+  _computeRemoveIcon(file) {
+    if (file.base) {
```
andreacornaglia, line 1341: we could avoid nesting ifs by checking !file.base first (same on below method)

elements/document/nuxeo-document-import.js
```diff
@@ -174,13 +196,22 @@ Polymer({
...
       }
 
-      .file-overview iron-icon {
-        margin-top: 0.1em;
+      .file-overview.selected .name {
+        color: var(--nuxeo-primary-color);
+      }
+
+      .file-overview .name:hover {
+        color: var(--nuxeo-primary-color);
```
semisse, line 204: nit: we should add a default value

elements/document/nuxeo-document-import.js
```diff
@@ -174,13 +196,22 @@ Polymer({
         @apply --nuxeo-block-selected;
       }
 
-      .file-overview iron-icon {
-        margin-top: 0.1em;
+      .file-overview.selected .name {
+        color: var(--nuxeo-primary-color);
```
semisse, line 200: nit: we should add a default value

elements/document/nuxeo-document-import.js
Gabez0r: It's a zero-width space. Them main goal there it's to just take up vertical space (as if there was text there).

addons/amazon-s3-online-storage/index.js
Gabez0r: You're right. Thanks!

elements/document/nuxeo-document-import.js
Gabez0r: I left the nested ifs because without it, it would read as follows:
```
    if (!file.base || (file.base && !this.hasAbort() && !(file.base.complete || file.base.error))) {
      return '';
    }
    if (file.base.complete || file.base.error) {
      return 'nuxeo:remove';
    }
    return 'icons:cancel';
```
Which I think is harder to read. WDYT?

elements/document/nuxeo-document-import.js
Gabez0r: Not sure I understood the first suggestion.

Regarding applying the filter, we can't do it because it would change the meaning of `index`, which is used by `_setFileProp`.

RSalem07: >Not sure I understood the first suggestion.

I am wondering if the 
```
this.hasLocalFilesUploaded =this.hasLocalFiles && this.localFiles.every((file) => !file || file.complete || file.error);
```
is a requirement for looping over the files ` this._getAllFiles()` I think it's like we have a double check on `f.error` and I am wondering if we can remove the first one ` this.hasLocalFiles && this.localFiles.every((file) => !file || file.complete || file.error);` and just keep the second `if (f.error) {  this._setFilePro...`
 
>Regarding applying the filter, we can't do it because it would change the meaning of index, which is used by _setFileProp.

I see, you are right about loosing then index 👍

Gabez0r: Ah, sorry. `hasLocalFilesUploaded` is needed for the UI to update! It's independent from the looping on `this._getAllFiles()`.

elements/document/nuxeo-document-import.js
```diff
@@ -164,6 +185,7 @@ Polymer({
         font-weight: bold;
         text-transform: none;
         color: var(--secondary-text-color);
+        height: 86px;
```
semisse, line 188: nit: we should align with the multiples of 4px grid defined by our Design Guidelines. Is it possible to change to 84 or 88px?

elements/document/nuxeo-document-import.js
andreacornaglia: gocha, yes in that case is better to go for legibility

addons/amazon-s3-online-storage/index.js
```diff
@@ -118,35 +152,45 @@ class S3Provider {
...
           computeChecksums: true,
+          httpOptions: {
+            timeout:
+              Number(Nuxeo && Nuxeo.UI && Nuxeo.UI.config && Nuxeo.UI.config.s3 && Nuxeo.UI.config.s3.timeout) || 0,
+          },
         });
       });
   }
 
+  _handUploadError(error, callback) {
```
semisse, line 163:
```suggestion
  _handleUploadError(error, callback) {
```

Gabez0r:
Landed some minor CSS fixes.

PR #936, opened 2020-06-05
NXP-29215: align filtering between audit and document history

RSalem07:
PR related to ELEMENT one https://github.com/nuxeo/nuxeo-elements/pull/250
Also this also a backport part of https://github.com/nuxeo/nuxeo-web-ui/pull/910 not the whole things only the first one https://github.com/nuxeo/nuxeo-web-ui/commit/d912d21ee8807c7b025ab71e2795d13c495095af

T&P In Progress: https://qa2.nuxeo.org/jenkins/job/WebUI/job/web-ui-pipeline-10.10/451/


elements/nuxeo-document-history/nuxeo-document-history.html
```diff
@@ -100,56 +131,70 @@
...
+        }
+        if (this.selectedEventCategory) {
+          params.eventCategory = this.selectedEventCategory;
+        }
+
+        if (this._hasValidDate(this.startDate) && this._hasValidDate(this.endDate)) {
+          var start = Date.parse(this.startDate);
+          var end = Date.parse(this.endDate);
+          if (start > end) {
+            this.startDate = moment(end)
```
Gabez0r, line 177: I think this will cause the `_refresh` (and thus `_buildParams`) to be triggered multiple times. Perhaps we should have a separate observer to handle the logic of fixing an invalid combination of start and end date?

Gabez0r: Ideally, we'd make use of the `min` and `max` properties of the date-picker for this effect. The end date would be the max of the start date, and start date the min of the end date.

RSalem07: > I think this will cause the _refresh (and thus _buildParams) to be triggered multiple times. Perhaps we should have a separate observer to handle the logic of fixing an invalid combination of start and end date?

Right, I add a new logic to avoid it, I will create a new ticket for master. for information the without the new logic (master version) we will have one more call.

RSalem07: >Ideally, we'd make use of the min and max properties of the date-picker for this effect. The end date would be the max of the start date, and start date the min of the end date.

also Done, Thanks ;)

elements/nuxeo-document-history/nuxeo-document-history.html
```diff
@@ -35,31 +37,62 @@
...
-        height: calc(100vh - 370px);
+      .row-container * {
+        flex: 1 0 0;
+        margin-right: 8px;
       }
 
-      nuxeo-date-picker {
-        padding: 0 16px;
+      #table {
+        height: calc(100vh - 370px);
```
Gabez0r, line 51: We do get a double scrollbar (if we get more content on the table), which I don't think we want. Can you reproduce? Could you please try with:
```suggestion
        height: calc(100vh - 420px);
```

RSalem07: Thanks, I increase the part of fixed height. Just a a quick question How did you check this part I mean you retrieved my branch and test or do you have another way to check it

Gabez0r: I made a checkout, because I was curious about the min/max of the date picker (I don't think I have ever used it :thinking:), a it works quite well! :P

RSalem07: Thanks ;), yeah  for `min` and `max` it works really good :)

elements/nuxeo-document-history/nuxeo-document-history.html
```diff
@@ -100,56 +133,81 @@
...
+      },
+
+      _getFormattedDate: function(dateAsString) {
+        return this._hasValidDate(dateAsString) ? moment(Date.parse(dateAsString)).format('YYYY-MM-DD') : '';
+      } ,
+
+      _observeDates: function() {
+        var start = this._hasValidDate(this.startDate) ? Date.parse(this.startDate) : null;
+        var end = this._hasValidDate(this.endDate) ? Date.parse(this.endDate) : null;
+        var refresh = !start || !end || start < end
```
semisse, line 173: missing a semicolon
```suggestion
        var refresh = !start || !end || start < end;
```

elements/nuxeo-document-history/nuxeo-document-history.html
```diff
@@ -100,56 +133,81 @@
...
+        return dateAsString && dateAsString.length > 0;
+      },
+
+      _getFormattedDate: function(dateAsString) {
+        return this._hasValidDate(dateAsString) ? moment(Date.parse(dateAsString)).format('YYYY-MM-DD') : '';
+      } ,
+
+      _observeDates: function() {
+        var start = this._hasValidDate(this.startDate) ? Date.parse(this.startDate) : null;
+        var end = this._hasValidDate(this.endDate) ? Date.parse(this.endDate) : null;
```
semisse, line 172:
```suggestion
        var start = this._hasValidDate(this.startDate) && Date.parse(this.startDate);
        var end = this._hasValidDate(this.endDate) && Date.parse(this.endDate);
```

elements/nuxeo-document-history/nuxeo-document-history.html
```diff
@@ -100,56 +133,79 @@
...
         },
+        principalName: {
+          type: String,
+          value: '',
+        },
         startDate: {
           type: String,
           notify: true,
-          observer: '_observeStartDate'
+          observer: '_observeDates'
```
semisse, line 143: just noticed this now, but we are missing commas in the end of these last lines

PR #941, opened 2020-06-12
Fix nxp 29251 filter on event audit admin

RSalem07:


elements/nuxeo-audit/nuxeo-audit-search.js
```diff
@@ -14,6 +14,15 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
 */
+
+import '@nuxeo/nuxeo-elements/nuxeo-audit-page-provider.js';
+import '@nuxeo/nuxeo-ui-elements/nuxeo-data-table/iron-data-table.js';
+import '@nuxeo/nuxeo-ui-elements/widgets/nuxeo-date.js';
+import '@nuxeo/nuxeo-ui-elements/widgets/nuxeo-date-picker.js';
+import '@nuxeo/nuxeo-ui-elements/widgets/nuxeo-directory-suggestion.js';
+import '@nuxeo/nuxeo-ui-elements/widgets/nuxeo-user-suggestion.js';
+import '@nuxeo/nuxeo-ui-elements/widgets/nuxeo-user-tag.js';
+import '@nuxeo/nuxeo-ui-elements/widgets/nuxeo-card.js';
```
andreacornaglia, line 25: @RSalem07 why are we adding these now? wasn't it working before?

RSalem07: It worked fine, but when I tried to backport (only a part not the whole), it worked correctly on my local machine but  Jenkins Q2 fails [here](https://qa2.nuxeo.org/jenkins/job/WebUI/job/web-ui-pipeline-10.10/447/console) its why I make the choice to change it. also I posted a message on our Slack to get more details from UI team

elements/nuxeo-audit/nuxeo-audit-search.js
```diff
@@ -225,6 +226,25 @@ class AuditSearch extends mixinBehaviors([FormatBehavior, RoutingBehavior], Nuxe
     return dateAsString && dateAsString.length > 0;
   }
 
+  _getFormattedDate(dateAsString) {
```
andreacornaglia, line 229: should we move this method to Nuxeo Format? Is this something we may need to reuse somewhere else?

RSalem07: good point, thanks, I tried to use the same thing that I improved on 10.10 [here](https://github.com/nuxeo/nuxeo-web-ui/pull/936/files#diff-36519008abdfd671a9bc29a5e4cdaac5R66) and [here](https://github.com/nuxeo/nuxeo-web-ui/pull/936/files#diff-36519008abdfd671a9bc29a5e4cdaac5R69) but it not work correctly. I will try to ping Gabriel next week. at meanwhile I will remove it

elements/nuxeo-audit/nuxeo-audit-search.js
Gabez0r: We should always import the elements we use. I think Polymer lint is not doing it's job properly on master.

PR #946, opened 2020-06-16
NXP-29267: align with Elements release candidates

Gabez0r:


package.json
```diff
@@ -11,7 +11,7 @@
   },
   "devDependencies": {
     "@nuxeo/nuxeo-web-ui-ftest": "file:./packages/nuxeo-web-ui-ftest",
-    "@nuxeo/testing-helpers": "3.0.0-SNAPSHOT",
+    "@nuxeo/testing-helpers": "~3.0.0-rc || 3.0.0-SNAPSHOT",
```
Gabez0r, line 14: Using `SNAPSHOT` would also work, because we're tagging all snapshot releases with that version. But it might be dangerous, since we might have snapshots from different branches?

PR #949, opened 2020-06-18
NXP-29268: release when merging on reference branches

Gabez0r:


Jenkinsfile
```diff
@@ -58,21 +92,25 @@ pipeline {
...
+    VERSION = getVersion()
   }
   stages {
-    stage('Prepare') {
+    stage('Update package version') {
       steps {
         container('mavennodejs') {
+          echo """
+          ------------------------=---------
+          Update package versions ${VERSION}
```
nelsonsilva, line 104: nit: i guess this should be either "update packages version" or "update package version"

Jenkinsfile
```diff
@@ -203,44 +241,59 @@ pipeline {
         }
       }
     }
-    stage('Publish Maven artifacts') {
+    stage('Git commit, tag and push') {
       when {
-       branch 'master'
+        allOf {
+          not {
```
nelsonsilva, line 247: shouldn't this run only in master and maintenance branches at most?

Gabez0r: That was the goal here, by I'm assuming that the only branches where the pipeline is run is controlled by jx configuration.

Jenkinsfile
```diff
@@ -284,21 +344,24 @@ pipeline {
...
-            dir('server') {
-              skaffoldBuild()
-            }
+          echo """
+          ------------------------------
+          Build and deploy Docker images
+          ------------------------------
+          Image tag: ${VERSION}
+          """
+          dir('server') {
```
nelsonsilva, line 361: we dont' want to do this anymore for now, right ?

Gabez0r: Without it, skaffold build fails because of https://github.com/nuxeo/nuxeo-web-ui/blob/85b78959adee330cd87b2f8c032eed9c223be6eb/skaffold.yaml#L30

PR #963, opened 2020-07-07
NXP-29365: refresh collections view when deleting collection doctype

tmartins:
And ensure consistency on untrash event payload


tmartins:
T&P was successful at https://qa2.nuxeo.org/jenkins/job/WebUI/job/web-ui-pipeline-9.10/338/

PR #964, opened 2020-07-08
NXP-29210: rely on the current document id if available

mnixo:
The goal is to navigate to the document referenced by the action response. Additionally, we no longer rely on the `document-updated` event to close the dialog, so we now have two new (and more specific) events: `save` and `cancel`.

### DIFF
--- a/elements/nuxeo-app.js
+++ b/elements/nuxeo-app.js
@@ -630,7 +630,8 @@ Polymer({
     if (this.page === 'search') {
       this._refreshSearch();
     } else if ((this.docPath && this.docPath.length > 0) || (this.docId && this.docId.length > 0)) {
-      this.load('browse', this.docId, this.docPath, this.docAction);
+      const id = this.docId || (this.currentDocument && this.currentDocument.uid);
+      this.load('browse', id, this.docPath, this.docAction);
     } else {
       this.navigateTo('home');
     }


elements/document/nuxeo-document-form-layout.js
```diff
@@ -163,14 +165,10 @@ Polymer({
   },
 
   cancel() {
-    this._refresh();
+    this.fire('cancel');
     this.document = undefined;
   },
 
-  _refresh() {
-    this.fire('document-updated');
```
Gabez0r, line 171: IIRC, we still need to fire this on cancel, otherwise the document won't update the fetch the latest changes.

elements/document/nuxeo-document-form-layout.js
```diff
@@ -148,8 +149,9 @@ Polymer({
       action = this.$.doc.put();
     }
     action
-      .then(() => {
-        this._refresh(this);
+      .then((response) => {
+        this.fire('save');
```
Gabez0r, line 153: If we are picky, this could be considered an API change, since anyone relying on this to know that the document was saved will be in the dark. I don't think this is used by anyone other then `nuxeo-document-form-button`. As a matter of fact, this event was just used to communicate with `nuxeo-app`, and its meaning is ambiguous when fired from `nuxeo-document-form-layout`, since we cannot tell save from cancel. So I think this improvement makes perfect sense.

Nonethess, if this is considered a risk, then we'd have to go for a more conservative option: fire the `document-updated` event with the reason on the detail, and then we'd have to prevent it from bubbling up on `nuxeo-document-form-button` if the reason was "save". I'd consider this very hacky.

elements/document/nuxeo-document-form-layout.js
nelsonsilva: yes, we should fire the event to force a fetch of the update document at top level, including all the enrichers

nelsonsilva: iirc the main problem here is that we’re refreshing the doc based on its path which can change. If we refreshed based on the uid or on event detail sent here (updated doc) we should be good

Gabez0r: I'd give this a shot. We already have the id, so perhaps we could tweak the [refresh method on nuxeo-app](https://github.com/nuxeo/nuxeo-web-ui/blob/master/elements/nuxeo-app.js#L633), to not actually pass the docPath to the load method, for example. :crossed_fingers:

mnixo:
This is a much simpler approach. Because `this.docId` is `""`, the load is based on the provided `this.docPath`. However, we can use `this.currentDocument.uid`, which will take precedence over the provided document path.

elements/nuxeo-app.js
```diff
@@ -630,7 +630,7 @@ Polymer({
     if (this.page === 'search') {
       this._refreshSearch();
     } else if ((this.docPath && this.docPath.length > 0) || (this.docId && this.docId.length > 0)) {
-      this.load('browse', this.docId, this.docPath, this.docAction);
+      this.load('browse', this.docId || this.currentDocument.uid, this.docPath, this.docAction);
```
nelsonsilva, line 633: :+1: assuming `this.currentDocument` is always defined when we get here.

PR #965, opened 2020-07-08
NXP-29273: add workflow for rc promotion

nelsonsilva:


.github/workflows/promote.yaml
```diff
@@ -0,0 +1,102 @@
...
+          npm i @nuxeo/nuxeo-dataviz-elements@$NEW_ELEMENTS_VERSION --save-prefix="~"
+          npm i @nuxeo/nuxeo-ui-elements@$NEW_ELEMENTS_VERSION --save-prefix="~"
+          npm i @nuxeo/testing-helpers@$NEW_ELEMENTS_VERSION --save-prefix="~"
+
+          find . -type f -not -path "./node_modules/*" -regex ".*\.\(yaml\|sample\|xml\)" -exec sed -i 's/'"$OLD_SNAPSHOT"'/'"$NEW_SNAPSHOT"'/g' {} \;
+
+          # bump server version
+          OLD_SERVER_SNAPSHOT=11.1-SNAPSHOT # this can be read from the POM
+          NEW_SERVER_SNAPSHOT=11.2-SNAPSHOT # this is INPUT
+          sed -i 's/'"$OLD_SERVER_SNAPSHOT"'/'"$NEW_SERVER_SNAPSHOT"'/g' pom.xml
```
Gabez0r, line 98: I don't think we want to keep these.

.github/workflows/promote.yaml
```diff
@@ -0,0 +1,102 @@
...
+
+          # commit and push
+          git commit -a -m "Update $OLD_SNAPSHOT to $NEW_SNAPSHOT"
+          git push origin master # (or push to a branch and open PR if doing this manually)
```
Gabez0r, line 102: Should we also push the docker images and the maven artifacs?

nelsonsilva: Not publishing anything apart from the MP for now

PR #967, opened 2020-07-14
NXP-29187: position and style document import buttons

mnixo:


mnixo:
🔵  https://qa.nuxeo.org/jenkins/job/web-ui-pipeline-mnixo/42/

elements/document/nuxeo-document-import.js
```diff
@@ -1500,4 +1498,16 @@ Polymer({
     return form._form.checkValidity(); */
     return true;
   },
+
+  _hideEditPrevious() {
```
Gabez0r, line 1502: nit: since this will be used to disable and not to hide, I think `_disableEditPrevious` would make this method more self explanatory. Same for the two methods below.

mnixo: Absolutely!

PR #968, opened 2020-07-14
NXP-28238: refactor nuxeo-breadcrumb

semisse:


elements/nuxeo-browser/nuxeo-breadcrumb.js
```diff
@@ -14,192 +14,239 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-      if (index === this._enrichers.breadcrumb.entries.length - 3) {
-        return 'breadcrumb-item-grand-parent';
+    _resize() {
+      this._setBreadcrumbElements();
+      const ancestors = this.shadowRoot.querySelector('.ancestors');
+      const deletedNodes = [];
+      const ellipsis = document.createElement('span');
+      ellipsis.classList.add('ellipsis');
+      ellipsis.textContent = '... >';
+      ancestors.insertBefore(ellipsis, ancestors.firstChild);
```
Gabez0r, line 216: If we are adding the ellipsis before we measure, then we'll display them even if we have just the right amount of space available to display the whole breadcrumb.

elements/nuxeo-browser/nuxeo-breadcrumb.js
```diff
@@ -14,192 +14,239 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+      while (this.contentWidth + ellipsis.offsetWidth > ancestors.offsetWidth) {
+        deletedNodes.push(ancestors[1]);
+        ancestors.childNodes[1].remove();
+      }
+      while (ancestors.scrollWidth < ancestors.clientWidth && deletedNodes.length > 0) {
```
Gabez0r, line 221: When is this run? I added a breakpoint inside this loop and it was never hit.

I think this happens because we are always reconstructing the breadcrumb, so we are always starting from a position where we just have to remove stuff.

elements/nuxeo-browser/nuxeo-breadcrumb.js
```diff
@@ -14,192 +14,239 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+      this._setBreadcrumbElements();
+      const ancestors = this.shadowRoot.querySelector('.ancestors');
+      const deletedNodes = [];
+      const ellipsis = document.createElement('span');
+      ellipsis.classList.add('ellipsis');
+      ellipsis.textContent = '... >';
+      ancestors.insertBefore(ellipsis, ancestors.firstChild);
+      while (this.contentWidth + ellipsis.offsetWidth > ancestors.offsetWidth) {
+        deletedNodes.push(ancestors[1]);
+        ancestors.childNodes[1].remove();
```
Gabez0r, line 219: We should make sure we handle edge cases here. We should always keep one child element visible, and it can even have a very large title (might require more ellipsis? :scream:).

Can be tested by having a parent folder with a large title. 
Screenshot from 2020-07-15 11-09-14

elements/nuxeo-browser/nuxeo-breadcrumb.js
```diff
@@ -14,192 +14,239 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+      ancestors.innerHTML = '';
+      this._breadcrumb = this._getBreadcrumbElements();
+      this._breadcrumb.forEach((element) => {
+        const listItem = document.createElement('li');
+        const span = document.createElement('span');
+        span.classList.add('breadcrumb-item');
+
+        const spanTitle = document.createElement('span');
+        spanTitle.classList.add('breadcrumb-item-title');
+        spanTitle.textContent = element.title;
```
Gabez0r, line 192: We could add a title attribute as well. It can be useful for long tittles, right? (see last comment)

elements/nuxeo-browser/nuxeo-breadcrumb.js
```diff
@@ -14,192 +14,257 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+      this._breadcrumb.forEach((element) => {
+        const listItem = document.createElement('li');
+
+        const anchor = document.createElement('a');
+        anchor.textContent = element.title;
+        anchor.setAttribute('title', element.title);
+        anchor.setAttribute('href', this.urlFor('browse', element.path));
+
+        listItem.appendChild(anchor);
+        this.ancestors.appendChild(listItem);
```
Gabez0r, line 202: If `listItem` is large enough, we display the ellispis at the end, but they are truncated. This happens because the `li` elements are overflowing the `#ancestors`. Can't we do something about this overflow?

elements/nuxeo-browser/nuxeo-breadcrumb.js
```diff
@@ -14,192 +14,257 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+    _getBreadcrumbElements() {
+      if (this._enrichers) {
+        return this._enrichers.breadcrumb.entries.slice(0, this._enrichers.breadcrumb.entries.length - 1);
       }
-      if (index === this._enrichers.breadcrumb.entries.length - 2) {
-        return 'breadcrumb-item-parent';
+    }
+
+    _setBreadcrumbElements() {
+      this.deletedNodes = [];
```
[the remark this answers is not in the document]
semisse: I actually think we need to initialize it here to make sure it exists and it is empty, even before the resize.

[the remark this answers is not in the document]
Gabez0r: But we only use this on `_resize`, right? On resize, we are already checking if it is empty and if so, we do this very same logic.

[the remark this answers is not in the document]
semisse: we should initialize it here and remove it in the `_resize` method

elements/nuxeo-browser/nuxeo-breadcrumb.js
semisse: We want the overflow on the `#ancestors` in order to do the trick. This problema affects a very specific edge case, where we don't have space to introduce the ellipsis generated by the truncation via css text-overflow property. The alternative would be to include yet another method via javascript that would detect the size of the breadcrumb item; detect the size of the generated ellipsis and remove character by character until the ellipsis don't cause text-overflow.

elements/nuxeo-browser/nuxeo-breadcrumb.js
```diff
@@ -14,192 +14,252 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+      return this.document && this.document.contextParameters;
+    }
+
+    get contentWidth() {
+      return Array.from(this.ancestors.children).reduce((sum, current) => sum + current.offsetWidth, 0);
+    }
+
+    _getBreadcrumbElements() {
+      if (this._enrichers) {
+        return this._enrichers.breadcrumb.entries.slice(0, this._enrichers.breadcrumb.entries.length - 1);
```
Gabez0r, line 185: nit: could be a one-liner:
```suggestion
      return this._enrichers && this._enrichers.breadcrumb.entries.slice(0, this._enrichers.breadcrumb.entries.length - 1);
```

elements/nuxeo-browser/nuxeo-breadcrumb.js
```diff
@@ -14,192 +14,252 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+            letter-spacing: 0.02rem;
+            text-decoration: none;
+          }
 
-      a:hover {
-        color: var(--nuxeo-link-hover-color);
-        opacity: 1;
-      }
+          #ancestors a:hover {
+            color: var(--nuxeo-link-hover-color, #0066ff);
```
mnixo, line 109: By setting this fallback, does it mean that we might expect `--nuxeo-link-hover-color` to **not** be defined?

elements/nuxeo-browser/nuxeo-breadcrumb.js
```diff
@@ -14,192 +14,252 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-      if (index === this._enrichers.breadcrumb.entries.length - 3) {
-        return 'breadcrumb-item-grand-parent';
+    disconnectedCallback() {
+      super.disconnectedCallback();
+      this.removeEventListener('iron-resize', this._resize);
+    }
+
+    get ancestors() {
+      const ancestors = this.shadowRoot.getElementById('ancestors');
+      return ancestors;
```
mnixo, line 172:
```suggestion
      return this.shadowRoot.getElementById('ancestors');
```
Maybe?

elements/nuxeo-browser/nuxeo-breadcrumb.js
semisse: Yes, this is a best practice that we should implement every time we use css variables, since it might not be defined

elements/nuxeo-browser/nuxeo-breadcrumb.js
```diff
@@ -14,192 +14,252 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-      return document.type === 'Root' ? this.i18n('browse.root') : document.title;
+    get contentWidth() {
+      return Array.from(this.ancestors.children).reduce((sum, current) => sum + current.offsetWidth, 0);
     }
-  },
 
-  _icon(document, url) {
-    if (document && document.properties && document.properties['common:icon']) {
-      return url ? url + document.properties['common:icon'] : '';
+    _getBreadcrumbElements() {
```
Gabez0r, line 183: nit: we don't actually get much from having since as a separate method, since this is only called once and it's fairly short.

test/nuxeo-breadcrumb.test.js
```diff
@@ -0,0 +1,207 @@
...
+const router = {
+  browse: (path) => path.substring(1),
+};
+
+suite('nuxeo-breadcrumb', () => {
+  let server;
+
+  setup(async () => {
+    server = fakeServer.create();
+    await login();
```
Gabez0r, line 86: Login already returns a new fake server. I'd expect this to work as well:```
```suggestion
    server = await login();
```

test/nuxeo-breadcrumb.test.js
```diff
@@ -0,0 +1,207 @@
...
+  teardown(() => {
+    server.restore();
+  });
+
+  test('Should display a breadcrumb when a document is set', async () => {
+    const breadcrumb = await fixture(
+      html`
+        <nuxeo-breadcrumb .document=${document} .router=${router}></nuxeo-breadcrumb>
+      `,
+    );
```
Gabez0r, line 98: DRY: we have repeated this 8 times on this file. We could wrap in in a helper function.

test/nuxeo-breadcrumb.test.js
```diff
@@ -0,0 +1,207 @@
...
+      );
+
+      const ancestors = breadcrumb.shadowRoot.querySelector('#ancestors');
+      const separators = [];
+      for (let index = 1; index < ancestors.childNodes.length; index++) {
+        const separator = window.getComputedStyle(ancestors.childNodes[index], ':before').getPropertyValue('content');
+        separators.push(separator);
+        expect(separator).to.equal('" > "');
+      }
+      expect(separators.length).to.equal(5);
```
Gabez0r, line 157: If we are only measure the size of separators, I guess we could just use a counter instead?

test/nuxeo-breadcrumb.test.js
nmpcunha: Indeed, that method is deprecated and you [should use MockClient](https://github.com/nuxeo/nuxeo-elements/blob/master/testing-helpers/test-helpers.js#L144-L147) instead.

elements/nuxeo-browser/nuxeo-breadcrumb.js
```diff
@@ -14,192 +14,247 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-      }
+    static get properties() {
+      return {
+        document: {
+          type: Object,
+          observer: '_setBreadcrumbElements',
+        },
+        _breadcrumb: {
+          type: Object,
+          value: {},
```
Gabez0r, line 156: This is an array, right?
```suggestion
          type: Array,
```

Gabez0r:
Also, please move this to a WEBUI ticket.

test/nuxeo-breadcrumb.test.js
```diff
@@ -0,0 +1,168 @@
...
+    test('Should set links in ancestors when document has breadcrumb entries', async () => {
+      const ancestors = breadcrumb.shadowRoot.querySelector('#ancestors');
+      ancestors.childNodes.forEach((ancestor, index) => {
+        const breadcrumbEntries = document.contextParameters.breadcrumb.entries[index].path;
+        expect(ancestor.firstChild.getAttribute('href')).to.equal(breadcrumbEntries);
+      });
+    });
+
+    test('Should add ellipsis to ancestors when parent is smaller than nuxeo-breadcrumb', async () => {
+      const element = await fixture(
```
richardsd, line 138: Since this is shared by two test scenarios maybe you can move it to the suite setup as well, wdty?

richardsd (approved):

Gabez0r (dismissed):

test/nuxeo-breadcrumb.test.js
Gabez0r: But then we'd be initializing it on the other three tests of the suit that do not require it, right? We could also extract this to a function to prevent duplication.

Gabez0r (approved):

richardsd (approved):
The new structure makes sense for me 👍 sorry for all the noise 🙂

semisse:
Not at, @richardsd, it's much better right now. Thank you all for the reviews

Gabez0r (changes requested):
Please do not push the package-lock.json file.

Gabez0r (approved):

mnixo (approved):

PR #974, opened 2020-07-21
NXP-29268: move the maven-internal repo outside the QA profile

mnixo:


pom.xml
```diff
@@ -40,19 +40,6 @@
...
-          <releases>
-            <enabled>true</enabled>
-          </releases>
-          <snapshots>
-            <enabled>true</enabled>
-            <updatePolicy>always</updatePolicy>
-          </snapshots>
-        </repository>
-      </repositories>
       <pluginRepositories>
```
Gabez0r, line 43: I think we can move this to after `<repositories>...</repositories>` and just remove the QA profile.

PR #975, opened 2020-07-21
NXP-29439: add package lock for nuxeo-web-ui-ftest

Gabez0r:


Gabez0r:
Here, we are just locking the dependencies of the functional test framework at the Web UI level. Other Web UI dependencies are not locked.

PR #981, opened 2020-07-24
NXP-28972: handle 403 error after task completion

Gabez0r:


elements/nuxeo-app.js
```diff
@@ -716,9 +716,11 @@ Polymer({
         this.docAction = action;
         this.loading = false;
         this.show(page);
+        return doc;
       })
       .catch((err) => {
         this.showError(err.status, this.i18n('browse.error'), err.message);
+        return err;
```
Gabez0r, line 723: If we just re-throw the error, we'd get a bunch of ugly errors on the console, for stuff that has been handled at the UI level only so far.

PR #984, opened 2020-07-29
NXP-29388: avoid showing toast for errors regarding aborted requests

mnixo:


elements/search/nuxeo-search-form.js
```diff
@@ -908,6 +908,9 @@ Polymer({
   },
 
   _onError(e) {
+    if (e.detail.error && e.detail.error.name === 'AbortError') {
```
richardsd, line 911: By kind of swallowing the abort errors will we lose traceability for the errors? 🤔

mnixo: It will be swallowed in the context of the UI, but it will still be logged in the browser console.
Screenshot 2020-08-04 at 11 03 23

PR #985, opened 2020-07-30
WEBUI-14: allow navigation by uid

Gabez0r:


Gabez0r:
The first three commits (629eef73cc4508f1e579a33c7f7b54ae4c13de41, 4c0625b9cabf303bf315eb660098e0871e462927 and 1978ed6d59c78c7361dc472a4e2bee3443beb24c) correspond to the first alternative, which is without resorting to a route resolver.

The second approach uses the route resolver (96f2cf6c72c0ac67ea6b4f4a3d0b6ba74ca88127, d0c2ddec243758fd4100e06bd4ba41529e2acdc8 and https://github.com/nuxeo/nuxeo-web-ui/pull/985/commits/256fd2d8434be5e35bd6e5bb3dd7716187b4ccc9).

If we thing that the second approach is too risky, we can drop the commits. If not, we should squash:
629eef73cc4508f1e579a33c7f7b54ae4c13de41 with 96f2cf6c72c0ac67ea6b4f4a3d0b6ba74ca88127
4c0625b9cabf303bf315eb660098e0871e462927 with d0c2ddec243758fd4100e06bd4ba41529e2acdc8 and
1978ed6d59c78c7361dc472a4e2bee3443beb24c with 256fd2d8434be5e35bd6e5bb3dd7716187b4ccc9

In either case, we must drop baee05c6a788b9d9fc926d5093044f4c6e4d6a53, as it is only required for testing.

elements/routing.js
```diff
@@ -127,7 +128,44 @@ app.router = {
...
+    if (args.length === 0) {
+      return;
+    }
+    const [obj] = args;
+    if (typeof obj !== 'object') {
+      throw new Error(`cannot resolve route: "${obj}" is not a valid entity object`);
+    }
+    let entityType = obj['entity-type'];
+    if (!entityType) {
+      // XXX Somethimes we don't have the entity type. For example, on nuxeo-document-storage, we were not storing it.
```
mnixo, line 141:
```suggestion
      // XXX Sometimes we don't have the entity type. For example, on nuxeo-document-storage, we were not storing it.
```

mnixo (dismissed):
Looks good to me, there's just one remark:
- `urlFor` was consistently changed to be used like `urlFor(item)` instead of `urlFor('browse', item.path)` (for example).
- `navigateTo` on the other hand, seems to be changed from `navigateTo('document', doc.uid)` to either `navigateTo('document', doc)` or just `navigateTo(doc)`, which seems a bit weird. Is this expected?

Gabez0r:
@mnixo, thank you for the review. I kept the `navigateTo('document', doc)` in the cases we are navigating to a version, because navigating by uid is the only reliable way to do it. If it was replaced with `navigateTo(doc)` it would break with the default config which is to navigate by path.

index.html
```diff
@@ -93,6 +93,14 @@
...
               document: ['properties'],
             },
+            /* pageprovider: {
+              method: 'post',
+            },
+            router: {
+              key: {
+                document: 'uid',
+              },
+            }, */
```
mnixo, line 103: Is this important to keep? If so, could we leave a note to clarify why it's here?

mnixo (dismissed):

index.html
Gabez0r: It's useful to enable navigation by UID and search by POST (instead of get), but I guess we could remove it. 🤷

mnixo (approved):

addons/nuxeo-csv/elements/nuxeo-document-import-csv.js
```diff
@@ -684,7 +684,7 @@ Polymer({
   _close() {
     if (this.stage === 'progress' && this._count > 0) {
       this.fire('document-updated');
-      this.navigateTo('browse', this.targetPath);
+      this.navigateTo('document', this.targetPath);
```
nelsonsilva, line 687: we keep explicit navigation by path here or should we switch to using the `parent` document property of `nuxeo-path-suggestion` in order to use tge full document here ?

elements/nuxeo-restore-version-button/nuxeo-restore-version-button.js
```diff
@@ -99,8 +99,18 @@ Polymer({
...
-        this.navigateTo('browse', this.document.path);
+        if (
+          Nuxeo &&
+          Nuxeo.UI &&
+          Nuxeo.UI.config &&
+          Nuxeo.UI.config.router &&
+          Nuxeo.UI.config.router.key &&
+          Nuxeo.UI.config.router.key.document === 'uid'
+        ) {
+          this.navigateTo('document', this.document.versionableId);
```
nelsonsilva, line 110: Is this an unrelated bug fix ? couldn't this logic be in the default navigation logics ?

perhaps we could do `this.navigateTo('document', (Nuxeo && ...) || this.document.path)`

elements/routing.js
```diff
@@ -138,8 +139,19 @@ app.router = {
...
   },
 
-  document(id) {
-    return `/doc/${id}`;
+  document(idOrPath, subPage) {
+    const isId = idOrPath && !idOrPath.startsWith('/');
+    if (isId) {
+      return `/doc/${idOrPath}${subPage ? `?p=${encodeURIComponent(subPage)}` : ''}`;
+    }
+    return `/browse${
```
nelsonsilva, line 147: can we just call `browse` here ?

nelsonsilva: we might as well just use `/doc` everytime and change the page.js handler to accept both paths and ids and just fwd them to `app.load()` no ?

elements/nuxeo-restore-version-button/nuxeo-restore-version-button.js
```diff
@@ -99,8 +99,18 @@ Polymer({
       this.$.opRestoreVersion.input = this.document.uid;
       this.$.opRestoreVersion.params = { checkout: true };
       this.$.opRestoreVersion.execute().then(() => {
-        this.fire('document-updated');
-        this.navigateTo('browse', this.document.path);
+        if (
```
richardsd, line 102: So many conditions 😱

index.html
```diff
@@ -119,6 +119,8 @@
             dateFormat: 'LL',
             dateTimeFormat: 'LLL',
             firstDayOfWeek: '',
+            // pageprovider: { method: 'post', }, // uncomment to enable search using POST (see NXP-29126)
```
richardsd, line 122: This is just used locally, right?

richardsd (dismissed):
It looks good to me 👍 , but please take in consideration my limited knowledge on routing 😬

index.html
Gabez0r: Yes! Just on dev mode.

addons/nuxeo-csv/elements/nuxeo-document-import-csv.js
Gabez0r: This might not be safe, because the suggester parent for `/default-domain/workspaces/WS1` is `/default-domain/workspaces/` whereas for `/default-domain/workspaces/WS1/` it is `/default-domain/workspaces/WS1`. But we can use `this.parent` which is updated by the `nuxeo-document-create-popup`.

elements/nuxeo-restore-version-button/nuxeo-restore-version-button.js
Gabez0r: This is totally related. How do you return to the original document if you are on a version document? Browsing to it's path does the trick, but what if we want to to navigate by UID? Don't know where else to put this logic, since this behavior is specific to this action.

Thank you for the suggestion. I will use a single `this.navigateTo`.

elements/nuxeo-restore-version-button/nuxeo-restore-version-button.js
Gabez0r: We need a helper for this...

elements/routing.js
Gabez0r: > can we just call browse here ?

Yes, thank you. :+1: 

> we might as well just use /doc everytime and change the page.js handler to accept both paths and ids and just fwd them to app.load() no ?

Not sure if I follow. Can you provide a couple of examples?

mnixo (dismissed):

richardsd (approved):

mnixo (approved):

nelsonsilva (approved):

richardsd (approved):

PR #992, opened 2020-08-04
WEBUI-15: add support for overriding routing

Gabez0r:


elements/routing.js
```diff
@@ -181,4 +180,18 @@ customElements.whenDefined('nuxeo-app').then(() => {
 
     navigate: page,
   };
+};
+
+customElements.whenDefined('nuxeo-app').then(() => {
+  const app = document.querySelector('nuxeo-app');
+  if (Nuxeo && Nuxeo.UI && Nuxeo.UI.routing && Nuxeo.UI.routing.htmlImportOverride) {
```
Gabez0r, line 187: `Nuxeo.UI.routing.htmlImportOverride` could perfectly be named differently and be a boolean instead of a string, in which case `routing.html` would be hardcoded in this file as well.

Gabez0r: Also, didn't put it inside a `Nuxeo.UI.config`, because it is controlled by us and not read from a config file.

index.js
```diff
@@ -76,4 +78,23 @@ Promise.all(
...
+  .then(() => customElements.whenDefined('nuxeo-app'))
+  .then(() => {
+    Nuxeo.UI.app = document.querySelector('nuxeo-app');
+    if (!Nuxeo.UI.app) {
+      console.error('could not find nuxeo-app');
+    }
+  })
+  .then(async () => {
+    if (Nuxeo.UI.config.router && Nuxeo.UI.config.router.htmlImport) {
+      window.page = page;
```
nelsonsilva, line 91: could've done this regardless of the use of the html import, just like we do for Polymer compat above

index.js
```diff
@@ -76,4 +78,23 @@ Promise.all(
...
+      importHref(
+        Nuxeo.UI.app.resolveUrl('routing.html'),
+        () => console.warn(`default routing overridden by "routing.html"`),
+        () => import(/* webpackMode: "eager" */ './elements/routing.js'),
```
nelsonsilva, line 95: I'd drop the fallback and log error in this case. Also the warning level on the override is too much imho as we expect this to be done on purpose and in production so we don't want to see warnings ootb

nelsonsilva (approved):

index.js
Gabez0r: Do we want to keep exposing page, specially if we want to introduce our own wait to configure routing?

nelsonsilva: we don't but we do already, so just like everything else, if we were to do it it would be only for compat reasons since people might be relying on page.js today

richardsd (approved):

nelsonsilva (approved):

richardsd (approved):

PR #994, opened 2020-08-07
WEBUI-28: fix icon misalignment in create dialog

richardsd:
Still to be discussed. The best solution should be to define the `--nuxeo-button-icon-margin-end` [instead ](https://github.com/nuxeo/nuxeo-web-ui/blob/0cab8027de899ec0bc1344136d40801daa5fed5f/themes/base.js#L560)

### DIFF
--- a/themes/base.js
+++ b/themes/base.js
@@ -557,7 +557,6 @@ const template = html`
         }
 
         --nx-button-icon: {
-          margin-inline-end: var(--nuxeo-button-icon-margin-end, 4px);
           width: 16px;
           height: 16px;
         }


elements/document/nuxeo-document-create.js
```diff
@@ -50,7 +50,10 @@ Polymer({
         --paper-dialog-scrollable: {
           padding: 0;
           overflow-x: hidden;
-        }
+        };
+        --nx-button-icon: {
+          margin-inline-end: 0;
+        };
```
Gabez0r, line 56: Couldn't we just replace this with
```suggestion
        --nuxeo-button-icon-margin-end: 0;
```
?

Gabez0r: @semisse do you know if we really need to have `margin-inline-end: var(--nuxeo-button-icon-margin-end, 4px);` on every button by default on base.js?

richardsd: @Gabez0r I tried that, but it doesn't work. About the use of `margin-inline-end` for every button on default, we were discussing that with @andreacornaglia as well to understand the use cases.

richardsd: The purpose of having the `margin-inline-end: var(--nuxeo-button-icon-margin-end, 4px);` for every button by default is to support horizontal buttons with an icon and text. It seems to me that the base class should not define any margin at all, and we should add another class to be used for those buttons. wdyt?

Gabez0r: I agree. I a button with an icon does not have a label, or have the label on the bottom, this doesn't make sense anymore.

richardsd: Same for me. [This should be removed](https://github.com/nuxeo/nuxeo-web-ui/blob/master/themes/base.js#L560) and handled for specific use cases if necessary.

themes/base.js
```diff
@@ -557,7 +557,6 @@ const template = html`
         }
 
         --nx-button-icon: {
-          margin-inline-end: var(--nuxeo-button-icon-margin-end, 4px);
```
semisse, line 560: if we remove this line we will lose the spacing, at least, in nuxeo-user-management. Maybe we could create a separate ticket to handle those?

elements/document/nuxeo-document-create.js
Gabez0r: As far as I understoond, only four elements seem to need this:

- nuxeo-user-group-management
- nuxeo-group-management
- nuxeo-user-management
- nuxeo-user-profile

Gabez0r: I'd be ok width just removing the `margin-inline-end: var(--nuxeo-button-icon-margin-end, 4px);` from base.js, since this doesn't seem to be a common case...

Gabez0r: Should we add a new class for the users and group management elements?

richardsd: Yes, I agree with adding a new class for both the users and group management elements. I'll create a separate ticket for handling this and assign it to me, so I can tackle both at the same time.

themes/base.js
richardsd: Yes, I agree. I'll create a separate ticket to handle those elements and assign it to me, so I can tackle it at the same time.

richardsd: I opened ELEMENTS-1222 to handle it 👍

semisse:
Please, @richardsd, for consistency, please remove the final dot on the commit message 🙏

richardsd:
> Please, @richardsd, for consistency, please remove the final dot on the commit message 🙏

Sure, I'll remove it. I thought the pattern was to use the dot.

semisse (approved):

mnixo (approved):

PR #996, opened 2020-08-10
WEBUI-26: ai suggestion fix

andreinechaev:


addons/nuxeo-ai-core/index.js
```diff
@@ -95,11 +95,12 @@ const AISuggestionManager = (() => {
       const isModelInput = path && aiModels && aiModels.inputs.includes(path.replace('document.properties.', ''));
       const noPropertyChanged = !path || 'document.properties'.startsWith(path);
 
-      const model = layout._getBoundElements('document.properties');
+      const model = layout.document.properties;
```
Gabez0r, line 98: Not sure I follow this change. `_getBoundElements` returns a map where the keys represent all the document fields present in the layout and the values represent the elements bound to them. With this change, model will just be an object with the element properties. Won't this break the suggestions?

Also, `_getBoundElements` can only be undefined if layout is not actually an instance of `nuxeo-layout`.

andreinechaev: frankly, idk. I was checking what's coming as the part of the object. and it had a document associated.

Gabez0r: Were you still getting suggestions after this change?

andreinechaev: ok, I changed the code. It seems that the transfer from 10.10 wasn't done right and some methods were missing

andreinechaev: but I still have doubts about ```createNestedObject(this.document.properties, this.property.split('/'));``` because I don't think that `this.document.properties` is an object. In 10.10 version we had our own implementation. Trying to test it.

addons/nuxeo-ai-core/index.js
```diff
@@ -45,6 +45,37 @@ const AISuggestionManager = (() => {
...
 
+  function _getBoundElements(element, property) {
+    const model = {};
+    for (let i = 0; i < element.__templateInfo.nodeInfoList.length; i++) {
+      const nodeInfo = element.__templateInfo.nodeInfoList[i];
+      const node = element.__templateInfo.nodeList[i];
+      // Checking if the node is introspectable
+      if (node.nodeType === Node.ELEMENT_NODE) {
+        // In case we're having other node use cases
+        if (typeof node.hasAttribute === 'undefined') {
```
Gabez0r, line 56: When what sort of situation does this happen? We are already making sure the node is of type ELEMENT, so according to the API, [it is supposed to have an hasAttribute](https://developer.mozilla.org/en-US/docs/Web/API/Element) method.

addons/nuxeo-ai-core/elements/nuxeo-ai-suggestions.js
```diff
@@ -149,11 +148,34 @@ class AISuggestions extends mixinBehaviors([IronFormElementBehavior], Nuxeo.Elem
...
+   */
+  _createNestedObjectRecursive(obj, path) {
+    if (path.length === 0) {
+      return;
+    }
+    if ((!Object.prototype.hasOwnProperty.call(obj, path[0]) && !obj[path[0]]) || typeof obj[path[0]] !== 'object') {
+      obj[path[0]] = {};
+    }
+    return this._createNestedObjectRecursive(obj[path[0]], path.slice(1));
+  }
```
Gabez0r, line 172: What's the difference between this method a the [createNestedObject](https://github.com/nuxeo/nuxeo-elements/blob/master/core/utils.js#L49-L55) it replaces?

Gabez0r:
Also, we are now using the WEBUI project instead of the NXP (official communication coming soon). Could you please convert your NXP to a WEBUI ticket and amend the commit message? Thank you.

addons/nuxeo-ai-core/elements/nuxeo-ai-suggestions.js
andreinechaev: it doesn't throw an exception which breaks the recursion

nmpcunha: Hello!
That's because what you are passing is not an object, right?
So, what is it in fact? Would you have access to the other Object API methods? 🤔

nmpcunha: Sorry for the all the questions, but I was trying to understand what kind of scenario do you have so we can help you properly 🙂 
Because it really seems to be an edge case that we didn't consider at the time and it may worth to think further about it.

addons/nuxeo-ai-core/index.js
andreinechaev: frankly, don't know. That's a copy of what we had in 10.10. I can remove this condition if you think it's excessive

Gabez0r: I just trying to understand if this change makes any difference, and if so, in what conditions. Because if it doesn't, then I think the method does the same as `layout._getBoundElements('document.properties');`, which was replaced in line 129. (see below :arrow_down_small:)

addons/nuxeo-ai-core/index.js
```diff
@@ -95,11 +126,12 @@ const AISuggestionManager = (() => {
       const isModelInput = path && aiModels && aiModels.inputs.includes(path.replace('document.properties.', ''));
       const noPropertyChanged = !path || 'document.properties'.startsWith(path);
 
-      const model = layout._getBoundElements('document.properties');
+      const model = _getBoundElements(layout, 'document.properties');
```
Gabez0r, line 129: How is `_getBoundElements` different from the `layout._getBoundElements` method?

andreinechaev: the layout in this case doesn't have this method. That's exactly why the ticket was started and that's the fix.

Gabez0r: Can you inspect what `layout` is in these cases? This means that `layout` might actually not be a layout...

addons/nuxeo-ai-core/elements/nuxeo-ai-suggestions.js
andreinechaev: it's an object - initially. As you can see it's a recursive method. so, at some point it gets to the node which is not an Object and throws an exception, that shouldn't be the way a recursion is used. it should fail gracefully.

nmpcunha: Probably it was not the purpose of the method, if we look to its JSDoc. 🤷🏼‍♂️ 

Anyway, are you able to provide us the input that is making it failing? Not "one input" that fails since that's easy, but your specific instance/object?
Thanks!

andreinechaev: give in PM

addons/nuxeo-ai-core/elements/nuxeo-ai-suggestions.js
```diff
@@ -149,11 +148,34 @@ class AISuggestions extends mixinBehaviors([IronFormElementBehavior], Nuxeo.Elem
...
+
   _selectSuggestion(event) {
     const suggestion = this.suggestions[event.model.index];
     const propertyPath = this._parsePropertyPath();
     if (!this.get(propertyPath)) {
-      createNestedObject(this.document.properties, this.property.split('/'));
+      this._createNestedObjectRecursive(this.document.properties, this.property.split('/'));
```
Gabez0r, line 178: I think we can keep using `createNestedObject` but on the line above we should do instead `this.get(propertyPath) === undefined`, since we seem to be trying to initialize a property that is already initialized. In the end this would do a bunch of function calls for nothing. Do you mind testing this?

Gabez0r (approved):

nmpcunha (approved):

PR #998, opened 2020-08-12
WEBUI-21: Update Web UI elements to use WriteProperties permission

mnixo:


packages/nuxeo-web-ui-ftest/features/step_definitions/document.js
```diff
@@ -311,6 +322,48 @@ Then('I can see {int} validation error(s) in the {string} edit form', function(n
...
+  page.view.el.element('nuxeo-dropzone').waitForVisible();
+});
+
+Then("I can't see the option to add a main blob", function() {
+  const page = this.ui.browser.documentPage(this.doc.type);
+  page.waitForVisible();
+  page.view.waitForVisible();
+  page.view.el.element('nuxeo-dropzone').waitForVisible(browser.options.waitforTimeout, true).should.be.true;
+});
+
```
mnixo, line 366: This chunk right here can probably be improved. Things to consider:
- Is this the right place for step definitions this specific?
- Using `el.element()` here is probably wrong, where should I create proper helpers?
- Is there a way to generalize this logic so we don't repeat it so much?

elements/nuxeo-document-attachments/nuxeo-document-attachments.js
```diff
@@ -69,7 +69,7 @@ Polymer({
         </template>
       </div>
 
-      <template is="dom-if" if="[[_hasWritePermission(document)]]">
+      <template is="dom-if" if="[[_isDropzoneAvailable(document)]]">
```
richardsd, line 72: Why renaming the helper? Due to poor naming in the first place?

ftest/features/edit-note.feature
```diff
@@ -6,9 +6,35 @@ Feature: Edit Note
...
+
+  Examples:
+  | format   |
+#  | HTML     |
+  | XML      |
+  | Markdown |
+  | Text     |
 
   Scenario Outline: Edit <format> Note
+    And I have permission ReadWrite for this document
```
richardsd, line 37: It is kind of weird to start the scenario outline with `And`. I think it improves readability to have something like:

```
Given I have permission ReadWrite for this document
And I have a <format> Note
```

packages/nuxeo-web-ui-ftest/features/step_definitions/document.js
```diff
@@ -163,6 +163,10 @@ Then(/I can see (.+) metadata with the following properties:/, function(docType,
   });
 });
 
+Then(/^I can't edit the document metadata$/, function() {
+  this.ui.browser.editButton.waitForVisible(browser.options.waitforTimeout, true).should.be.true;
```
richardsd, line 167: Do you always have to specify the timeout if it is the default?

packages/nuxeo-web-ui-ftest/features/step_definitions/document.js
```diff
@@ -311,6 +322,48 @@ Then('I can see {int} validation error(s) in the {string} edit form', function(n
...
+  page.waitForVisible();
+  page.view.waitForVisible();
+  page.view.el.element('nuxeo-replace-blob-button').waitForVisible(browser.options.waitforTimeout, true).should.be.true;
+});
+
+Then('I can see the option to add new attachments', function() {
+  const page = this.ui.browser.documentPage(this.doc.type);
+  page.waitForVisible();
+  page.metadata.waitForVisible();
+  page.metadata.el.element('nuxeo-dropzone').waitForVisible();
```
richardsd, line 343: Missing at least one assert here, right?

`page.metadata.el.element('nuxeo-dropzone').waitForVisible().should.be.true;`

packages/nuxeo-web-ui-ftest/features/step_definitions/document.js
```diff
@@ -311,6 +322,48 @@ Then('I can see {int} validation error(s) in the {string} edit form', function(n
...
+  page.waitForVisible();
+  page.metadata.waitForVisible();
+  page.metadata.el.element('nuxeo-dropzone').waitForVisible(browser.options.waitforTimeout, true).should.be.true;
+});
+
+Then('I can see the option to add a main blob', function() {
+  const page = this.ui.browser.documentPage(this.doc.type);
+  page.waitForVisible();
+  page.view.waitForVisible();
+  page.view.el.element('nuxeo-dropzone').waitForVisible();
```
richardsd, line 357: Same here, in the then you should assert some conditions.

packages/nuxeo-web-ui-ftest/features/step_definitions/document.js
richardsd: Leaving @Gabez0r to discuss a bit about the structure of the code.

richardsd: All step definitions look pretty generic to be in the `document.js` file, except the `I can't edit the Note`, but I wouldn't be surprised to see it here if is an exception to the rule.

elements/nuxeo-document-attachments/nuxeo-document-attachments.js
mnixo: Yes, I noticed that in other elements the method is named `_isAvailable` and it's used to hide/show itself. In this case we're just looking at the dropzone part. Plus, the method does way more than just checking for the write permission.

ftest/features/edit-note.feature
mnixo: Ah yes, that's a mistake.

packages/nuxeo-web-ui-ftest/features/step_definitions/document.js
mnixo: I don't know, the examples that I found to expect something to not be visible would use this format. I can try to see what happens if I don't specify it.

packages/nuxeo-web-ui-ftest/features/step_definitions/document.js
mnixo: Doesn't `waitForVisible` timeout after a certain amount of time if the `<nuxeo-dropzone>` element is not visible? I will add it just be sure, though.

packages/nuxeo-web-ui-ftest/features/step_definitions/document.js
Gabez0r: > Is this the right place for step definitions this specific?

We do not have clear rules for this. We could create a separate step definitions file for blob related steps, for example.

> Using el.element() here is probably wrong, where should I create proper helpers?

We use these mostly on page helpers and try to use the helpers instead on the step definitions. But please, see my comments above, it seems you already have the helpers you need.

> Is there a way to generalize this logic so we don't repeat it so much? 

Maybe already answered above?

ftest/features/upload.feature
```diff
@@ -0,0 +1,24 @@
...
+  Background:
+    Given user "John" exists in group "members"
+    And I login as "John"
+    And I have a File document
+
+  Scenario: Blob upload options for Read permission
+    Given I have permission Read for this document
+    When I browse to the document
+    Then I can't see the option to add a main blob
+    When This document has file "sample.png" for content
```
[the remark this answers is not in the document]
mnixo: For this case in particular I will keep this line because the current user can't actually upload anything as "John" only has "Read" permission. I will replace it on the scenario below with the interactive step though!

ftest/features/edit-note.feature
```diff
@@ -6,10 +6,36 @@ Feature: Edit Note
...
+
+  Scenario Outline: Edit <format> Note with Read permission
+    Given I have permission Read for this document
+    And I have a <format> Note
+    When I browse to the document
+    Then I can't edit the Note
+
+  Examples:
+  | format   |
+#  | HTML     |
```
richardsd, line 18: Out of curiosity, do we have issues with the HTML? 😬

mnixo: I think we do, @nmpcunha disabled it a while back in the context of [NXP-28691](https://jira.nuxeo.com/browse/NXP-28691).

nmpcunha: Indeed. When the editor was replaced it broke the HTML scenario and, at the time, we decided to comment it out "temporarily". Then I guess nobody followed the its track 😅 🤷🏼‍♂️

richardsd (approved):

semisse (approved):

PR #1000, opened 2020-08-19
SUPINT-1759: new typography

semisse:


themes/base.js
```diff
@@ -27,22 +27,52 @@ const template = html`
           display: none !important;
         }
 
-        /* headings */
         h1,
+        .headline-1 {
```
Gabez0r, line 31: Wouldn't like to commit to these new `headline-*` classes unless we really have to. And from the looks of it, maybe we can just move on with the HTML heading tags?

semisse:
I'm going to close this PR and replace it with a new one: https://github.com/nuxeo/nuxeo-web-ui/pull/1064

PR #1006, opened 2020-08-28
SUPINT-1709: grid poc

Gabez0r:


elements/nuxeo-grid/nuxeo-grid.js
```diff
@@ -0,0 +1,212 @@
...
+    };
+  }
+
+  _updateGrid() {
+    let style = `
+  :host {
+    display: grid;
+    grid-template-columns: ${this.cols && this.cols > 1 ? '1fr '.repeat(this.cols) : 'auto'};
+    grid-template-rows: ${this.rows && this.rows > 1 ? 'auto '.repeat(this.rows) : 'auto'};
+    ${this.gap ? `grid-gap: ${this.gap}` : ''};
```
richardsd, line 140: How does it work if we specify both the `gap` and `columnGap` or `rowGap`?

Gabez0r: We just let the browser decide. Usually, column and row gap have precedence.

elements/nuxeo-grid/nuxeo-grid.js
```diff
@@ -0,0 +1,212 @@
...
+    ${this.columnGap ? `grid-column-gap: ${this.columnGap};` : ''}
+    ${this.rowGap ? `grid-row-gap: ${this.rowGap};` : ''}
+    ${this.alignItems ? `align-items: ${this.alignItems};` : ''}
+    ${this.justifyItems ? `justify-items: ${this.justifyItems};` : ''}
+  }
+
+    `;
+    Array.from(this.children)
```
richardsd, line 148: Is this a polymer helper to get the children elements in the DOM? Never saw it

Gabez0r: It's part of the DOM api. [Parent nodes expose this property to get the children](https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/children).

richardsd: ah, my bad (stupidity should say) 🤦 `this` is a DOM node by itself

elements/nuxeo-grid/nuxeo-grid.js
richardsd: ok 👍

PR #1008, opened 2020-09-02
WEBUI-38: update to selenium-standalone@6.19.0

nelsonsilva:


RSalem07 (approved):

nmpcunha (approved):
We should also include it on https://github.com/nuxeo/plugin-nuxeo-web-ui, right? For 10.10

PR #1010, opened 2020-09-03
WEBUI-41: refresh clipboard based on content update

richardsd:
Ok, I'm opening the PR for further discussion on the topic.
- `nuxeo-clipboard` is using the `nuxeo-document-storage` to store the documents in the clipboard. 
- `nuxeo-document-storage` is [updating](https://github.com/nuxeo/nuxeo-web-ui/blob/master/elements/nuxeo-document-storage/nuxeo-document-storage.js#L85) the `documents` array and [notifying](https://github.com/nuxeo/nuxeo-web-ui/blob/master/elements/nuxeo-document-storage/nuxeo-document-storage.js#L95) the `nuxeo-clipboard`
- However, `nuxeo-clipboard` is not listening to array splices or any update inside the `documents` property (`!canPaste(documents, targetDocument)` is only triggered when the `documents` changes).

What do you think would be a good approach to handle this? 
I thought about listening to splices (`documents.splices`) or even all updates (`documents.*`), but I am wondering what you usually do in similar cases.

### DIFF
--- a/elements/nuxeo-clipboard/nuxeo-clipboard.js
+++ b/elements/nuxeo-clipboard/nuxeo-clipboard.js
@@ -172,7 +172,7 @@ Polymer({
             id="paste"
             on-tap="execute"
             data-op="Document.Copy"
-            disabled="[[!canPaste(documents, targetDocument)]]"
+            disabled="[[!canPaste(documents, targetDocument, documents.splices)]]"
             noink
             class="primary clear"
           >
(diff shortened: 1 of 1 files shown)


richardsd:
copy_move_clipboard_buttons

richardsd:
Pretty much the approach I was following. There is no other way to detect changes in the array without listening for [splices](https://stackoverflow.com/questions/54599020/how-to-notify-array-mutation-to-parent-component-in-polymer-3).

nmpcunha (changes requested):
It seems that the commit message has the wrong JIRA number

richardsd:
> It seems that the commit message has the wrong JIRA number

I am wondering what the heck happened here 🤔

nmpcunha (approved):

semisse (approved):

PR #1011, opened 2020-09-03
WEBUI-35: use WriteVersion instead of Write permission

mnixo:


semisse (approved):

richardsd (dismissed):

semisse (approved):

semisse (dismissed):

elements/nuxeo-publication/nuxeo-unpublish-button.js
```diff
@@ -68,7 +68,7 @@ Polymer({
   },
 
   _isAvailable() {
-    return this.document && this.document.isProxy && this.hasPermission(this.document, 'Write');
+    return this.document && this.document.isProxy && this.hasPermission(this.document, 'WriteVersion');
```
nelsonsilva, line 71: what about the publish button ? 🤔

richardsd: From what I could see in the [nuxeo-publish-button](https://github.com/nuxeo/nuxeo-web-ui/blob/master/elements/nuxeo-publication/nuxeo-publish-button.js#L162), it is using the [`isPublishable`](https://github.com/nuxeo/nuxeo-elements/blob/master/ui/nuxeo-filters-behavior.js#L229) method from the `nuxeo-filters-behavior` and that one needs to be updated as well, right?

mnixo: So it seems. I will create an ELEMENTS ticket for that.

richardsd (approved):

semisse (approved):

PR #1016, opened 2020-09-08
WEBUI-44: enqueue debounce to prevent request cancelling.

richardsd:
Don't see an easy option for this one. The problem is caused by two elements ([nuxeo-search-form](https://github.com/nuxeo/nuxeo-web-ui/blob/master/elements/search/nuxeo-search-form.js#L922) and [nuxeo-results](https://github.com/nuxeo/nuxeo-web-ui/blob/master/elements/nuxeo-results/nuxeo-results.js#L454)) performing a `fetch` in the page provider which debounces requests. 

The debounce itself is not bad, but if the request made by the `nuxeo-search-form` is canceled because the `nuxeo-results` performs another request due to a `view change`, then we get this issue. The `nuxeo-search-form` expects the request to resolve to [remove the loading state](https://github.com/nuxeo/nuxeo-web-ui/blob/master/elements/search/nuxeo-search-form.js#L925).

Feedback is very welcome 😄

### DIFF
--- a/elements/nuxeo-results/nuxeo-results.js
+++ b/elements/nuxeo-results/nuxeo-results.js
@@ -30,6 +30,8 @@ import '@nuxeo/nuxeo-ui-elements/widgets/nuxeo-sort-select.js';
 import '../nuxeo-selection/nuxeo-selection-toolbar.js';
 import { Polymer } from '@polymer/polymer/lib/legacy/polymer-fn.js';
 import { html } from '@polymer/polymer/lib/utils/html-tag.js';
+import { timeOut } from '@polymer/polymer/lib/utils/async.js';
+import { Debouncer } from '@polymer/polymer/lib/utils/debounce.js';
 
 /**
 An element to display results from a page provider.
@@ -395,20 +397,16 @@ Polymer({
 
   fetch() {
     return new Promise((resolve, error) => {
-      this.debounce(
-        'fetch',
-        () => {
-          if (this.view) {
-            this.view
-              .fetch()
-              .then(resolve)
-              .catch(error);
-          } else {
-            resolve();
-          }
-        },
-        100,
-      );
+      this._fetchDebouncer = Debouncer.debounce(this._fetchDebouncer, timeOut.after(100), () => {
+        if (this.view) {
+          this.view
+            .fetch()
+            .then(resolve)
+            .catch(error);
+        } else {
+          resolve();
+        }
+      });
     });
   },
 


richardsd:
Trying to assert if that actually fixes the issue.

richardsd:
I pushed a commit with a contribution to the drawer so we could assert the behavior.

semisse:
Just tested your solution on my machine and it works fine 💪

elements/search/simple/nuxeo-simple-search-results.html
```diff
@@ -0,0 +1,114 @@
...
+      nuxeo-justified-grid {
+        display: block;
+        position: relative;
+        min-height: calc(100vh - 130px - var(--nuxeo-app-top));
+      }
+
+      nuxeo-justified-grid {
+        height: calc(100vh - 10em);
+      }</style
+    ><nuxeo-results
```
mnixo, line 21: Not sure if it's the formatter doing this but it looks a bit funky.
```suggestion
      }
    </style>
    <nuxeo-results
```

package.json
```diff
@@ -82,7 +82,7 @@
     "format:eslint": "eslint --ext .js,.html . --fix",
     "lint:prettier": "prettier \"**/*.{js,html}\" --list-different || (echo '↑↑ these files are not prettier formatted ↑↑' && exit 1)",
     "format:prettier": "prettier \"**/*.{js,html}\" --write",
-    "lint": "npm run lint:polymer && npm run lint:eslint && npm run lint:prettier",
+    "lint": "npm run lint:polymer && npm run lint:eslint",
```
mnixo, line 85: Are you sure?

mnixo (approved):

mnixo:
(didn't mean to approve 🤦‍♂️, we should clarify the `prettier` thing before merging)

elements/search/simple/nuxeo-simple-search-results.html
semisse: AFAIK these files are only meant to test the solution suggested

richardsd: Yes, please ignore the second commit. Not going to merge that one.

package.json
richardsd: Please ignore the second commit, just used for testing. I'll drop it later.

richardsd:
> (didn't mean to approve 🤦‍♂️, we should clarify the `prettier` thing before merging)

@mnixo please ignore the second commit. As I said above, it is used just for having some contributions available in the preview 😄

semisse (approved):

richardsd:
Pipeline at https://qa.nuxeo.org/jenkins/job/web-ui-pipeline-rdias/10/

mnixo (approved):

PR #1031, opened 2020-10-01
WEBUI-37: use RemoveChildren instead of Write/Everything permission

mnixo:


elements/nuxeo-document-bulk-actions/nuxeo-delete-documents-button.js
```diff
@@ -136,13 +136,10 @@ Polymer({
...
   /*
-   * Checks if a single given document has 'Everything' permission to delete or 'Remove' to trash
+   * Checks if a single given document has the 'RemoveChildren' permission to trash
    */
   _docHasPermissions(document) {
-    return (
-      !this.isUnderRetentionOrLegalHold(document) &&
-      (this.hasPermission(document, 'Everything') || (!this.hard && this.hasPermission(document, 'Write')))
-    );
+    return !this.isUnderRetentionOrLegalHold(document) && !this.hard && this.hasPermission(document, 'RemoveChildren');
```
Gabez0r, line 142: Shouldn't we use `Remove` here instead of `RemoveChildren`? This applies to the documents that are being deleted, right?

richardsd: Maybe we should clarify that on the ticket as well, otherwise it could be confusing.

Gabez0r: It's already in a [comment](https://jira.nuxeo.com/browse/WEBUI-37?focusedCommentId=487541&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-487541). Miguel found out that we had further requirements to be able to delete documents in a Folderish document.

mnixo: There seem to be two different paths here:
a. We only check if the user has `RemoveChildren` for the current document in order to show the button.
b. We check every single permission that is needed to make sure that the user is indeed able to delete the documents: must have the `RemoveChildren` on the container, must have the `AddChildren` on the container (so that the deleted documents are properly trashed) and must have the `Remove` on each document that is selected.

Gabez0r: But right now, we are not doing either of them right? If I understood correctly, we are doing a mix: check if every document to be deleted (the selection, not the current document, the parent) has the `RemoveChildren` permission.

I think we should do b), but we don't have information about the current document in this element, do we? If so, we could do the children check (that we "must have the Remove on each document that is selected") now and create an improvement ticket to try to address the parent check later. This will require further thinking, because we will require the parent's info which will be tricky to get on search, where you might not have a single parent document.

mnixo:
These changes end up not achieving much, we'll handle the other questions in their own contexts.
I also ended up removing the check for `this.hard`, as we don't care if it's a hard delete or trash delete, as long as the user has the `Remove` permission.

elements/nuxeo-document-bulk-actions/nuxeo-delete-documents-button.js
```diff
@@ -136,13 +136,10 @@ Polymer({
   },
 
   /*
-   * Checks if a single given document has 'Everything' permission to delete or 'Remove' to trash
+   * Checks if a single given document has the 'Remove' permission to delete/trash
    */
   _docHasPermissions(document) {
-    return (
-      !this.isUnderRetentionOrLegalHold(document) &&
-      (this.hasPermission(document, 'Everything') || (!this.hard && this.hasPermission(document, 'Write')))
```
Gabez0r, line 144: Looks good to me. Does it work as expected with `hard` set to true?

mnixo: So it seems. [The `hard` property is set to `true` when the element is declared for the `TRASH_RESULTS_SELECTION_ACTIONS`](https://github.com/nuxeo/nuxeo-web-ui/blob/5bbf2d3b32a9fb5d9ba43a09d238b09ae8cc177b/elements/nuxeo-web-ui-bundle.html#L702-L706), so you can easily validate this by:
1. Creating a Folder with 2 Files inside.
2. Give the following permissions to a second user for the Folder: `Read`, `AddChildren` and `RemoveChildren`.
3. Give the following permissions to a second user for the 2 Files: `Remove`.
4. With the second user, access the Folder and delete the 2 Files (`hard` is `false`).
5. In the "TRASH" tab, permanently delete the 2 Files (`hard` is `true`).

Gabez0r (approved):

richardsd (approved):

PR #1033, opened 2020-10-06
NXP-29490: bypass service worker for API requests

nelsonsilva:


mnixo (dismissed):

Gabez0r (approved):

index.html
```diff
@@ -121,5 +121,9 @@
       };
       /* eslint-enable no-var, no-unused-vars */
     </script>
+
+    <!-- <script>
```
richardsd, line 125: Do we need this block? Or it is used just for reference?

nelsonsilva: It's to ease doing sw related work that's all

PR #1035, opened 2020-10-07
WEBUI-64: prevent triggering search when form is invalid

mnixo:
Based on the [existing logic for `<nuxeo-document-create>`](https://github.com/nuxeo/nuxeo-web-ui/blob/e83d9d314edfc65e13470c6295df84a7885770b3/elements/document/nuxeo-document-create.js#L348-L360).

### DIFF
--- a/elements/search/nuxeo-search-form.js
+++ b/elements/search/nuxeo-search-form.js
@@ -686,10 +686,24 @@ Polymer({
     }
   },
 
+  _validate() {
+    const innerLayout = this.$.layout.$.layout;
+    const valid = innerLayout.validate();
+    if (!valid) {
+      const elementsToValidate = innerLayout._getValidatableElements(innerLayout.element.root);
+      const invalidField = elementsToValidate.find((node) => node.invalid);
+      if (invalidField) {
+        invalidField.scrollIntoView();
+        invalidField.focus();
+      }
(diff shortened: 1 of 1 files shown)


semisse (dismissed):

richardsd (dismissed):

elements/search/nuxeo-search-form.js
```diff
@@ -690,6 +690,17 @@ Polymer({
...
+      const valid = innerLayout.validate();
+      if (!valid) {
+        const elementsToValidate = innerLayout._getValidatableElements(innerLayout.element.root);
+        const invalidField = elementsToValidate.find((node) => node.invalid);
+        if (invalidField) {
+          invalidField.scrollIntoView();
+          invalidField.focus();
+        }
+        return;
+      }
```
Gabez0r, line 703: This works fine for search forms in auto mode, but what about if we have auto toggled off and people have to hit the search button instead? Form validation will be skipped, right?

I think we could wrap this new piece of code in a new method that does validation and can return `true` or `false` accordingly. We could plug it here and in the `_search` method and use the boolean value to determine whether to execute the search or not.

richardsd (approved):

Gabez0r (approved):
👏

PR #1042, opened 2020-10-20
WEBUI-89: add selector parameter to document layout

PGoncalvesNX:
The purpose of this improvement is to add some precision in the selection of what fields we want to get inside the document_layout

### DIFF
--- a/packages/nuxeo-web-ui-ftest/pages/ui/browser/document_layout.js
+++ b/packages/nuxeo-web-ui-ftest/pages/ui/browser/document_layout.js
@@ -1,14 +1,14 @@
 import BasePage from '../../base';
 
 export default class DocumentLayout extends BasePage {
-  getField(field) {
+  getField(field, selector = '') {
     driver.waitForExist(this._selector);
     driver.waitForVisible(this._selector);
-    return this.el.element(`[name="${field}"]`);
+    return this.el.element(`${selector ? `${selector} ` : ''}[name="${field}"]`);
   }
 
-  getFieldValue(field) {
-    const fieldEl = this.getField(field);
+  getFieldValue(field, selector = '') {
+    const fieldEl = this.getField(field, selector);
     fieldEl.waitForVisible();
     return fixtures.layouts.getValue(fieldEl);
   }


packages/nuxeo-web-ui-ftest/pages/ui/browser/document_layout.js
```diff
@@ -1,14 +1,19 @@
...
 
-  getFieldValue(field) {
-    const fieldEl = this.getField(field);
+  getFieldValue(field, selector = '') {
+    let fieldEl;
+    if (selector === '') {
+      fieldEl = this.getField(field);
+    } else {
+      fieldEl = this.getField(field, selector);
+    }
```
RSalem07, line 16:
```suggestion
  return selector ? this.getField(field) : this.getField(field, selector);
```

PGoncalvesNX: Since the goal is to return `fixtures.layouts.getValue(fieldEl);` you probably mean I should do
`const fieldEl = selector ? this.getField(field, selector) : this.getField(field);`
It makes sense, thanks for the input!

packages/nuxeo-web-ui-ftest/pages/ui/browser/document_layout.js
```diff
@@ -1,14 +1,14 @@
...
     driver.waitForExist(this._selector);
     driver.waitForVisible(this._selector);
-    return this.el.element(`[name="${field}"]`);
+    return this.el.element(`${selector} [name="${field}"]`);
   }
 
-  getFieldValue(field) {
-    const fieldEl = this.getField(field);
+  getFieldValue(field, selector = '') {
+    const fieldEl = selector ? this.getField(field, selector) : this.getField(field);
```
andreacornaglia, line 11: we probably don't need that check here because we're already having selector as empty by default.

PGoncalvesNX: Yes, it's empty by default but not always. Then, we need the condition in order to use or not use the `selector`

packages/nuxeo-web-ui-ftest/pages/ui/browser/document_layout.js
```diff
@@ -1,14 +1,14 @@
 import BasePage from '../../base';
 
 export default class DocumentLayout extends BasePage {
-  getField(field) {
+  getField(field, selector = '') {
     driver.waitForExist(this._selector);
     driver.waitForVisible(this._selector);
-    return this.el.element(`[name="${field}"]`);
+    return this.el.element(`${selector} [name="${field}"]`);
```
nmpcunha, line 7: The only "issue" I can foresee is that when we don't pass a selector we will be searching for: ` [name="..]` (with a white space).
In theory it shouldn't be a big deal, but we can maybe move the ternary to here, to avoid that 🤷🏼‍♂️ 

Just an idea 😅

PGoncalvesNX: Do you mean doing this check? 
```
    const fieldEl = selector ? this.el.element(`${selector} [name="${field}"]`) : this.el.element(`[name="${field}"]`)
    return fieldEl;
```

nmpcunha: Or we could do it inside with a ternary or such thing.
```suggestion
    return this.el.element(`${selector ? selector + ' ' : '' } [name="${field}"]`);
```
Anyway, having the white space doesn't seem to be critical and all the tests passed 👍🏼  🙂 
Don't bother to much with the idea 😅

packages/nuxeo-web-ui-ftest/pages/ui/browser/document_layout.js
nmpcunha: Guess want @andreacornaglia is stating is that here we can simply rely on `this.getField(field, selector)` since we have a default value for the selector and, if it is not passed as an argument the fallback behaviour will be the existent one 🙂 
Or am I missing something? 🤔

nmpcunha (commented):
Left couple of notes.
Guess it may also be important for the UI team to understand how this is a good change. Providing a couple of examples would be nice 👍🏼 

Thanks!

packages/nuxeo-web-ui-ftest/pages/ui/browser/document_layout.js
PGoncalvesNX: Ok, then we didn't even need the condition that I had put in the first place, as well as in the override in aspera-ftests?

PGoncalvesNX: Took your suggestion into account :) Thanks for the input!

packages/nuxeo-web-ui-ftest/pages/ui/browser/document_layout.js
```diff
@@ -1,14 +1,14 @@
 import BasePage from '../../base';
 
 export default class DocumentLayout extends BasePage {
-  getField(field) {
+  getField(field, selector = '') {
     driver.waitForExist(this._selector);
     driver.waitForVisible(this._selector);
-    return this.el.element(`[name="${field}"]`);
+    return this.el.element(`${selector ? `${selector} ` : ''}[name="${field}"]`);
```
andreacornaglia, line 7: I believe just return this.el.element(`${selector}[name="${field}"]`); should work.

PGoncalvesNX: We need the space between the `selector` and `name` for the query to work. Also, that's how it was before Nuno's [suggestion](https://github.com/nuxeo/nuxeo-web-ui/pull/1042#discussion_r510090010) (with the spacing I mentioned)

nmpcunha: It can work, @andreacornaglia, but it should mean that the `selector` variable would need to include a white space, which is far from ideal from a developer perspective (who uses the method), IMHO.

nmpcunha (approved):
LGTM 👍🏼

packages/nuxeo-web-ui-ftest/pages/ui/browser/document_layout.js
andreacornaglia: ah indeed i now noticed the white space on selector :) thanks for the clarification!

andreacornaglia (approved):

Gabez0r (commented):
Could you please provide an example where we need this?

PGoncalvesNX:
Closing this PR because we are opting for adding names to the target elements, instead of needing to add this complexity to the helpers.

PR #1044, opened 2020-10-21
WEBUI-23: avoid resetting array params on splice/length events

mnixo:


elements/search/nuxeo-search-form.js
```diff
@@ -898,7 +898,10 @@ Polymer({
       // e.detail.path is params.prop_name eg: params.ecm_fulltext
       if (e.detail.path) {
         const param = e.detail.path.split('.')[1];
-        this.set(`params.${param}`, e.detail.value);
+        if (!e.detail.path.match(/params\.\w+\.(splices|length)/)) {
```
Gabez0r, line 901: We should make it match the beginning and the end of the line, otherwise something like `params.prop.splicesss` would be a match:
```suggestion
        if (!e.detail.path.match(/^params\.\w+\.(splices|length)$/)) {
```

Gabez0r (approved):

Gabez0r (approved):

semisse (dismissed):

richardsd (dismissed):

Gabez0r (approved):

richardsd (approved):

PR #1047, opened 2020-10-26
fix-NXP-28080-TasksInEndState

nmpcunha:
👋🏼 

The idea was to try to have a "middle term" solution for this issue, since it was raised by Support and needs backporting.
Discussing with @carolinagouveia, we got to the this proposal. The idea was basically:
- "warn" the users that they are seeing a task that has already been processed;
- remove all the possible actions that have side effects;
- still display the information introduced by the user who processed the task, but on a "read only" kind of view;
- display actions that don't have an impact nor generate server side errors (e.g. "Show Graph" or user tag);


nmpcunha:
The preview will help us to get a feeling of it, but this screenshot may also give you an idea 🙂 
Screenshot 2020-10-26 at 18 21 45

lisamcintyre:
@nmpcunha , regarding the task and the focus of this, this looks good to me.  In a perfect world, the link to the item would be more "visible" to the user (and parhaps part of the message), but I don't think that is the function of this ticket, and more around the default task view and the header with the file name. and too, this is a default view, and can be configured 😄

nmpcunha:
> @nmpcunha , regarding the task and the focus of this, this looks good to me. In a perfect world, the link to the item would be more "visible" to the user (and parhaps part of the message), but I don't think that is the function of this ticket, and more around the default task view and the header with the file name. and too, this is a default view, and can be configured 😄

Thanks for the review.
Indeed! And we should keep in mind that workflow (and its tasks) can be related to multiple documents.
This to say that the feature that you are suggesting may need further investigation and proper requirements.

lisamcintyre (approved):
for this ticket, this looks good to me.  I notice some default view issues with this version of webui, but not part of this "task" 😄

elements/workflow/nuxeo-document-task.js
```diff
@@ -100,27 +118,42 @@ Polymer({
...
+                <paper-button
+                  id="reassignBtn"
+                  noink
+                  dialog-confirm
+                  on-tap="_toggleAssignmentDialog"
+                  data-args="reassign"
+                  hidden$="[[!task.taskInfo.allowTaskReassignment]]"
+                  >[[i18n('tasks.reassign')]]</paper-button
+                >
+                <paper-button
```
andreacornaglia, line 147: as Lisa was asking, shouldn't we include the styling for this? class="secondary"

elements/workflow/nuxeo-document-task.js
```diff
@@ -86,6 +86,24 @@ Polymer({
         color: var(--nuxeo-primary-color, #0066ff);
         font-weight: bolder;
       }
+      
+      .info {
+        margin-bottom: 1rem;
```
Gabez0r, line 91: Please double check with Carolina. This currently maps to 13px, and [I think it should either be 12px or 16px](https://nuxeowiki.atlassian.net/wiki/spaces/DT/pages/1091797171/Spacing).

i18n/messages.json
```diff
@@ -1078,17 +1078,18 @@
   "tags.removedFromDocument": "The tag {0} has been removed from the document.",
   "tasks.actors.assigned": "Actors in this task:",
   "tasks.actors.delegated": "Delegated Actors in this task:",
+  "tasks.alreadyProcessed": "The task you are trying to seeing has already been completed. Please check the correspondent document for current status.",
```
carolinagouveia, line 1081: We should change this message to "The task you are trying to _see_ (...)"

elements/workflow/nuxeo-document-task.js
```diff
@@ -86,6 +86,24 @@ Polymer({
...
+      }
+      
+      .info > iron-icon {
+        color: var(--nuxeo-validated, #42BE65);
+        margin-inline-end: 0.5rem;
```
Gabez0r, line 96: Same here.

carolinagouveia (changes requested):

elements/workflow/nuxeo-document-task.js
nmpcunha: Since this is a support ticket, we can do that if we have clear guidelines. But, it is out of scope... so, it would be done on a separate commit and only for master.
Does anyone have clear instructions about this? @lisamcintyre , @andreacornaglia ?

andreacornaglia: @nmpcunha makes sense! I think @carolinagouveia should take a look but IMO the "delegate" button should be secondary and so the "cancel review" (that now is primary).

i18n/messages.json
nmpcunha: 🤦🏼 
Thanks! 😅

elements/workflow/nuxeo-document-task.js
```diff
@@ -142,30 +175,33 @@ Polymer({
...
           </div>
 
-          <nuxeo-layout
-            id="layout"
-            href="[[_href]]"
-            model="[[_model]]"
-            error="[[i18n('documentView.layoutNotFound', task.nodeName)]]"
-            on-element-changed="_elementChanged"
-          ></nuxeo-layout>
+          <div class$="[[_computeLayoutVisibility(task)]]">
```
Gabez0r, line 178: Unfortunately, this does not prevent one from using the keyboard and reaching out to the elements and inputing data. I guess we should either:
1) avoid simulating this disabled state, as the fact that we can still input data can be considered a bug?
2) do the required work on nuxeo-layout so we can have a proper way to disable the fields

Gabez0r (commented):
Great work. I just have a couple of questions:
1) I'm not sure the disclaimer on the top has enough emphasis.
2) Not sure if showing the "View Graph" action is very helpful, since it will show the graph as it is currently and not how it was when the task was still available. It might still make sense for consistency reasons ofc, but I was just wondering.

elements/workflow/nuxeo-document-task.js
carolinagouveia: As discussed with @nmpcunha, we'll apply the style of the text button for both "reassign" and "delegate" actions.

nmpcunha: For the "Cancel Review", there is no possibility to to it.
The buttons (or actions) are retrieved from server side and they can be multiple (i.e. more than the usual two) since they are configured by the users on Studio side. This would imply a whole refactor, end-to-end, in order to give some field to the UI so we could know which kind of style to apply to which button.

elements/workflow/nuxeo-document-task.js
nmpcunha: Double checked! We decided to keep like it is.
Thanks!

elements/workflow/nuxeo-document-task.js
nmpcunha: Indeed!
Option _2_ is clearly out of scope, and TBH, not sure if we even want that balancing the use cases and the effort 🤷🏼‍♂️ 

It is true that the users can use the keyboard to access the fields (it is a matter of a visual styling to, somehow simulate that they are disabled). Anyway, guess this is a decision to @lisamcintyre and/or @carolinagouveia.
I don't expect a lot of problems with that and if we receive a bug stating that we can change the values using the keyboard, guess we have arguments to explain it. Otherwise, we can go with the fields "active", but it seems to provide a worse UX.

Let's wait for more feedback on this! Thanks!

carolinagouveia: I do understand that option 2 is out of this ticket's scope but I prefer to do the required work and having the fields properly disabled, to avoid future issues.

nmpcunha: Does that mean that on this ticket we should remove the opacity and pointer events and let the fields be editable?
Or are you suggesting to keep the current approach (with opacity and pointer events disabled) and try to create a ticket to assess and schedule some work to have that ability on `nuxeo-layout`.

carolinagouveia: The second option, create the ticket to work on disabling it on nuxeo-layout.

nmpcunha:
Think I've covered all the suggestions and also added a 🎁  on the third commit (only for master), correcting a visual/styles bug.

Please let me know if it is ok from your side 👍🏼

elements/workflow/nuxeo-document-task.js
nmpcunha: Thanks!
I've kept that piece of code.

Regarding the ticket creation, @Gabez0r how do you want us to contribute on that? (cc @lisamcintyre)

lisamcintyre (commented):
I just now took a look on the preview, different browsr than what I used before, so not cached, and I still see some of the requested changes not done (lile carolina's wording suggestion.  last note I see from Nuno says it was updated.  Did i miss something 😓 I even started a new task on a new item (knowing that wf changes might only appear on newly instantiated flows, or at least they used to).
Screen Shot 2020-10-27 at 2 05 12 PM

andreacornaglia (dismissed):

semisse:
Can you please consider the possibility of putting the icon and the message side-by-side, instead of wrapping the text around it? 
97349947-1ac45900-185e-11eb-8e7d-81dd3b7e6052

nmpcunha:
> Can you please consider the possibility of putting the icon and the message side-by-side, instead of wrapping the text around it?
> 97349947-1ac45900-185e-11eb-8e7d-81dd3b7e6052

Hey! 👋🏼 

Done! Should be available when the preview up and running.
Had the need to define a `min-width` to the icon, just to try to don't expose the way this specific screen reacts to breakpoints. If you prefer to remove it, just let me know. (cc @carolinagouveia )

carolinagouveia:
> > Can you please consider the possibility of putting the icon and the message side-by-side, instead of wrapping the text around it?
> > 97349947-1ac45900-185e-11eb-8e7d-81dd3b7e6052
> 
> Hey! 👋🏼
> 
> Done! Should be available when the preview up and running.
> Had the need to define a `min-width` to the icon, just to try to don't expose the way this specific screen reacts to breakpoints. If you prefer to remove it, just let me know. (cc @carolinagouveia )
> 
> Thank you!


I'm still seeing the old message on the latest preview.
image

nmpcunha:
Indeed @carolinagouveia , it seems that the preview generation for this repo has some issues. Seems to be stuck on the first version of things 😕 
Anyway, I'll wait for code review too, but please feel free to reach me for screenshots or quick demo.

Gabez0r (approved):

mnixo (approved):

carolinagouveia (approved):

PR #1051, opened 2020-10-28
WEBUI-90: load but disable spreadsheet by default

nelsonsilva:


addons/nuxeo-spreadsheet/index.js
```diff
@@ -15,9 +15,3 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 import './elements/nuxeo-spreadsheet-button.js';
-
-import html from './nuxeo-spreadsheet.html';
```
Gabez0r, line 19: Shouldn't we delete `nuxeo-spreadsheet.html` as well?

nelsonsilva: we should, yes 👍

Gabez0r (approved):

PR #1063, opened 2020-11-05
NXP-29175: cleanup/improve messages for audit

RSalem07:


nmpcunha (commented):
Thanks for your work on this.
I would suggest to squash both commits in terms of `i18n` files.

The `nuxeo-retention-events` change, I would ask you to perform it already on the new repo: https://github.com/nuxeo/nuxeo-retention
Those files on Web UI are already scheduled to be removed.

RSalem07:
Thanks @nmpcunha perhaps I missed something but Yesterday you asked me to make a separate commit for the retention part. If I understand correctly:
- Move to WEB-UI ticket
- Squash the two commits into only one
- Make the needed modification on new repo

Thx

nmpcunha:
> Thanks @nmpcunha perhaps I missed something but Yesterday you asked me to make a separate commit for the retention part. If I understand correctly:
> 
> * Move to WEB-UI ticket
> * Squash the two commits into only one
> * Make the needed modification on new repo
> 
> Thx

Indeed. 😅 
Sorry about the back and forward, but in the meanwhile we already have a PR opened for removing retention from WebUI (https://github.com/nuxeo/nuxeo-web-ui/pull/1068) and the move to the new repo is already merged (https://github.com/nuxeo/nuxeo-retention/pull/1)

So, regarding the actions needed you got it right 🙂 
The only caveat is that we don't want to modify any file under `addons/nuxeo-retention` which means that the change made on `nuxeo-retention-events` must only be done on the new repo 👍🏼

RSalem07:
@nmpcunha the needed modifications are done:

- I moved the NXP-29175 to [WEBUI-115](https://jira.nuxeo.com/browse/WEBUI-115)
- I reverted the `nuxeo-retention-events` as it was on master and I will rebase once the other PR is merged #1068
- I squashed the tow commits into one with the correct jira ticket (web-ui one)

i18n/messages-ar.json
```diff
@@ -435,18 +435,6 @@
   "documentEdit.layoutNotFound": "Failed to find edit layout for {0}.",
   "documentEdit.saveError": "Failed to save.",
   "documentEditAction.tooltip": "Edit document",
-  "documentHistory.category": "Category",
```
Gabez0r, line 438: Don't think we can simply remove these labels, as they can still be used by other custom elements. We must keep them but document them as deprecated on the upgrade notes. Eventually we will be able to delete them.

lisamcintyre: @Gabez0r thanks, I think this was something Bertrand was trying to ask me about today.  I think he suggested we follow the "standard" deprecation that Platform tends to use: 'announce' deprecation and then in the following year, remove.  @RSalem07 let's stick with this process if possible (at least if I've interpreted Bertrand's suggestion properly with how it relataes to what @Gabez0r says.

lisamcintyre (commented):
Based on info from Gabriel and Bertrand, we might need to leave in and "deprecate" the "old" values for a year or so since the labels might be used elsewhere by clients.

i18n/messages-ar.json
nmpcunha: Just to give my 2 cents (on the other side of the coin 🙂 ):
Indeed, someone may be using that. AFAIK we don't have a process to deprecate labels nor to show it to the users/developers, right @Gabez0r? 🤷🏼‍♂️ 

This to say that I would not be "shocked" by removing them and documenting that on the upgrade notes. Will it break the screen? Yes, but we'll do a major release soon, so it would be the chance to perform this kind of change, otherwise it will be even harder (maybe on the next year, LTS? 😅 ).

Gabez0r: I agree with removing deprecated stuff on an LTS, but it needs to be deprecated in the LTS before, so we give time for people to adapt. Otherwise, we'll get a bunch of support tickets requesting the labels to be added back. So I think we should keep them, and document on the ticket's upgrade notes that the labels are deprecated and should be replaced by some other labels instead. Something like:
```
The follow labels were deprecated:
XXXX (use YYYY instead)
ZZZZ (use NNNN instead)
```
We should also update https://doc.nuxeo.com/nxdoc/next/web-ui-upgrade-notes/ to feature this deprecation, but we already have a ticket to update this: https://jira.nuxeo.com/browse/NXDOC-2221.

nmpcunha: Yesterday we've discussed a little bit on how to proceed with labels deprecation.
For this PR, to don't postpone it, we should keep the obsolete keys. But, hopefully soon, we'll try to scratch a process for that 🙂 
Thanks @Gabez0r

RSalem07: As discussed I revert the modification on removing the oldest entries the ones prefixed by `documentHistory` they will be noted in the JIRA ticket as deprecated.

nmpcunha (approved):
LGTM! 👍🏼 

Please make sure the ticket `Upgrade Notes` are filled correctly 🙂 
And I guess we can remove the `wip` tag 😉

RSalem07:
Thx @nmpcunha, yes for the upgrade note already noted I was waiting for at least one approval :)  before going further

lisamcintyre (approved):
LGTM (as far as I can tell).  😄

PR #1064, opened 2020-11-05
WEBUI-107: apply new typography styles and vertical alignment

semisse:


richardsd (dismissed):

elements/document/nuxeo-picture-exif.js
```diff
@@ -28,6 +28,7 @@ Polymer({
   _template: html`
     <style>
       .properties label {
+        @apply --nuxeo-label;
```
Gabez0r, line 31: May we should to the same to `nuxeo-video-info` and `nuxeo-video-conversions`?

richardsd (approved):
This is awesome 👏

mnixo (approved):

PR #1065, opened 2020-11-06
WEBUI-111: add nuxeo-grid element

Gabez0r:
Please note that the scope of this ticket is just to add `nuxeo-grid`. All other changes are just for testing purposes.
TODO:
- [x] Functional fests
- [x] Document bug with video viewer (to be fixed under [ELEMENTS-1271](https://jira.nuxeo.com/browse/ELEMENTS-1271))
- [x] Discuss value coercion (do not change the value but throw the warning).
- [x] Discuss hard-coded responsiveness
- [x] Discuss handling of nuxeo-cards
- [x] Discuss how to document child level properties
- [x] Discuss how row gaps still persists if the row is completely empty (default css grid behavior)
- [x] Unit tests


elements/nuxeo-grid/nuxeo-grid.js
```diff
@@ -0,0 +1,371 @@
...
+@media ${mquery} {
+${css}
+}
+  `;
+}
+
+function buildGridStyle(grid, mquery) {
+  // TODO are we OK with this coercion system?
+  const cGrid = {};
+  cGrid.templateColumns = coerceValue(grid.templateColumns, /^(\d+fr|auto)$/, true, 'template-columns');
```
Gabez0r, line 131: WDYT about this strategy to reduce the public API, so that we don't have to support the whole spectrum of CSS values?

elements/nuxeo-grid/nuxeo-grid.js
```diff
@@ -0,0 +1,371 @@
...
+      ? Array(cGrid.rows)
+          .fill('auto')
+          .join(' ')
+      : 'auto')};
+  ${cGrid.gap ? `grid-gap: ${cGrid.gap}` : ''};
+  ${cGrid.columnGap ? `grid-column-gap: ${cGrid.columnGap};` : ''}
+  ${cGrid.rowGap ? `grid-row-gap: ${cGrid.rowGap};` : ''}
+  ${cGrid.alignItems ? `align-items: ${cGrid.alignItems};` : ''}
+  ${cGrid.justifyItems ? `justify-items: ${cGrid.justifyItems};` : ''}
+  /* --nuxeo-card-margin-bottom: 0; this will not work on all places and causes misalignments */
```
Gabez0r, line 162: Right now we have a issue when using nuxeo cards, because they already have a margin bottom. This means that we usually only set a column gap and not a row gap to make up for this. But we should handle this better. Any ideas?

elements/nuxeo-grid/nuxeo-grid.js
```diff
@@ -0,0 +1,371 @@
...
+    });
+    // RESPONSIVENESS
+    grid.templateColumns = `1fr`;
+    grid.templateRows = `auto`;
+    // TODO should this media query be hardcoded?
+    // should this behavior be toggleable?
+    // should this consider the screen with or height?
+    // we could leave this hardcoded for now and if we get complaints, we can introduce a global property or a
+    // element property.
+    style += `${buildGridStyle(grid, '(max-width: 1024px)')}`;
```
Gabez0r, line 313: I don't like it, but I can leave with a hardcoded media query, provided that we accept that we might have to introduce later a element property and/or global config property to control this behavior. WDYT?

elements/nuxeo-grid/nuxeo-grid.js
```diff
@@ -0,0 +1,371 @@
...
+    // TODO how to document these properties and what their values are?
+    return {
+      COL: 'nuxeo-grid-col',
+      COLSPAN: 'nuxeo-grid-colspan',
+      ROW: 'nuxeo-grid-row',
+      ROWSPAN: 'nuxeo-grid-rowspan',
+      CHILDID: 'nuxeo-grid-child-id',
+      ALIGN: 'nuxeo-grid-align',
+      JUSTIFY: 'nuxeo-grid-justify',
+    };
```
Gabez0r, line 32: Since this will only be attributes and not properties on elements, I don't think we have a good way to document them, so that this documentation can be leveraged on Designer. Any ideas?

elements/nuxeo-grid/nuxeo-grid.js
```diff
@@ -0,0 +1,371 @@
...
+      <slot></slot>
+    `;
+  }
+
+  static get properties() {
+    return {
+      /**
+       * Number of columns. Will be ignored if templateColumns is defined.
+       */
+      cols: {
```
Gabez0r, line 217: Should this be named `columns` instead, to be consisted with the other properties above?

elements/nuxeo-grid/nuxeo-grid.js
richardsd: I agree with you, we can later deal with it. And how about putting this behavior toggleable? Do you think people would actually disable the default behavior? I think that most people will be happy with the default media query and won't disable it, but maybe it could be useful for a minority. WDYT?

richardsd (changes requested):

elements/nuxeo-grid/nuxeo-grid.js
Gabez0r: I also think most people will be OK with it. But there's always the exception...

elements/nuxeo-grid/nuxeo-grid.js
```diff
@@ -0,0 +1,371 @@
...
+  get id() {
+    return this.el.getAttribute(Child.ATTRS.CHILDID);
+  }
+
+  set id(val) {
+    this.el.setAttribute(Child.ATTRS.CHILDID, val);
+  }
+
+  get col() {
+    return this.el.getAttribute(Child.ATTRS.COL) || '';
```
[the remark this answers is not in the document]
Gabez0r: We can, but we still have the exception of the get id, which does not fallback to ''. So we cannot include the ` || ''` on this method, or it would have to be configurable.

elements/nuxeo-grid/nuxeo-grid.js
```diff
@@ -0,0 +1,371 @@
...
+        type: String,
+      },
+      /**
+       * Gap between columns, in pixels (e.g. 16px).
+       */
+      columnGap: {
+        type: String,
+      },
+      /**
+       * Default item alignment on the block axis (typically vertical). Can be `stretch`, `center`, `stard` and `end`.
```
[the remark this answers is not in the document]
Gabez0r: The best we could do would be to annotate this with jsdoc. Are we able to read these back from the comments on Desginer?

elements/nuxeo-grid/nuxeo-grid.js
richardsd: We can always expose the toggle, if it makes up for the exception 😄

elements/nuxeo-grid/nuxeo-grid.js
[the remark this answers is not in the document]
richardsd: Oh, you're right. Didn't notice the `id` was different. In this case, maybe we don't need to introduce the method, since we cannot include the `||` 🤷‍♂️

elements/nuxeo-grid/nuxeo-grid.js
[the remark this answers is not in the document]
richardsd: I think it will depend on the structure we adopt to annotate the property.

elements/nuxeo-grid/nuxeo-grid.js
Gabez0r: But we should rather expose this later right? When we actually know the use case and know if this needs to be local or global config.

elements/nuxeo-grid/nuxeo-grid.js
```diff
@@ -0,0 +1,412 @@
...
+ limitations under the License.
+ */
+import '@polymer/polymer/polymer-legacy.js';
+import { html } from '@polymer/polymer/lib/utils/html-tag.js';
+import '@nuxeo/nuxeo-elements/nuxeo-element.js';
+
+class Child {
+  static get ATTRS() {
+    return {
+      COLUMN: 'nuxeo-grid-column',
```
Gabez0r, line 24: I guess that we should not be naming it this way if we ever want to use native elements (divs for example). A div with a custom attribute works just fine, but it is not valid html and would fail on a validator. We should use a `data-*` attribute instead here, but I guess it will make the name more verbose. Also, I'm not sure how much of a problem this actually is, since we probably a lot of more stuff going on that would result in invalid html.

richardsd (dismissed):

elements/nuxeo-grid/nuxeo-grid.js
mnixo: I'd vote for replacing `nuxeo-*` with `data-*`, if that means being compliant with the spec. Although not very pretty, it shouldn't make much of a difference, and it's not like the `nuxeo-*` prefix was doing much anyway.

Gabez0r: I agree. Thank you. I will create a poll on slack to decide what to replace, since I have seen people in favor of other approaches.

mnixo (approved):
👏

plugin/itests/addon/src/main/resources/web/nuxeo.war/ui/document/gridfile/nuxeo-gridfile-create-layout.html
```diff
@@ -0,0 +1,126 @@
...
+
+    <nuxeo-grid columns="2" gap="16px">
+      <nuxeo-input
+        role="widget"
+        label="[[i18n('title')]]"
+        name="title"
+        value="{{document.properties.dc:title}}"
+        autofocus
+        required
+        data-column="1"
```
nelsonsilva, line 35: we could omit the explict row and column for these two, right?
could be nice to do so here to be sure things don't break in such cases

plugin/itests/addon/src/main/resources/web/nuxeo.war/ui/nuxeo-document-sidebar.html
```diff
@@ -0,0 +1,235 @@
...
+        }
+        .scroller {
+          top: 0;
+          position: relative;
+          display: block;
+        }
+      }
+    </style>
+
+    <div class="side">
```
nelsonsilva, line 114: can we drop this wrapper div and move the styles to `:host`?

plugin/itests/addon/src/main/resources/web/nuxeo.war/ui/nuxeo-document-sidebar.html
```diff
@@ -0,0 +1,235 @@
...
+        </div>
+
+        <template is="dom-if" if="[[hasFacet(document, 'NXTag')]]">
+          <div class="section">
+            <h3>[[i18n('documentPage.tags')]]</h3>
+            <nuxeo-tag-suggestion
+              document="[[document]]"
+              allow-new-tags
+              placeholder="[[i18n('documentPage.tags.placeholder')]]"
+              readonly="[[!isTaggable(document)]]"
```
nelsonsilva, line 140: As this is a boolean shouldn't we be using `readonly$=[[..]]` here ?

plugin/itests/addon/src/main/resources/web/nuxeo.war/ui/nuxeo-document-sidebar.html
```diff
@@ -0,0 +1,235 @@
...
+      },
+
+      detached() {
+        if (this.resizeObserver) {
+          this.resizeObserver.disconnect();
+        }
+      },
+
+      _toggleSize() {
+        if (window.matchMedia('(min-width: 1024px)').matches && this.expandedWidth && this.opened) {
```
nelsonsilva, line 198: we're observing the size of the element but doing responsiveness based on window size ? 🤔

plugin/itests/addon/src/main/resources/web/nuxeo.war/ui/nuxeo-document-sidebar.html
```diff
@@ -0,0 +1,235 @@
...
+              <nuxeo-document-comment-thread name="comments" uid="[[document.uid]]"></nuxeo-document-comment-thread>
+            </template>
+            <nuxeo-document-activity name="activity" document="[[document]]"></nuxeo-document-activity>
+          </iron-pages>
+        </div>
+      </div>
+    </div>
+  </template>
+  <script>
+    Polymer({
```
nelsonsilva, line 164: Shouldn't we be using class syntax here ?

elements/nuxeo-grid/nuxeo-grid.js
```diff
@@ -0,0 +1,412 @@
...
+ */
+import '@polymer/polymer/polymer-legacy.js';
+import { html } from '@polymer/polymer/lib/utils/html-tag.js';
+import '@nuxeo/nuxeo-elements/nuxeo-element.js';
+
+class Child {
+  static get ATTRS() {
+    return {
+      COLUMN: 'data-column',
+      COLUMNSPAN: 'data-columnspan',
```
nelsonsilva, line 25: i know we've reviewed these already but now looking at it I feel this should be `data-column-span` and `data-row-span` for consistency, wdyt?

elements/nuxeo-grid/nuxeo-grid.js
```diff
@@ -0,0 +1,412 @@
...
+  }
+  return removeEmptyLines(`@media ${mquery} {
+${css.replace(/^(.+)$/gm, '  $1') /* apply indentation */}
+}
+`);
+}
+
+function buildGridStyle(grid, validate = true) {
+  const cGrid = {};
+  cGrid.templateColumns = validateValue(grid.templateColumns, /^(\d+fr|\d+px|auto)$/, validate, 'template-columns');
```
nelsonsilva, line 174: this makes me think we should have proper setters where we enforce the validation... if tricky to do with a Polymer element perhaps we could do at least with the `Child` class (which could be called `Cell`no? )

elements/nuxeo-grid/nuxeo-grid.js
```diff
@@ -0,0 +1,412 @@
...
+import '@nuxeo/nuxeo-elements/nuxeo-element.js';
+
+class Child {
+  static get ATTRS() {
+    return {
+      COLUMN: 'data-column',
+      COLUMNSPAN: 'data-columnspan',
+      ROW: 'data-row',
+      ROWSPAN: 'data-rowspan',
+      CHILDID: 'data-child-id',
```
nelsonsilva, line 28: what's child id for ?

elements/nuxeo-grid/nuxeo-grid.js
```diff
@@ -0,0 +1,412 @@
...
+      templateRows: {
+        type: String,
+        reflectToAttribute: true,
+      },
+    };
+  }
+
+  _updateGrid() {
+    let style = '';
+    const children = Array.from(this.querySelectorAll(':scope > *')).map((child) => new Child(child));
```
nelsonsilva, line 328: don't we normally use `slot.assignedNodes()` to get this ?

elements/nuxeo-grid/nuxeo-grid.js
```diff
@@ -0,0 +1,412 @@
...
+          id,
+          justify,
+        },
+        false,
+      )}`;
+    });
+    style += wrapMediaQuery(responsiveStyle, '(max-width: 1024px)');
+    // END OF RESPONSIVENESS
+
+    const styleEl = document.createElement('style');
```
nelsonsilva, line 370: this looks weird to me... shouldn't we generate the style as part of the template function?

plugin/itests/addon/src/main/resources/web/nuxeo.war/ui/nuxeo-document-sidebar.html
Gabez0r: I tried, but we started to get all sorts of artifacts. I spent way too much time trying to fix it, for an element that we are just keeping for testing purposes.

plugin/itests/addon/src/main/resources/web/nuxeo.war/ui/document/gridfile/nuxeo-gridfile-create-layout.html
Gabez0r: Sure, why not!

plugin/itests/addon/src/main/resources/web/nuxeo.war/ui/nuxeo-document-sidebar.html
Gabez0r: Yes. This was an early prototype where I was still playing with with the ability of having responsiveness on this element, so we might have some mixed concepts here. I kept it here because it we might find it useful on how to do it later, but this is not used by the tests at all. Maybe it is better to just remove it?

elements/nuxeo-grid/nuxeo-grid.js
Gabez0r: To generate the css rule for that element. E.g.:
```
::slotted([data-child-id="1"]) {
  grid-column: 1;
  grid-row: 1;
}
```

plugin/itests/addon/src/main/resources/web/nuxeo.war/ui/nuxeo-document-sidebar.html
Gabez0r: Eh... it's on a test project where anything else uses class based syntax. Nor did the original element use class based syntax. But I guess we can start doing it.

elements/nuxeo-grid/nuxeo-grid.js
Gabez0r: In the case of the Child, we are not even validating when we set, but when we get from the attributes instead. Wouldn't it be weird to enforce validation on a getter? And I fear that having it in one way for the Child and another way for the Grid will make the code harder to maintain.

> Child class (which could be called `Cell` no? )

Not necessarily, since a child can take multiple cells.

elements/nuxeo-grid/nuxeo-grid.js
richardsd: It also increases readability here. Maybe a good call.

plugin/itests/addon/src/main/resources/web/nuxeo.war/ui/nuxeo-document-sidebar.html
nelsonsilva: Just noticed this is in itest... Nevermind

plugin/itests/addon/src/main/resources/web/nuxeo.war/ui/nuxeo-document-sidebar.html
nelsonsilva: Same, just realized this is in itest so better keep ES5 syntax since it's what VD supports

elements/nuxeo-grid/nuxeo-grid.js
Gabez0r: Need to be assignedElements, otherwise we will get text nodes as well. :grimacing:

elements/nuxeo-grid/nuxeo-grid.js
nelsonsilva: Ah! I see 👍 Thought we were going with inline styles or some pseudo selector like nth-child or nth-of-type (not sure these work though)

mnixo (approved):

elements/nuxeo-grid/nuxeo-grid.js
richardsd: Yes, you're right. We can do it later, when we have more information about the actual use case.

richardsd (approved):

richardsd (approved):

PR #1066, opened 2020-11-06
WEBUI-112: apply --nuxeo-label to nuxeo-document-info

semisse:


elements/nuxeo-document-info/nuxeo-document-info.js
```diff
@@ -44,6 +44,8 @@ Polymer({
       }
 
       .item label {
+        @apply --nuxeo-label;
+        line-height: 2.2rem;
```
nmpcunha, line 48: Since the `line-height` is already applied to the parent div, do we need to (re)apply it? 🤔

semisse: We need to apply it here to override the line-height coming from `--nuxeo-label`. Maybe we can delete the line-height on the `.item` class.

nmpcunha: It is also applying to the value, if I'm seeing it correctly.
Is the `line-height` from `--nuxeo-label` from the consistency changes? As it is right now on master, there is no `line-height` on it, right? 🤔  ([here](https://github.com/nuxeo/nuxeo-web-ui/blob/master/themes/base.js#L329-L338))

semisse: Yes, the new line-height is [defined here](https://github.com/nuxeo/nuxeo-web-ui/pull/1064/files#diff-b71744508f259def2cc5dfa607f4251445b84703c01837367743ebd3891880dfR295). And I also confirm we should maintain the line-height defined above, on the `item` class in order to be correctly aligned.

Gabez0r (commented):
I think this change should be done in https://github.com/nuxeo/nuxeo-web-ui/pull/1064/, since it assumes a line height that is only define there. LGTM, though.

nmpcunha:
> I think this change should be done in #1064, since it assumes a line height that is only define there. LGTM, though.

+1 on this.
It was kind of tricky to understand the context and infer the reason why.
Thanks for the clarification @semisse .

richardsd (approved):

semisse:
I will be closing this PR and merge the suggested solution in https://github.com/nuxeo/nuxeo-web-ui/pull/1064

PR #1074, opened 2020-11-17
WEBUI-62: Update default translation keys to match convention

mnixo:
To consider (open to discussion):
- Some of the [`videoViewLayout.` keys](https://github.com/nuxeo/nuxeo-web-ui/blob/d219dda5a6000cb390ac585262cd60e8bd296289/i18n/messages.json#L1243-L1254) were not changed (the ones referencing specific schema fields). The only usage I could find was in [`<nuxeo-video-view-layout>`](https://github.com/nuxeo/nuxeo-web-ui/blob/d219dda5a6000cb390ac585262cd60e8bd296289/elements/document/video/nuxeo-video-view-layout.html#L35), so we might be able to remove most of these entirely.
- Some of the [`pictureViewLayout.` keys](https://github.com/nuxeo/nuxeo-web-ui/blob/d219dda5a6000cb390ac585262cd60e8bd296289/i18n/messages.json#L902-L926) were not changed (the ones referencing specific schema fields). The usages I could find were in [`<nuxeo-picture-formats>`](https://github.com/nuxeo/nuxeo-web-ui/blob/d219dda5a6000cb390ac585262cd60e8bd296289/elements/document/nuxeo-picture-formats.js), [`<nuxeo-picture-document-page>`](https://github.com/nuxeo/nuxeo-web-ui/blob/d219dda5a6000cb390ac585262cd60e8bd296289/elements/document/nuxeo-picture-document-page.js) and [`<nuxeo-picture-info>`](https://github.com/nuxeo/nuxeo-web-ui/blob/d219dda5a6000cb390ac585262cd60e8bd296289/elements/document/nuxeo-picture-info.js). It seems that we could still remove some of these.
- The [`searchResults.highlight.field.` keys](https://github.com/nuxeo/nuxeo-web-ui/blob/d219dda5a6000cb390ac585262cd60e8bd296289/i18n/messages.json#L1106-L1111) were not changed. Even though they reference schema specific fields, they are used dynamically in [`<nuxeo-document-highlights>`](https://github.com/nuxeo/nuxeo-web-ui/blob/d219dda5a6000cb390ac585262cd60e8bd296289/elements/nuxeo-document-highlight/nuxeo-document-highlights.js).
- The [`searchResults.sort.field.` keys](https://github.com/nuxeo/nuxeo-web-ui/blob/d219dda5a6000cb390ac585262cd60e8bd296289/i18n/messages.json#L1114-L1122) were not changed. These keys reference dublincore fields and are used in the `attached` methods of most of the results layouts for the default page providers, as well as in the `_computeSortOptions` methods of [`<nuxeo-document-content-behavior>`](https://github.com/nuxeo/nuxeo-web-ui/blob/d219dda5a6000cb390ac585262cd60e8bd296289/elements/nuxeo-results/nuxeo-document-content-behavior.js) and [`<nuxeo-document-trash-content>`](https://github.com/nuxeo/nuxeo-web-ui/blob/d219dda5a6000cb390ac585262cd60e8bd296289/elements/nuxeo-results/nuxeo-document-trash-content.js).
- Web UI uses a different format for referencing predicate fields (that reference dublincore fields). For example, [`<nuxeo-default-search-form>` uses `ecm_tags`](https://github.com/nuxeo/nuxeo-web-ui/blob/d219dda5a6000cb390ac585262cd60e8bd296289/elements/search/default/nuxeo-default-search-form.html#L79), although configuring a similar predicate in Studio would result in `system_tag`.
- A `deprecated.json` file was added to map the keys that were replaced to the ones that replace them. In some instances, keys map to `null` which indicates that the keys were just removed (as they are not used anywhere else). This PR will not be merged until the deprecation of the keys is discussed. Alternatively, we can also handle the deprecation in a following ticket (as @Gabez0r suggested).


richardsd (dismissed):

richardsd (dismissed):

richardsd (dismissed):

richardsd (approved):

elements/nuxeo-document-info/nuxeo-document-info.js
```diff
@@ -86,15 +86,15 @@ Polymer({
       <nuxeo-date datetime="[[document.properties.dc:modified]]"></nuxeo-date>
     </div>
     <div class="item">
-      <label>[[i18n('documentInfo.created')]]</label>
+      <label>[[i18n('label.ui.schema.dc.created')]]</label>
       <nuxeo-date datetime="[[document.properties.dc:created]]"></nuxeo-date>
     </div>
     <div class="item">
       <label>[[i18n('documentInfo.by')]]</label>
```
Gabez0r, line 93: It looks inconsistent to change some labels and not others in this element. While this makes perfect sense on layouts, I'm worried about updating the elements (the .js files in this PR), because we either leave them or we remove the ability to customize the label just for this element (which might be something that we want actually, but it might break in some customization cases).

Gabez0r (commented):
I see a few cases where these might represent breaking changes:
1) people that have customized the old labels but use the default layouts, will lose the customizations (but who does this?)
2) those that have customized the old labels but have scaffolded new layouts will also lose the customizations too
3) for the changed js files (e.g, nuxeo-document-info), which from this point on will use new labels and customizations will be lost too

I am not sure how common any of these cases are, but the only way for these not to be breaking would be to either have fallback or have studio do some sort of automatic label migration.

richardsd:
Closing the PR as we won't be doing this.

PR #1076, opened 2020-11-18
Fix webui 116 fix ticket reference in commit message

RSalem07:


RSalem07:
No more needed as at the end we make the choice to update the JIRA ticket WEBUI-116 with the a link to the right commit

PR #1084, opened 2020-11-20
WEBUI-133: improve dev flow for addons

nelsonsilva:


nmpcunha (approved):

webpack.config.js
```diff
@@ -70,11 +71,17 @@ const layouts = [
...
+  .filter(Boolean)
+  .filter((p) => existsSync(`addons/${p}`));
+
+// Prepare copy of addon resources
+const addons = BUNDLES.map((p) => {
+  return {
+    from: `addons/${p}/**/*`,
     to: TARGET,
-    globOptions: { ignore: ['*.js'] },
+    globOptions: { ignore: ['*.js', '**/node_modules/**', 'package*.*'] },
```
Gabez0r, line 84: What do we what to ignore with `'package*.*'`?

Gabez0r (approved):

webpack.config.js
nelsonsilva: it's to avoid copying over package.json and package-lock.json. Tried setting the `gitignore` option which should in theory skip copying everything that was in the source folder's `.gitignore` but it was crashing when i tried it :(

PR #1099, opened 2020-11-27
WEBUI-96: allow promotion.yaml to accept ref branch and dry run as input 

Gabez0r:
Related https://github.com/nuxeo/nuxeo-elements/pull/322.

As a bonus, also creating a github release.

### DIFF
--- a/.github/workflows/promote.yaml
+++ b/.github/workflows/promote.yaml
@@ -6,6 +6,12 @@ on:
       version:
         description: 'Version to promote'     
         required: true
+      referenceBranch:
+        description: 'Branch to be updated after the promotion'
+        required: true
+        default: 'master'
+      dryRun:
+        description: 'Runs the workflow without pushing code or publishing artifacts. If empty, Dry Run is disabled.'
 
 jobs:
   promote:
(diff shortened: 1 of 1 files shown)


mnixo (dismissed):

Gabez0r:
This is how the workflow form looks like:
image

.github/workflows/promote.yaml
```diff
@@ -92,17 +117,30 @@ jobs:
           npm version $NEW_SNAPSHOT --no-git-tag-version
           popd
 
+      - if: ${{ github.event.inputs.referenceBranch == 'master' }}
+        run: |
+          NEW_ELEMENTS_VERSION="^$NEW_VERSION-rc"
+          npm i @nuxeo/nuxeo-elements@$NEW_ELEMENTS_VERSION --save-prefix="^"
```
nelsonsilva, line 123: the caret is the default prefix, right ?
also, if we're going with the caret i guess in practice we don't really have to do anything in the master branch, right ?

Gabez0r: WDYM by _we don't really have to do anything_? We need to install it anyway, in order to update the version. But perhaps we don't need the `--save-prefix="^"` part?

nelsonsilva: nevermind, forgot about the pre-release version we want to install

nelsonsilva (dismissed):

mnixo (approved):

richardsd (approved):
I'll trust you on this one 😬 
This is going to be available here in Github right?

Gabez0r:
Yes, on the Actions tab.

nelsonsilva (approved):

PR #1102, opened 2020-11-27
WEBUI-143: do not enable all addons by default

nelsonsilva:


webpack.config.js
```diff
@@ -75,21 +75,21 @@ const layouts = [
...
 
 if (BUNDLES.length) {
   log.info(`Bundling addons:\n\t-${BUNDLES.join('\n\t-')}`);
+} else {
+  log.info(`Bundling all addons`);
 }
 
 // Prepare copy of addon resources
-const addons = BUNDLES.map((p) => {
+const addons = ((BUNDLES.length && BUNDLES) || ALL_ADDONS).map((p) => {
```
Gabez0r, line 92: nit: would be smaller and easier to read with a ternary operator
```suggestion
const addons = (BUNDLES.length ? BUNDLES : ALL_ADDONS).map((p) => {
```

Gabez0r (approved):

webpack.config.js
nelsonsilva: 👍

Gabez0r (approved):

PR #1106, opened 2020-12-07
WEBUI-140: pad package prerelease with zeros

Gabez0r:


richardsd (dismissed):

Jenkinsfile
```diff
@@ -85,6 +85,12 @@ def getVersion() {
   return isPullRequest() ? getPullRequestVersion() : getReleaseVersion()
 }
 
+def padPrerelease(version) {
+  def count = sh(returnStdout: true, script: "echo ${version} | sed -r s/[0-9]+\\.[0-9]+\\.[0-9]+-rc\\.\\([0-9]+\\)/\\\\1/g")
+  def padded = sh(returnStdout: true, script: "printf '%03d' ${count}")
+  return sh(script: "echo ${version} | sed -E 's/([0-9]+\\.[0-9]+\\.[0-9]+-rc\\.)[0-9]+/\\1${padded}/g'", returnStdout: true).trim()
+}
```
mnixo, line 92: Where does it start padding with 2 zeroes once you get to prerelease 10 and 1 zero once you get to prerelease 100?

Gabez0r: @mnixo, this is done by the printf command in `def padded = sh(returnStdout: true, script: "printf '%03d' ${count}")`. This is how it works:
```
$ printf '%03d' 5
005
$ printf '%03d' 50
050
$ printf '%03d' 500
500
$ printf '%03d' 5000
5000
```

mnixo (dismissed):

.github/workflows/promote.yaml
```diff
@@ -42,6 +42,7 @@ jobs:
           popd
 
           find . -type f -not -path "./node_modules/*" -regex ".*\.\(yaml\|sample\|xml\)" -exec sed -i 's/'"$OLD_VERSION"'/'"$VERSION"'/g' {} \;
+          sed -i -e 's/-rc\.[0-9]+//g' plugin/web-ui/marketplace/src/main/assemble/assembly.xml
```
nelsonsilva, line 45: i don't see any hardcoded versions in `assembly.xml` what are we replacing here ? 🤔

Jenkinsfile
```diff
@@ -85,6 +85,12 @@ def getVersion() {
   return isPullRequest() ? getPullRequestVersion() : getReleaseVersion()
 }
 
+def padPrerelease(version) {
+  def count = sh(returnStdout: true, script: "echo ${version} | sed -r s/[0-9]+\\.[0-9]+\\.[0-9]+-rc\\.\\([0-9]+\\)/\\\\1/g")
```
nelsonsilva, line 89: wouldn't these logics be easier to follow if done in Groovy (ie `rc.padLeft(3, '0')`) ?

.github/workflows/promote.yaml
Gabez0r: They are being committed [here](https://github.com/nuxeo/nuxeo-web-ui/pull/1106/files#diff-e6ffa5dc854b843b3ee3c3c28f8eae2f436c2df2b1ca299cca1fa5982e390cf8R118). But perhaps we don't have to commit them? However, if we don't, then building the marketplace locally will result in a package with a different version than the one that was pushed.

Jenkinsfile
Gabez0r: Didn't know about the groovy padLeft. This could could probably be useful for the GHActions, no?

Gabez0r: tried with `def padded = count.padLeft(3, '0')`, but no padding occurred. Keeping it as is.

richardsd (dismissed):

mnixo (approved):

nelsonsilva (approved):

mnixo (approved):

richardsd (approved):

PR #1107, opened 2020-12-08
WEBUI-189: handle a password on easyshare

s-guillaume:
Features :
- Secure Easyshare shared folder with a password from Web-UI available on creation or edit layouts

- On a secured Easyshare shared folder, when a « public » end-user is accessing a secured Easyshare shared folder, it will be prompt the accessCode

- If the « public » end-user entered the defined accessCode, he will have access to the shared folder content from the current browser. 
=> Some extra informations will be added the your browser URL
=> These extra informations are mandatory once connected and cannot be removed or shared to other « public » end-users (these informations are dynamically changed on each call with a matching based on browser JSESSIONID so using another browser will imply to re-enter accessCode or share it, notice it’s not working with Incognito session on Browsers)

- AccessCode is controled on each call that means if you change your accessCode in the EasyShare folder, previously connected users will be disconnected and prompted a new AccessCode

1.	Core/WebUI update

Two new metadatas as accessCodeRequired (Boolean) and accessCode (String) value added to existing easyshare schema (updated XSD)

WebUI Layouts (creation /edit) and JSF layout (creation/edit)
- Checkbox for accessCodeRequired metadata to activate password requirement on public EasyShare access
- Input text for accessCode to provide a password

Translation available in English/US, French, German and Spanish to check

2. Easyshare "public" access is not changing with no password

If EasyShare folder has no password set, access to public Easyshare remains available (no access code required so public access will remain as current version of EasyShare).

3. Capacity of checking checkbox "Access code required ?" and setting password to secure access to an Easyshare collection

If checkbox is checked, password is mandatory

4. Easyshare "public" access will ask for a password (access code) if the collection has password protected access (checkbox checked).
In that case, the user is prompted to enter the password. 

5. If the right password is entered, the user will be able to navigate in the collection. Otherwise, he must re-enter the password.
Notice all links are available for end-users include 3 new dynamic url parameters (a, s and j). They are calculated on the fly, encrypted with default and generated salt and only available on the current browser.

### DIFF
--- a/addons/easyshare/document/easysharefolder/nuxeo-easysharefolder-create-layout.html
+++ b/addons/easyshare/document/easysharefolder/nuxeo-easysharefolder-create-layout.html
@@ -49,7 +49,10 @@ limitations under the License.
     >
     </nuxeo-date-picker>
 
-    <paper-checkbox noink name="notification" checked$="{{document.properties.eshare:hasNotification}}">
+    <paper-checkbox
+      noink
+      name="notification"
+      checked$="{{document.properties.eshare:hasNotification}}">
       [[i18n('easysharefolder.notification')]]
     </paper-checkbox>
 
@@ -61,6 +64,24 @@ limitations under the License.
       value="{{document.properties.eshare:contactEmail}}"
     >
     </nuxeo-input>
+
+    <paper-checkbox
+      noink
+      name="accessCodeRequired"
+      checked="{{document.properties.eshare:accessCodeRequired}}"
+      id="accessCodeRequired">
+      [[i18n('easysharefolder.accessCodeRequired')]]
+    </paper-checkbox>
+
+    <nuxeo-input
+      role="widget"
+      label="[[i18n('easysharefolder.accessCode')]]"
+      id="accessCode"
+      name="accessCode"
+      value="{{document.properties.eshare:accessCode}}"
+      required="[[document.properties.eshare:accessCodeRequired]]">
+    </nuxeo-input>
+
   </template>
   <script>
     Polymer({
--- a/addons/easyshare/document/easysharefolder/nuxeo-easysharefolder-edit-layout.html
+++ b/addons/easyshare/document/easysharefolder/nuxeo-easysharefolder-edit-layout.html
@@ -49,7 +49,10 @@ limitations under the License.
     >
     </nuxeo-date-picker>
 
-    <paper-checkbox noink name="notification" checked="{{document.properties.eshare:hasNotification}}">
+    <paper-checkbox
+      noink
+      name="notification"
+      checked="{{document.properties.eshare:hasNotification}}">
       [[i18n('easysharefolder.notification')]]
     </paper-checkbox>
 
@@ -61,6 +64,24 @@ limitations under the License.
       value="{{document.properties.eshare:contactEmail}}"
     >
     </nuxeo-input>
+
+    <paper-checkbox
+      noink
+      name="accessCodeRequired"
+      checked="{{document.properties.eshare:accessCodeRequired}}"
+      id="accessCodeRequired">
+      [[i18n('easysharefolder.accessCodeRequired')]]
+    </paper-checkbox>
+
+    <nuxeo-input
+      role="widget"
+      label="[[i18n('easysharefolder.accessCode')]]"
+      id="accessCode"
+      name="accessCode"
+      value="{{document.properties.eshare:accessCode}}"
+      required="{{document.properties.eshare:accessCodeRequired}}">
+    </nuxeo-input>
+
   </template>
   <script>
     Polymer({
--- a/addons/easyshare/document/easysharefolder/nuxeo-easysharefolder-metadata-layout.html
+++ b/addons/easyshare/document/easysharefolder/nuxeo-easysharefolder-metadata-layout.html
@@ -71,6 +71,18 @@ limitations under the License.
       <label>[[i18n('easysharefolder.contactEmail')]]</label>
       <div name="email">[[document.properties.eshare:contactEmail]]</div>
     </div>
+
+   <div role="widget" hidden$="[[!document.properties.eshare:accessCodeRequired]]">
+      <paper-checkbox name="accessCodeRequired" checked$="[[document.properties.eshare:accessCodeRequired]]"
+        noink disabled>
+        [[i18n('easysharefolder.accessCodeRequired')]]
+      </paper-checkbox>
+    </div>
+
+    <div role="widget" hidden$="[[!document.properties.eshare:accessCode]]">
+      <label>[[i18n('easysharefolder.accessCode')]]</label>
+      <div name="accessCode">[[document.properties.eshare:accessCode]]</div>
+    </div>    
   </template>
   <script>
     Polymer({
--- a/i18n/messages.json
+++ b/i18n/messages.json
@@ -533,6 +533,8 @@
   "easyshare.copy.label": "Internal Access link to {0}.",
   "easysharefolder.notification": "Active Notification",
   "easysharefolder.contactEmail": "Email",
+  "easysharefolder.accessCodeRequired": "Access code required ?",
+  "easysharefolder.accessCode": "Access code",
   "easysharefolder.share": "Public EasyShare link to {0}",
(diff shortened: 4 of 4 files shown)


i18n/messages-de.json
```diff
@@ -533,6 +533,8 @@
   "easyshare.copy.label": "Kopieren Sie den Link und teilen Sie Dokument {0}.",
   "easysharefolder.notification": "Aktive Benachrichtigung",
   "easysharefolder.contactEmail": "E-Mail",
+  "easysharefolder.accessCodeRequired": "Zugangscode erforderlich?",
```
Gabez0r, line 536: On master, you only need to add labels to `messages.json`. All other messages files will be overridden once labels are synched with Crowdin.

addons/easyshare/document/easysharefolder/nuxeo-easysharefolder-create-layout.html
```diff
@@ -70,8 +87,27 @@
         /**
          * @doctype Collection
          */
-        document: Object,
+        document: Object
```
Gabez0r, line 90: The dangling comma was removed, which will cause linting to fail.

Gabez0r (changes requested):
Commits to the `nuxeo-web-ui` repo require a WEBUI Jira ticket. Please create one and amend the commit message.

addons/easyshare/document/easysharefolder/nuxeo-easysharefolder-create-layout.html
```diff
@@ -61,6 +61,23 @@
...
+      [[i18n('easysharefolder.accessCodeRequired')]]
+    </paper-checkbox>
+
+    <nuxeo-input
+      role="widget"
+      label="[[i18n('easysharefolder.accessCode')]]"
+      id="accessCode"
+      name="accessCode"
+      value="{{document.properties.eshare:accessCode}}"
+      on-value-changed="checkAccessCode"
```
Gabez0r, line 77: We should be able to achieve similar behavior without needing `checkAccessCodeRequired` and `checkAccessCode`, by changing:
```suggestion
      required="[[document.properties.eshare:accessCodeRequired]]"
```

addons/easyshare/document/easysharefolder/nuxeo-easysharefolder-edit-layout.html
```diff
@@ -61,6 +61,23 @@
...
+      [[i18n('easysharefolder.accessCodeRequired')]]
+    </paper-checkbox>
+
+    <nuxeo-input
+      role="widget"
+      label="[[i18n('easysharefolder.accessCode')]]"
+      id="accessCode"
+      name="accessCode"
+      value="{{document.properties.eshare:accessCode}}"
+      on-value-changed="checkAccessCode"
```
Gabez0r, line 77: Same here.

Gabez0r:
Depends on https://github.com/nuxeo/nuxeo/pull/4520.

addons/easyshare/document/easysharefolder/nuxeo-easysharefolder-create-layout.html
```diff
@@ -61,6 +61,23 @@
       value="{{document.properties.eshare:contactEmail}}"
     >
     </nuxeo-input>
+
+    <paper-checkbox noink name="accessCodeRequired"
+      checked="{{document.properties.eshare:accessCodeRequired}}"
+      id="accessCodeRequired" on-tap="checkAccessCodeRequired">
+      [[i18n('easysharefolder.accessCodeRequired')]]
```
andreacornaglia, line 68: can we format this as one property per line?

addons/easyshare/document/easysharefolder/nuxeo-easysharefolder-create-layout.html
```diff
@@ -70,8 +87,27 @@
...
+      },
+      checkAccessCodeRequired: function(event) {
+        // Set the field in error state to alert user
+        if (!this.$.accessCode.value)
+        {
+          this.$.accessCode.invalid = event.target.checked;
+        }
+        this.$.accessCode.required = event.target.checked;
       },
+      checkAccessCode: function(event) {
```
andreacornaglia, line 100: seems like we're not using the event argument passed to the function, should we remove it?

addons/easyshare/document/easysharefolder/nuxeo-easysharefolder-edit-layout.html
```diff
@@ -70,8 +87,27 @@
...
+      },
+      checkAccessCodeRequired: function(event) {
+        // Set the field in error state to alert user
+        if (!this.$.accessCode.value)
+        {
+          this.$.accessCode.invalid = event.target.checked;
+        }
+        this.$.accessCode.required = event.target.checked;
       },
+      checkAccessCode: function(event) {
```
andreacornaglia, line 100: here too seems like we're not using event.

andreacornaglia:
@s-guillaume thanks for your work! @Gabez0r and I left some comments. Let us know if it makes sense or you need further info.

addons/easyshare/document/easysharefolder/nuxeo-easysharefolder-create-layout.html
s-guillaume: On existing widget 10 lines above (for paper-checkbox name="notification" ...) do I need to apply the same (one property per line) as this widget was previously done by Nuxeo ?

s-guillaume: > can we format this as one property per line?

On existing widget 10 lines above (for paper-checkbox name="notification" ...) do I need to apply the same (one property per line) as this widget was previously done by Nuxeo ?

addons/easyshare/document/easysharefolder/nuxeo-easysharefolder-create-layout.html
s-guillaume: OK, tested as expected but regarding checkAccessCodeRequired method launched on-tap, the accessCode input is also changing untapping the checkbox (passing from red to blue and removing the required red star). Do I also need to remove it ?

addons/easyshare/document/easysharefolder/nuxeo-easysharefolder-create-layout.html
s-guillaume: You're right, done and tested.

i18n/messages-de.json
s-guillaume: 2 entries (easysharefolder.accessCodeRequired and easysharefolder.accessCode) removed from fr, de and es JSON files

Gabez0r: Yes, that's it. Thank you.

addons/easyshare/document/easysharefolder/nuxeo-easysharefolder-create-layout.html
andreacornaglia: @s-guillaume no, just formatting the code you modify is fine :)

andreacornaglia:
@s-guillaume thank you for considering our comments. The code LGTM but I would modify the commit message to state what the code is doing instead. ("add the option of password protection via easyshare" or something along those lines). Thanks!

addons/easyshare/document/easysharefolder/nuxeo-easysharefolder-create-layout.html
```diff
@@ -72,6 +94,10 @@
          */
         document: Object,
       },
+      checkAccessCodeRequired: function(event) {
+        // Set the accessCode field required or not
+        this.$.accessCode.required = event.target.checked;
```
Gabez0r, line 99: @s-guillaume: why do we need this? I would expect the `required="[[document.properties.eshare:accessCodeRequired]]"` to already have this behavior.

s-guillaume: Removed, you're right

s-guillaume:
> Commits to the `nuxeo-web-ui` repo require a WEBUI Jira ticket. Please use [WEBUI-189](https://jira.nuxeo.com/browse/WEBUI-189) instead and amend the commit message.

Done

PR #1109, opened 2020-12-10
Task web UI 139 labels 3.0.x

Gabez0r:
Cherry pick missing translations and missing commits.


themes/base.js
```diff
@@ -539,6 +539,11 @@ const template = html`
           transition: none;
         }
 
+        --paper-font-body1: {
```
richardsd, line 542: Is this change supposed to be backported as well?

richardsd (approved):

mnixo (approved):

themes/base.js
Gabez0r: I think so. I think https://jira.nuxeo.com/browse/WEBUI-144 should have been backported but it was forgotten.

richardsd: 👍 Good. It was just to confirm that it was supposed. Thanks.

PR #1118, opened 2020-12-15
WEBUI-156: replace ID by label in edit vocabulary layout

semisse:


elements/directory/xvocabulary/nuxeo-xvocabulary-edit-layout.html
```diff
@@ -38,7 +38,7 @@
       required$="[[_isParentRequired(parentDirectory, directory)]]"
     >
       <template is="dom-repeat" items="[[parentResponse.entries]]">
-        <paper-item name="[[item.properties.id]]">[[item.properties.id]]</paper-item>
+        <paper-item name="[[item.properties.label]]">[[item.properties.label]]</paper-item>
```
Gabez0r, line 41: I think changing the name to the label instead an id will introduce a bug, causing the new parent id to actually be a label. Can you please confirm?

semisse: yes, you are right. we need to keep the `id` on the `name` property

mnixo (approved):

Gabez0r (approved):

Gabez0r (commented):
image

Do we want to have the labels translated here?

richardsd (approved):

richardsd (dismissed):
Maybe you could separate this into two commits, wdyt? One with the actual update of the `nuxeo-xvocabulary-edit-layout` and the other one with the messages.json update.

elements/directory/xvocabulary/nuxeo-xvocabulary-edit-layout.html
```diff
@@ -119,6 +119,12 @@
       _isParentRequired() {
         return this.directory && this.parentDirectory && this.directory !== this.parentDirectory;
       },
+
+      _computeLabel(item) {
+        return item && item.properties && item.properties.label
```
richardsd, line 124: Looking at https://github.com/nuxeo/nuxeo-ui-elements/blob/master/nuxeo-i18n-behavior.js#L46 it looks like we use the key as fallback if there's no translation, right? Wdyt?

richardsd (approved):

elements/directory/xvocabulary/nuxeo-xvocabulary-edit-layout.html
semisse: You're right, we don't need the fallback here. Thanks!

richardsd (dismissed):

mnixo (dismissed):

mnixo (approved):

elements/directory/xvocabulary/nuxeo-xvocabulary-edit-layout.html
```diff
@@ -38,7 +38,7 @@
       required$="[[_isParentRequired(parentDirectory, directory)]]"
     >
       <template is="dom-repeat" items="[[parentResponse.entries]]">
-        <paper-item name="[[item.properties.id]]">[[item.properties.id]]</paper-item>
+        <paper-item name="[[item.properties.id]]">[[_computeLabel(item)]]</paper-item>
```
Gabez0r, line 41: `_computeLabel` as a dependency on `this.i18n`, which in theory can change, so we would like the label to recompute if this happens:
```suggestion
        <paper-item name="[[item.properties.id]]">[[_computeLabel(item, i18n)]]</paper-item>
```

Gabez0r (commented):
On the master branch, we don't need to add any messages file other than `messages.json`, since they will be overridden anyway when the sync occurs. We should add them for 3.0.x and 10.10, though.

richardsd (approved):

mnixo (approved):

PR #1121, opened 2020-12-22
Fix promotion script

nelsonsilva:


richardsd (approved):

semisse (approved):

.github/workflows/promote.yaml
```diff
@@ -117,6 +117,8 @@ jobs:
           popd
 
       - if: ${{ github.event.inputs.referenceBranch == 'master' }}
+        env:
```
richardsd, line 120: For future reference, we need to set the token in every step we need it 🤯

richardsd (approved):

PR #1125, opened 2021-01-04
NXBT-3461: do manual helm init before building the preview

nelsonsilva:


Gabez0r (approved):
LGTM. We also need to do this on [nuxeo-elements](https://github.com/nuxeo/nuxeo-elements/blob/master/storybook/charts/preview/Makefile#L18), right?

semisse (dismissed):

PR #1137, opened 2021-01-19
Proposed fix for ELEMENTS-1300

mnixo:


gulpfile.js
```diff
@@ -98,7 +98,7 @@ gulp.task('polymer-build', function() {
       .pipe($.if(/\.js$/, babel({
         presets: babelPresetES2015,
         compact: false,
-        ignore: 'custom-elements-es5-adapter.js,webcomponents-*.js'})))
+        ignore: 'custom-elements-es5-adapter.js,webcomponents-*.js,pdf.worker.js'})))
```
richardsd, line 101: Can you explain a bit about the rationale here? Is there anything happening during the babel phase that leads us to exclude the pdf worker? 🤔

Gabez0r, line 101: I'm ok with this. This will probably prevent this from working on IE11, but [we dropped support for it](https://www.nuxeo.com/blog/nuxeo-web-ui-ie11/) on LTS2017 (9.10) a long time ago.

Gabez0r (approved):

nelsonsilva (approved):

mnixo:
Opening a non-draft PR.

PR #1140, opened 2021-01-25
WEBUI-172/WEBUI-173: RTE Image Picker

mnixo:
Pairs with [nuxeo-elements/pull/360](https://github.com/nuxeo/nuxeo-elements/pull/360).


richardsd (dismissed):

elements/elements.js
```diff
@@ -173,7 +173,6 @@ import './directory/nuxeo-vocabulary-management.js';
 
 // Search layout elements -->
 import './search/nuxeo-saved-search-actions.js';
-import './search/nuxeo-search-results-layout.js';
```
Gabez0r, line 176: Not sure we can remove this. Isn't this required for the element to be made available on Designer?

mnixo: Ah yes, I think you're right. I wonder why `nuxeo-search-form-layout` isn't here, though... 🤔

Gabez0r: I guess someone forgot (most likely me 😬 ✌️ ).

Gabez0r (approved):

semisse (dismissed):

elements/elements.js
richardsd: I don't think that we actually use the `nuxeo-search-results-layout` or the `nuxeo-search-form-layout` anywhere in Designer (at least directly). But it doesn't hurt to expose it on the catalog, since it might be needed for some obscure reason 😄

richardsd (dismissed):

i18n/messages-ar.json
```diff
@@ -460,7 +460,6 @@
   "documentPage.to.process": "You have a task to process before ",
   "documentPrevNextDoc.nextDocument": "Next Document",
   "documentPrevNextDoc.previousDocument": "Previous Document",
-  "documentSearchForm.layoutNotFound": "Failed to find search layout for {0}.",
```
Gabez0r, line 463: No need to edit any translation file other then `messages.json` on the master branch, as they will be overridden when the sync with Crowdin runs.

Gabez0r (dismissed):

richardsd (dismissed):

elements/search/picker/nuxeo-picker-search-form.html
```diff
@@ -0,0 +1,48 @@
...
+@element nuxeo-picker-search-form
+-->
+<dom-module id="nuxeo-picker-search-form">
+  <template>
+    <style include="nuxeo-styles"></style>
+    <nuxeo-input
+      role="widget"
+      id="searchInput"
+      value="{{params.title}}"
+      label="Quick Search"
```
Gabez0r, line 13: This should internationalized.

elements/search/picker/nuxeo-picker-search-results.html
```diff
@@ -0,0 +1,46 @@
+<!--
+`nuxeo-picker-search-results`
+@group Nuxeo UI
+@element nuxeo-picker-search-results
+-->
+<dom-module id="nuxeo-picker-search-results">
+  <template>
+    <style include="nuxeo-styles">
+      nuxeo-data-grid {
+        min-height: 200px;
```
Gabez0r, line 10: Without this, things break, right?

mnixo: Yup. Without setting this explicitly, `min-height: 500px;` is used ([set on `<nuxeo-data-grid>`](https://github.com/nuxeo/nuxeo-elements/blob/802d59518211da4681d0c85d38773fb3f8946819/ui/nuxeo-data-grid/nuxeo-data-grid.js#L77)), which is way too much for the dialog sizing restrictions that we have.

semisse (dismissed):

semisse (approved):

elements/search/picker/nuxeo-picker-search-form.html
```diff
@@ -0,0 +1,43 @@
...
+  <template>
+    <style include="nuxeo-styles"></style>
+    <nuxeo-input
+      role="widget"
+      id="searchInput"
+      value="{{params.title}}"
+      label="[[i18n('pickerSearch.title')]]"
+      type="text"
+      autofocus
+      on-keydown="_checkForEnter"
```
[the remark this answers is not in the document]
mnixo: I agree with that. But in this case I just copied the same behavior that we have in the other default results layouts ([`assets`](https://github.com/nuxeo/nuxeo-web-ui/blob/f3f784dbfff2180c424a2b58effcc01928ddba65/elements/search/assets/nuxeo-assets-search-form.html#L154), [`default`](https://github.com/nuxeo/nuxeo-web-ui/blob/f3f784dbfff2180c424a2b58effcc01928ddba65/elements/search/default/nuxeo-default-search-form.html#L171), [`expired`](https://github.com/nuxeo/nuxeo-web-ui/blob/f3f784dbfff2180c424a2b58effcc01928ddba65/elements/search/expired/nuxeo-expired-search-form.html#L93) and [`nxql`](https://github.com/nuxeo/nuxeo-web-ui/blob/f3f784dbfff2180c424a2b58effcc01928ddba65/elements/search/nxql/nuxeo-nxql-search-form.html#L67)).
If we decide that it's a good opportunity to use `_onKeydown` we can go with that, if not I'll keep this one for consistency with the others.

richardsd (dismissed):

elements/search/document_picker/nuxeo-document_picker-search-form.html
```diff
@@ -0,0 +1,43 @@
+<!--
+`nuxeo-document_picker-search-form`
+@group Nuxeo UI
+@element nuxeo-document_picker-search-form
+-->
+<dom-module id="nuxeo-document_picker-search-form">
+  <template>
+    <style include="nuxeo-styles"></style>
+    <nuxeo-input
```
richardsd, line 9: Since this is a new element, we might want to add here the `aria-label` too. What do you think @semisse? Should we do it for the search form and search results?

richardsd (dismissed):

elements/search/document_picker/nuxeo-document_picker-search-form.html
semisse: Well noted, Ricardo. In this case we should add:
```suggestion
    on-keydown="_checkForEnter"
    aria-label$="[[i18n('pickerSearch.title')]]"
    ></nuxeo-input>
```

richardsd (approved):

semisse (approved):

Gabez0r (approved):

Gabez0r (approved):

richardsd (approved):

semisse (approved):

PR #1141, opened 2021-01-26
WEBUI-188: Migrate to GH Actions

nelsonsilva:


.github/workflows/main.yaml
```diff
@@ -0,0 +1,136 @@
...
+      with:
+        java-version: 11
+
+    - name: Cache local Maven repository
+      uses: actions/cache@v2
+      with:
+        path: ~/.m2/repository
+        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+        restore-keys: |
+          ${{ runner.os }}-maven-
```
Gabez0r, line 30: Is this used between runs, right?

.github/workflows/main.yaml
```diff
@@ -0,0 +1,136 @@
...
+    - name: Archive packages
+      uses: actions/upload-artifact@v2
+      with:
+        name: packages
+        path: |
+          plugin/web-ui/marketplace/target/nuxeo-web-ui-marketplace-*.zip
+          plugin/itests/marketplace/target/nuxeo-web-ui-marketplace-itests-*.zip
+
+    - name: Tag
+      if: github.event_name == 'push'
```
Gabez0r, line 117: If by any chance the worflow fails and we rerun it, will the `event_name` be the same (i.e., `push`)?

.github/workflows/main.yaml
```diff
@@ -0,0 +1,136 @@
+name: Build
+
+on:
+  push:
+    branches:
+      - master
+      - maintenance-3.0.x
+  pull_request:
```
Gabez0r, line 8: Don we need to specify types and branches for the `pull_request`? Will it use the same from `push`?

.github/workflows/main.yaml
nelsonsilva: Yep. TBH this is "borrowed" from their usage guides https://github.com/actions/cache/blob/main/examples.md#java---maven

.github/workflows/main.yaml
nelsonsilva: Good question 🤔  docs say "Re-running a workflow uses the same GITHUB_SHA (commit SHA) and GITHUB_REF (Git ref) of the original event that triggered the workflow run." but don't mention anything about the event_name

.github/workflows/main.yaml
nelsonsilva: The workflow will only kick in if it is defined in the current branch so depending on which branch you start from it'll run or not.
For the push, since we don't want to run the workflow when you push to feature branches we have to keep an explicit list.

mnixo (dismissed):

.github/workflows/main.yaml
Gabez0r: I guess we will find out? 😛

charts/preview/templates/_helpers.tpl
```diff
@@ -0,0 +1,20 @@
...
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+*/}}
+{{- define "nuxeo.webui.fullname" -}}
+{{- if .Values.fullnameOverride -}}
+{{- printf .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{- $name := default .Chart.Name .Values.nameOverride -}}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+{{- end -}}
```
Gabez0r, line 20: nit: missing newline?

charts/preview/values.yaml
```diff
@@ -29,17 +30,19 @@ nuxeo:
...
-          fabric8.io/expose: "true"
-          fabric8.io/ingress.path: /
-          fabric8.io/ingress.name: preview
+  webui:
+    fullname: preview
+    image:
+      repository: $DOCKER_REGISTRY/$DOCKER_IMAGE
+      tag: $VERSION
+      pullPolicy: Always
+    packages: nuxeo-dam
```
Gabez0r, line 42: Don't think we need this, since nuxeo-dam is no more?

.github/workflows/clean.yaml
```diff
@@ -0,0 +1,37 @@
+name: Cleanup
+
+on:
+  pull_request:
+    types: [closed, unlabeled]
+    branches:
+      - master
```
richardsd, line 7: To be removed.

richardsd:
Closing this PR and opened https://github.com/nuxeo/nuxeo-web-ui/pull/1307 for maintenance-3.0.x

PR #1155, opened 2021-02-10
WEBUI-232: set parent to root when none is provided

richardsd:


richardsd:
Pipeline running here https://qa.nuxeo.org/jenkins/job/web-ui-pipeline-rdias/52/

richardsd:
vocabulary_management_no_parent

elements/directory/l10nxvocabulary/nuxeo-l10nxvocabulary-edit-layout.html
```diff
@@ -168,6 +168,10 @@ <h2>[[i18n('vocabularyManagement.selectParent')]]</h2>
 
       _entryChanged() {
         this.entry.properties.obsolete = this.entry.properties.obsolete ? 1 : 0;
+        const { parent } = this.entry.properties;
+        // if the entry doesn't have a parent, then we set it has the root
+        this.entry.properties.parent =
+          typeof parent === 'undefined' || parent === null ? this._root.properties.id : parent;
```
andreacornaglia, line 174: can't we just do parent ?  this._root.properties.id : parent;

elements/directory/l10nxvocabulary/nuxeo-l10nxvocabulary-edit-layout.html
```diff
@@ -168,6 +168,10 @@ <h2>[[i18n('vocabularyManagement.selectParent')]]</h2>
 
       _entryChanged() {
         this.entry.properties.obsolete = this.entry.properties.obsolete ? 1 : 0;
+        const { parent } = this.entry.properties;
```
andreacornaglia, line 171: also suggest to do this on the first line, and can also use it for obsolete.

andreacornaglia (commented):
I am confused by the commit message since it says to set parent to '' but in the code we set it to the this._root.properties.id

elements/directory/l10nxvocabulary/nuxeo-l10nxvocabulary-edit-layout.html
richardsd: Actually, you would have to do `!parent ? this._root.properties.id : parent`, but it doesn't work if the `parent` is an empty string. For instance, try to run this block here:

```
const parents = [null, undefined, '', 'bla'];

parents.forEach(parent => {
	const result1 = typeof parent === 'undefined' || parent === null ? 'newParent' : parent;
  const result2 = !parent ? 'newParent' : parent;
  console.log(`Result for parent: ${parent} (${result1}, ${result2})`);
});
```

elements/directory/l10nxvocabulary/nuxeo-l10nxvocabulary-edit-layout.html
richardsd: I can update that one as well. Thanks

richardsd:
> I am confused by the commit message since it says to set parent to '' but in the code we set it to the this._root.properties.id

You're right. It is a poor commit message. I used an empty string in the first place, but then realized we should be using the root (which in fact is an empty string, but that is just a detail). I'll reword the ticket.

Gabez0r (approved):

elements/directory/l10nxvocabulary/nuxeo-l10nxvocabulary-edit-layout.html
```diff
@@ -167,7 +167,10 @@ <h2>[[i18n('vocabularyManagement.selectParent')]]</h2>
       observers: ['_inputChanged(entry, entries)'],
 
       _entryChanged() {
-        this.entry.properties.obsolete = this.entry.properties.obsolete ? 1 : 0;
+        const { obsolete, parent } = this.entry.properties;
+        this.entry.properties.obsolete = obsolete ? 1 : 0;
+        // if the entry doesn't have a parent, then we set it has the root
```
mnixo, line 172: as?
```suggestion
        // if the entry doesn't have a parent, then we set it as the root
```

mnixo (dismissed):

mnixo (approved):

Gabez0r (approved):

PR #1164, opened 2021-02-23
WEBUI-257: introduce default language resolver

nelsonsilva:
Couple of things to discuss here before merging:

- [ ] Usage of `||=` and `?.``
- [ ] Reloading i18n when user changes (and doing first load without user to avoid displaying label keys in the meantime)

### DIFF
--- a/elements/nuxeo-app.js
+++ b/elements/nuxeo-app.js
@@ -61,7 +61,6 @@ import './nuxeo-browser/nuxeo-breadcrumb.js';
 import './nuxeo-browser/nuxeo-repositories.js';
 import './nuxeo-document-storage/nuxeo-document-storage.js';
 import './nuxeo-results/nuxeo-results.js';
-import '../i18n/i18n.js';
 import '../themes/base.js';
 import '../themes/loader.js';
 import './nuxeo-search-page.js';
@@ -76,6 +75,7 @@ import { dom } from '@polymer/polymer/lib/legacy/polymer.dom.js';
 import { PolymerElement } from '@polymer/polymer/polymer-element.js';
 import { afterNextRender } from '@polymer/polymer/lib/utils/render-status.js';
 import { importHref } from '@nuxeo/nuxeo-ui-elements/import-href.js';
+import { loadI18n } from '../i18n/i18n.js';
 
 import { Performance } from './performance.js';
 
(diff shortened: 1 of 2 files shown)


elements/nuxeo-app.js
```diff
@@ -1036,13 +1038,14 @@ Polymer({
     this.fire('document-updated');
   },
 
-  _observeCurrentUser() {
-    if (this.currentUser) {
+  _observeCurrentUser(user) {
+    loadI18n(user);
```
Gabez0r, line 1042: We will be calling loadI18n twice, right? On ready, and then when the user changes. But this will do a second fetch which we might not need if the language did not change. Perhaps we should improve this?

i18n/i18n.js
```diff
@@ -18,9 +18,20 @@ import { XHRLocaleResolver } from '@nuxeo/nuxeo-ui-elements/nuxeo-i18n-behavior.
...
-  /* Set html lang attribute. Required by the better-dateinput element */
-  document.getElementsByTagName('html')[0].lang = window.nuxeo.I18n.language;
-});
+
+window.nuxeo.I18n.languageResolver ||= () =>
+  document.getElementsByTagName('html')[0].lang || navigator.language || navigator.userLanguage || 'en';
+
+window.nuxeo.I18n.localeResolver ||= new XHRLocaleResolver(msgFolder);
+
+const loadI18n = (user) => {
```
Gabez0r, line 27: Should we use a generic `opts` as a parameter instead, that today would contain opts.user, but tomorrow would allows us to expose more stuff?

Gabez0r (commented):
I'm ok with the ?. and ||= as long as we don't have to backport to 10.10, as we might still have to keep it working with edge legacy (?)

elements/nuxeo-app.js
nelsonsilva: yep, I know, but in the userprofile enricher is currently retrieved by the JS client and it follows a couple of earlier requests to actually login which means we'd have to delay loading the UI, or at least the messages, until this requests is fulfilled. Main use case (and only promoted one for now) would be setting the `lang` attribute in `<html>` for now in which case the default language provider would load the proper messages in the first place.

i18n/i18n.js
nelsonsilva: we can expose more stuff as more arguments, but I get the point about named arguments being nicer to work with. I can just do destructuring here to make it nicer `({ user }) => ...` but it's not really something we'd been doing in other places

PR #1168, opened 2021-03-01
WEBUI-132: enable remove from collection button when user has WriteProperties

richardsd:
JSF UI is relying in the platform CollectionManager to assert if the user [can add/remove](https://github.com/nuxeo/nuxeo/blob/master/modules/platform/nuxeo-collections/src/main/java/org/nuxeo/ecm/collections/core/CollectionManagerImpl.java#L132) items from a collection. This checks if the user has the WriteProperties permission on the collection, and on Web UI we are only do the assertion for the [ReadCanCollect](https://github.com/nuxeo/nuxeo/blob/8ce732309dc35ebe6c53aa8e0dd52757e8faf74a/modules/platform/nuxeo-collections/src/main/resources/OSGI-INF/collection-security-contrib.xml#L9) permission (that includes both the Read and WriteProperties permissions). 
To make this consistent, we should check for the `WriteProperties` permission too.

### DIFF
--- a/elements/nuxeo-collections/nuxeo-collection-remove-action.js
+++ b/elements/nuxeo-collections/nuxeo-collection-remove-action.js
@@ -91,7 +91,7 @@ Polymer({
     if (collection && collection.contextParameters && collection.contextParameters.permissions) {
       // NXP-21408: prior to 8.10-HF01 the permissions enricher wouldn't return ReadCanCollect
       // Action will therefore not be available
-      return collection.contextParameters.permissions.indexOf('ReadCanCollect') > -1;
+      return collection.contextParameters.permissions.indexOf('WriteProperties') > -1;
     }
     return false;
   },


richardsd:
https://qa.nuxeo.org/jenkins/job/web-ui-pipeline-rdias/61/ 🕐

elements/nuxeo-collections/nuxeo-collection-remove-action.js
```diff
@@ -91,7 +91,10 @@ Polymer({
     if (collection && collection.contextParameters && collection.contextParameters.permissions) {
       // NXP-21408: prior to 8.10-HF01 the permissions enricher wouldn't return ReadCanCollect
       // Action will therefore not be available
-      return collection.contextParameters.permissions.indexOf('ReadCanCollect') > -1;
+      return (
+        collection.contextParameters.permissions.indexOf('ReadCanCollect') > -1 ||
+        collection.contextParameters.permissions.indexOf('WriteProperties') > -1
```
Gabez0r, line 96: Since `ReadCanCollect` also includes `WriteProperties`, can't we just use `WriteProperties` here? Or is the `Read` permission also required?

richardsd: 🤔 well, if you don't have the `Read` permission, you won't be able to read the collection in the first place, right?

richardsd: So, in JSF UI they rely on the [canAddToCollection](https://github.com/nuxeo/nuxeo-jsf-ui/blob/0ae5d2226dca8bbe7e52c229b77865e4cdda8bd2/code/nuxeo-platform-webapp-base/src/main/java/org/nuxeo/ecm/webapp/collections/CollectionActionsBean.java#L219) to assert if one can remove from a collection, and the `canAddToCollection` uses the [CollectionManager](https://github.com/nuxeo/nuxeo-jsf-ui/blob/0ae5d2226dca8bbe7e52c229b77865e4cdda8bd2/code/nuxeo-platform-webapp-base/src/main/java/org/nuxeo/ecm/webapp/collections/CollectionActionsBean.java#L168) to check for permissions (which only checks for the `WriteProperties`. So maybe, we could go with your suggestion. wdyt?

mnixo (approved):

Gabez0r (approved):

PR #1169, opened 2021-03-02
WEBUI-265: automate a11y checks with axe-core

Gabez0r:
This is a POC of how we could integrate axe-core with our pipeline. For now, the goal would be for it to provide us control over  the type of violations we are expecting as well as the number of issues.

The DX is not ideal: we get a different amount of issues when running in dev mode. This is caused, in part, by the fact the server / local storage content differs. There might be other factors, which I am not currently aware of.

The fact that we do not have dummy content renders the home page testing somewhat limited, as we are certainly missing issues in the data tables, for example. I believe we should populate the server with a few documents so we could at least have data in the Recently edited. What is the best way to populate the server with some data? I would like to avoid using the rest API if possible.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
     "win-lint": "npm run win-lint:polymer && npm run lint:eslint && npm run lint:prettier",
     "format": "npm run format:polymer && npm run format:prettier && npm run format:eslint",
     "win-format": "npm run win-format:polymer && npm run format:prettier && npm run format:eslint",
-    "preinstall": "cd packages/nuxeo-web-ui-ftest && npm install",
+    "preinstall": "cd ./packages/nuxeo-web-ui-ftest && npm install && cd ../../plugin/a11y && npm install",
     "ftest": "node scripts/test-runner.js --report --screenshots --headless --tags='not @ignore'",
     "ftest:dev": "node scripts/test-runner.js --nuxeoUrl=http://localhost:8080/nuxeo --url=http://localhost:5000/",
     "ftest:watch": "node scripts/test-runner.js --nuxeoUrl=http://localhost:8080/nuxeo  --url=http://localhost:5000/ --tags='@watch' --debug",
--- a/plugin/a11y/package.json
+++ b/plugin/a11y/package.json
@@ -1,0 +1,30 @@
+{
+  "name": "@nuxeo/nuxeo-web-ui-a11y",
+  "version": "1.0.0",
+  "engines": {
+    "node": ">=10.23.0"
+  },
+  "scripts": {
+    "start": "npm run test",
+    "test": "wdio wdio.conf.js",
+    "test:dev": "NUXEO_WEB_UI_URL=http://localhost:5000/ NUXEO_URL=http://localhost:8080/nuxeo/ wdio wdio.conf.js"
+  },
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@nuxeo/nuxeo-web-ui-ftest": "file:../../packages/nuxeo-web-ui-ftest",
+    "axe-core": "^4.1.2"
+  },
+  "devDependencies": {
+    "@babel/cli": "^7.13.10",
+    "@babel/core": "^7.13.10",
+    "@babel/preset-env": "^7.13.10",
+    "@babel/register": "^7.13.8",
+    "@wdio/cli": "^7.0.7",
+    "@wdio/local-runner": "^7.0.7",
+    "@wdio/mocha-framework": "^7.0.7",
+    "@wdio/spec-reporter": "^7.0.7",
+    "@wdio/sync": "^7.0.7",
+    "chromedriver": "^88.0.0",
+    "wdio-chromedriver-service": "^7.0.0"
+  }
+}
--- a/plugin/a11y/test/axe-reporter.js
+++ b/plugin/a11y/test/axe-reporter.js
@@ -1,0 +1,34 @@
+import { source } from 'axe-core';
+
+export function runAxeCore() {
+  // inject the axe-core lib
+  browser.execute(source);
+
+  // https://github.com/dequelabs/axe-core/blob/develop/doc/API.md
+  const options = {
+    runOnly: {
+      type: 'tag',
+      values: ['ACT', 'best-practice', 'wcag2a', 'wcag2aa'],
+    },
+  };
+  // run inside browser and get results
+  const results = browser.executeAsync((opts, done) => {
+    // eslint-disable-next-line no-undef
+    axe
+      .run(opts)
+      .then((res) => done(res))
+      .catch((err) => {
+        throw err;
+      });
+  }, options);
+
+  return {
+    results,
+    incomplete: results.incomplete.map((a) => {
+      return { id: a.id, issues: a.nodes.length };
+    }),
+    violations: results.violations.map((a) => {
+      return { id: a.id, issues: a.nodes.length };
+    }),
+  };
+}
--- a/plugin/a11y/test/pageobjects/login.page.js
+++ b/plugin/a11y/test/pageobjects/login.page.js
@@ -1,0 +1,25 @@
+class LoginPage {
+  get inputUsername() {
+    return $('#username');
+  }
+
+  get inputPassword() {
+    return $('#password');
+  }
+
+  get btnSubmit() {
+    return $('input[name="Submit"]');
+  }
+
+  login(username, password) {
+    this.inputUsername.setValue(username);
+    this.inputPassword.setValue(password);
+    this.btnSubmit.click();
+  }
+
+  open() {
+    return browser.url(process.env.NUXEO_URL ? `${process.env.NUXEO_URL}logout` : 'logout');
+  }
+}
+
+export default new LoginPage();
--- a/plugin/a11y/test/specs/home.js
+++ b/plugin/a11y/test/specs/home.js
@@ -1,0 +1,39 @@
+import LoginPage from '../pageobjects/login.page.js';
+import HomePage from '../pageobjects/home.page.js';
+/* import Login from '@nuxeo/nuxeo-web-ui-ftest/pages/login';
+import UI from '@nuxeo/nuxeo-web-ui-ftest/pages/ui'; */
+import { reportA11y } from '../a11y-reporter.js';
+
+const EXPECTED_VIOLATIONS = {
+  'aria-command-name': 15,
+  'aria-input-field-name': 1,
+  'aria-required-children': 1,
+  'aria-tooltip-name': 1,
+  'duplicate-id-active': 28,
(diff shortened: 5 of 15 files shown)


plugin/a11y/test/specs/home.js
```diff
@@ -0,0 +1,118 @@
...
+              issues,
+            },
+          ]),
+        );
+      });
+    });
+
+    it(`should have ${nIssues} total issue(s)`, () => {
+      const vissues = report.violations.map(({ issues }) => issues).reduce((a, b) => a + b, 0);
+      expect(vissues).toBe(nIssues);
```
mnixo, line 75:
```suggestion
      const vIssues = report.violations.map(({ issues }) => issues).reduce((a, b) => a + b, 0);
      expect(vIssues).toBe(nIssues);
```

plugin/a11y/test/specs/home.js
```diff
@@ -0,0 +1,118 @@
...
+              issues,
+            },
+          ]),
+        );
+      });
+    });
+
+    it(`should have ${nIncompleteIssues} incomplete total issue(s)`, () => {
+      const vissues = report.incomplete.map(({ issues }) => issues).reduce((a, b) => a + b, 0);
+      expect(vissues).toBe(nIncompleteIssues);
```
mnixo, line 115:
```suggestion
      const vIssues = report.incomplete.map(({ issues }) => issues).reduce((a, b) => a + b, 0);
      expect(vIssues).toBe(nIncompleteIssues);
```

plugin/a11y/package.json
```diff
@@ -0,0 +1,20 @@
...
+    "@wdio/cli": "^7.0.7",
+    "@wdio/local-runner": "^7.0.7",
+    "@wdio/mocha-framework": "^7.0.7",
+    "@wdio/spec-reporter": "^7.0.7",
+    "@wdio/sync": "^7.0.7",
+    "axe-core": "^4.1.2",
+    "chromedriver": "^88.0.0",
+    "wdio-chromedriver-service": "^7.0.0"
+  }
+}
```
mnixo, line 20: Do we have any concerns regarding required `node` versions here?

mnixo (dismissed):

plugin/a11y/test/specs/home.js
```diff
@@ -0,0 +1,118 @@
...
+
+  _report = runAxeCore();
+
+  return _report;
+};
+
+describe('Home Page', () => {
+  context('Violations', () => {
+    const expectedViolations = {
+      'ARIA commands must have an accessible name': 15,
```
nelsonsilva, line 28: nit: perhaps we could manage this in an external file or at lest move it to the top of this file as a const `const EXPECTED_VIOLATIONS = { ...}`

plugin/a11y/test/specs/home.js
```diff
@@ -0,0 +1,118 @@
...
+      'ARIA input fields must have an accessible name': 1,
+      'Certain ARIA roles must contain particular children': 1,
+      'ARIA tooltip nodes must have an accessible name': 1,
+      'id attribute value must be unique': 28,
+      'Document must have one main landmark': 1,
+      'Zooming and scaling must not be disabled': 1,
+      'Page must contain a level-one heading': 1,
+
+      // the borwser extension is able to find 101 of these, sometimes more, in dev mode (port 5000)
```
nelsonsilva, line 37:
```suggestion
      // the browser extension is able to find 101 of these, sometimes more, in dev mode (port 5000)
```

plugin/a11y/test/specs/home.js
```diff
@@ -0,0 +1,118 @@
...
+
+    before(() => {
+      report = getReport();
+    });
+
+    it(`should have ${nViolations} violation(s)`, () => {
+      expect(report.violations.length).toBe(nViolations);
+    });
+
+    Object.entries(expectedViolations).forEach(([violation, issues]) => {
```
nelsonsilva, line 60: isn't this check enough ? The total violations isn't really meaningful as you may have gained one of a type and lost one of another and the check would still pass.

plugin/a11y/test/specs/home.js
```diff
@@ -0,0 +1,118 @@
...
+      it(`${violation}: ${issues} issue(s)`, () => {
+        expect(report.violations).toEqual(
```
nelsonsilva, line 62: would it make sense to check for "less than" instead of equal so we don't have to update the report? (I understand that then the expected violations map wouldn't be up to date but it would at least not prevent merging a PR when it actually improves a11y status)

plugin/a11y/test/specs/home.js
```diff
@@ -0,0 +1,118 @@
...
+    });
+
+    it(`should have ${nIssues} total issue(s)`, () => {
+      const vissues = report.violations.map(({ issues }) => issues).reduce((a, b) => a + b, 0);
+      expect(vissues).toBe(nIssues);
+    });
+  });
+
+  context('Incomplete violations', () => {
+    const expectedIncompleteViolations = {
```
nelsonsilva, line 80: same comment, to be moved to a constant. BTW, is there no rule id that we we could use instead of full name?

plugin/a11y/test/pageobjects/login.page.js
```diff
@@ -0,0 +1,38 @@
+/**
+ * sub page containing specific selectors and methods for a specific page
```
richardsd, line 2: The comments in this class are pretty generic, they will be cleaned up, right?

richardsd (dismissed):

semisse (dismissed):

plugin/a11y/package.json
Gabez0r: No't that we know of, but we could add:
```
"engines": {
    "node": ">=10.23.0"
  },
```
Which is probably the minimum required.

package.json
```diff
@@ -92,7 +92,7 @@
     "win-lint": "npm run win-lint:polymer && npm run lint:eslint && npm run lint:prettier",
     "format": "npm run format:polymer && npm run format:prettier && npm run format:eslint",
     "win-format": "npm run win-format:polymer && npm run format:prettier && npm run format:eslint",
-    "preinstall": "cd packages/nuxeo-web-ui-ftest && npm install",
+    "preinstall": "cd packages/nuxeo-web-ui-ftest && npm install && cd ../../plugin/a11y && npm install",
```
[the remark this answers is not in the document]
Gabez0r: Because of linting, it will fail owing to missing dependencies.

plugin/a11y/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "nuxeo-web-ui-a11y",
+  "version": "1.0.0",
+  "scripts": {
+    "start": "npm run test",
+    "test": "wdio wdio.conf.js",
+    "test:dev": "NUXEO_WEB_UI_URL=http://localhost:5000/ NUXEO_URL=http://localhost:8080/nuxeo/ wdio wdio.conf.js"
```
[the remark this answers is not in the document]
Gabez0r: I don't think I've seen this before, and I couldn't find this in the [docs](https://webdriver.io/docs/clioptions/). 🤔

plugin/a11y/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "nuxeo-web-ui-a11y",
```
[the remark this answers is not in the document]
Gabez0r: We should!

plugin/a11y/test/axe-reporter.js
```diff
@@ -0,0 +1,34 @@
...
+    runOnly: {
+      type: 'tag',
+      values: ['ACT', 'best-practice', 'wcag2a', 'wcag2aa'],
+    },
+  };
+  // run inside browser and get results
+  const results = browser.executeAsync((opts, done) => {
+    // eslint-disable-next-line no-undef
+    axe
+      .run({ exclude: [['nuxeo-path-suggestion']] }, opts)
```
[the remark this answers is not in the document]
Gabez0r: This is not even working... 😬  I will remove it.

plugin/a11y/test/specs/home.js
Gabez0r: Agree. Will move it up for now.

plugin/a11y/test/axe-reporter.js
```diff
@@ -0,0 +1,34 @@
...
+
+  // https://github.com/dequelabs/axe-core/blob/develop/doc/API.md
+  const options = {
+    runOnly: {
+      type: 'tag',
+      values: ['ACT', 'best-practice', 'wcag2a', 'wcag2aa'],
+    },
+  };
+  // run inside browser and get results
+  const results = browser.executeAsync((opts, done) => {
```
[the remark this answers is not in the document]
Gabez0r: From my understanding, browser.call doesn't  run code inside the browser. Did you mean browser.execute?

plugin/a11y/test/specs/home.js
Gabez0r: This would allow us to fix something on a commit and then break something else and the following commit, and we would be able to spot it.

The goal here is to make the effort to update this until we reach zero, and then we would never have to update this again, as we would have zero tolerance as we have with linting erros, for example. Also, this will be a record of how many issues we are fixing with the current ticket.

plugin/a11y/test/specs/home.js
Gabez0r: There is an id, but we'd have to keep the labels anyway, since these are the ones that show up on the axe tool. It could be worth it to refactor this an centralize the rules in on single place.

plugin/a11y/test/axe-reporter.js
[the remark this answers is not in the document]
Gabez0r: Tried with execute, and basically forces us into async wdio mode. awaits everywhere.

plugin/a11y/test/specs/home.js
Gabez0r: I agree.

plugin/a11y/test/pageobjects/login.page.js
Gabez0r: Yes, it was scaffolded by the wizard.

plugin/a11y/test/specs/home.js
```diff
@@ -0,0 +1,118 @@
+const runAxeCore = require('../axe-reporter');
+const LoginPage = require('../pageobjects/login.page');
```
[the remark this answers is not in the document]
Gabez0r: We can, but it requires https://github.com/nuxeo/nuxeo-web-ui/pull/1183 to land first, otherwise there will be missing wdio commands.

Gabez0r:
Closed in favor of https://github.com/nuxeo/nuxeo-web-ui/pull/1341.

PR #1172, opened 2021-03-04
WEBUI-191: fix drawer navigation by keyboard

semisse:


mnixo (approved):

richardsd (approved):

elements/nuxeo-app/nuxeo-menu-icon.js
```diff
@@ -145,4 +153,18 @@ Polymer({
       return this.urlFor(...[name].concat(args));
     }
   },
+
+  _dispatchEvent(e) {
+    // if space or enter keys were pressed
+    if (e && e.keyCode === 13 || e.keyCode === 32) {
```
mnixo, line 159: This is correct, but we're now relying purely on the precedence order of the logical operators ([which might not be taken generally as a good practice](https://softwareengineering.stackexchange.com/questions/201175/should-i-use-parentheses-in-logical-statements-even-where-not-necessary)).
If that's fine with everyone, I'm fine with it too.

mnixo (approved):

elements/nuxeo-app/nuxeo-menu-icon.js
semisse: yes you are correct, it's more robust wrapping them with parenthesis. I will do that, thanks

mnixo (approved):

richardsd (dismissed):

elements/nuxeo-app/nuxeo-menu-icon.js
```diff
@@ -145,4 +153,18 @@ Polymer({
       return this.urlFor(...[name].concat(args));
     }
   },
+
+  _dispatchEvent(e) {
```
Gabez0r, line 157: Don't think we need this if this element implements `PaperItemBehavior`:
```
import { PaperItemBehavior } from '@polymer/paper-item/paper-item-behavior.js';
...
behaviors: [I18nBehavior, RoutingBehavior, PaperItemBehavior],
```

semisse: Just tested with `PaperItemBehavior` and it seems to do the work 💪

mnixo (approved):

Gabez0r (approved):

PR #1176, opened 2021-03-11
WEBUI-191: fix drawer navigation by keyboard (Backport 3.0.x)

semisse:


richardsd (dismissed):

elements/nuxeo-app/nuxeo-menu-icon.js
```diff
@@ -145,4 +153,18 @@ Polymer({
       return this.urlFor(...[name].concat(args));
     }
   },
+
+  _dispatchEvent(e) {
+    // if space or enter keys were pressed
+    if ((e && e.keyCode === 13) || e.keyCode === 32) {
```
mnixo, line 159:
```suggestion
    if (e && (e.keyCode === 13 || e.keyCode === 32)) {
```
Right? 
Because we need `e` to not be `null` in both cases.
Let's say that `e` is `null`. The left side of the `||` (of your condition) will evaluate as `false`. On the right side you'll get a `cannot get property "keyCode" of null`.
A shorter option could also be:
```suggestion
    if (e && [13, 32].includes(e.keyCode)) {
```

semisse: Yes, good catch, probably a leftover that I didn't see. I will still use the first formulation since I think it is the most declarative if the team agrees.

mnixo (dismissed):

richardsd (dismissed):

mnixo (approved):

Gabez0r (approved):

PR #1183, opened 2021-03-17
WEBUI-278: upgrade webdriver.io

nelsonsilva:


packages/nuxeo-web-ui-ftest/pages/login.js
```diff
@@ -2,16 +2,18 @@ export default class Login {
   set username(username) {
     driver.waitForVisible('#username');
     driver.element('#username').setValue(username);
+    // $('#username').waitForDisplayed();
```
richardsd, line 5: don't we want to the use 

`$('#username').waitForDisplayed();`
`$('#username').setValue(username);`

as we do for the password?

packages/nuxeo-web-ui-ftest/pages/ui/browser.js
```diff
@@ -314,13 +317,6 @@ export default class Browser extends BasePage {
   clickDocumentActionMenu(selector) {
     const menu = this.el.element('nuxeo-actions-menu');
     menu.waitForExist(selector);
-    const children = menu.elements('nuxeo-actions-menu > *:not([show-label])');
```
richardsd, line 317: We don't this logic anymore?

packages/nuxeo-web-ui-ftest/pages/ui/browser/document_comment_thread.js
```diff
@@ -22,15 +22,19 @@ export default class DocumentCommentThread extends BasePage {
   }
 
   getComment(text, user) {
-    const comments = this.el.elements('nuxeo-document-comment').value;
+    // XXX - support the possessive form (ie John's) for compat
```
richardsd, line 25: 🤔 why did you changed the step definition signature, in favor of doing the assert here? To be more generic?

packages/nuxeo-web-ui-ftest/pages/ui/browser/document_comment_thread.js
```diff
@@ -22,15 +22,19 @@ export default class DocumentCommentThread extends BasePage {
...
+    const comments = this.el.elements('nuxeo-document-comment');
     const match = comments.find((item) => {
       const comment = new DocumentComment(item);
       return comment.author.getText() === user && comment.text.getText() === text;
     });
     if (match) {
       return new DocumentComment(match);
     }
-    throw new Error(`Not found any comment authored by "${user}" with the following text: "${text}"`);
+    throw new Error(`No comment authored by "${user}" with the following text: "${text}" found`);
```
richardsd, line 37: Nit: maybe `No comment authored by "${user}" with the following text: "${text}" was found`

packages/nuxeo-web-ui-ftest/pages/ui/group.js
```diff
@@ -3,7 +3,7 @@ import BasePage from '../base';
 export default class Group extends BasePage {
   getField(field) {
     driver.waitForExist(this._selector);
```
richardsd, line 5: To be consistent: `$(this._selector).waitForExist();`
wdyt? It works, right?

packages/nuxeo-web-ui-ftest/pages/ui/results.js
```diff
@@ -69,12 +69,15 @@ export default class Results extends BasePage {
...
     const row = this.el.element('nuxeo-data-table-row[header]');
     row.waitForVisible('nuxeo-data-table-cell:not([hidden])');
-    return row.elementByTextContent('nuxeo-data-table-cell:not([hidden])', heading);
+    return row.elements('nuxeo-data-table-cell:not([hidden])').find((e) => e.getText() === heading);
   }
 
   resultsCount(displayMode) {
     const rows = this.getResults(displayMode);
-    return rows.value.filter((result) => result.getAttribute('hidden') === null).length;
+    if (!rows) {
```
richardsd, line 77: nit:

`return rows ? rows.filter((result) => result.getAttribute('hidden') === null).length : 0;`

packages/nuxeo-web-ui-ftest/wdio-shadow-plugin.js
```diff
@@ -101,115 +102,101 @@ function findDeep(selector, multiple, baseElement, filterBy) {
...
-
-  browser.addCommand('elementByTextContent', function(selector, textContent) {
-    this.waitForVisible(selector);
-    return this.shadowElement(selector, true, `return element.textContent.trim() === "${textContent.trim()}";`);
-  });
-
-  browser.addCommand('scrollIntoView', function(selector) {
-    return this.shadowExecute(selector, (element) =>
-      (Array.isArray(element) && element.length > 0 ? element[0] : element).scrollIntoView(),
+    browser.addCommand('elementByTextContent', function(selector, textContent) {
```
richardsd, line 172: I still this method here, but you removed the calls from the step definitions, right? Do we still want to it?

packages/nuxeo-web-ui-ftest/scripts/specs-processor.js
```diff
@@ -1,7 +1,7 @@
 const Finder = require('fs-finder');
 const fs = require('fs');
 const minimist = require('minimist');
-const { TagExpressionParser } = require('cucumber-tag-expressions');
+const { TagExpressionParser } = require('@cucumber/tag-expressions');
```
richardsd, line 4: We want to delete the spec processor right? I noticed we are not using it in the wdio.conf file anymore.

richardsd (approved):
Awesome job Nelson 👏

packages/nuxeo-web-ui-ftest/wdio-shadow-plugin.js
```diff
@@ -101,115 +102,101 @@ function findDeep(selector, multiple, baseElement, filterBy) {
     if (Array.isArray(result)) {
       return result.filter(filter);
     }
-    return filterBy(result) ? result : null;
+    return filter(result) ? result : null;
```
Gabez0r, line 105: Good catch. 🙏

packages/nuxeo-web-ui-ftest/wdio-shadow-plugin.js
Gabez0r: Ideally, we would keep it because removing it would be a breaking change. But of course, that would only make sense if the method can actually be used.

packages/nuxeo-web-ui-ftest/wdio.conf.js
```diff
@@ -1,54 +1,65 @@
...
 const capability = {
   // maxInstances can get overwritten per capability. So if you have an in-house Selenium
   // grid with only 5 firefox instance available you can make sure that not more than
   // 5 instance gets started at a time.
   maxInstances: 1,
   browserName: process.env.BROWSER,
-  javascriptEnabled: true,
-  acceptSslCerts: true,
+  // javascriptEnabled: true,
+  // acceptSslCerts: true,
```
Gabez0r, line 31: Do we need to keep these disabled?

packages/nuxeo-web-ui-ftest/features/step_definitions/comments.js
```diff
@@ -103,28 +103,28 @@ Then("I can see the reply thread for {word}('s) comment {string} has a total of
   link.getText().should.be.equals(`View all ${total} replies`);
 });
 
-Then("I can see {word}('s) comment: {string}", function(user, text) {
```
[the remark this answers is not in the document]
nelsonsilva: iirc it was capturing the optional text as part of the word

packages/nuxeo-web-ui-ftest/features/step_definitions/search.js
```diff
@@ -151,7 +150,7 @@ Then(/^I can see (\d+) search results$/, function(numberOfResults) {
 
 Then(/^I can see more than (\d+) search results$/, function(minNumberOfResults) {
   const { displayMode } = this.ui.results;
-  this.ui.results.getResults(displayMode).waitForVisible();
+  // this.ui.results.getResults(displayMode).waitForVisible();
```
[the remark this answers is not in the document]
nelsonsilva: TBH I feel we have way to many `waitForVisible` throughout the code base and the check should normally be implicit by the following assertions where you try to do something with the element (which won't work if it's invisble).
In this particular case the problem is that we're doing `waitForVisible` on a list of elements, not a single one, which doesn't seem to be working as before.

packages/nuxeo-web-ui-ftest/features/step_definitions/browser.js
```diff
@@ -84,7 +86,7 @@ Then('I can move selection up', function() {
   this.ui.browser.selectionToolbar.moveUp();
 });
 
-Then('I can see the {string} child document is at position "{int}"', function(title, pos) {
+Then('I can see the {string} child document is at position {int}', function(title, pos) {
```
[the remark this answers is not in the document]
nelsonsilva: not required no but I failed to understand why we'd surround a number with quotes...

packages/nuxeo-web-ui-ftest/features/step_definitions/tasks.js
```diff
@@ -11,7 +11,6 @@ When(/^I (\w+) the task for following actors:$/, function(option, table) {
   } else if (option === 'reassign') {
     this.ui.browser.documentTaskView.reassignOption.click();
   }
-  this.ui.browser.documentTaskView.assignmentDialog.waitForVisible();
```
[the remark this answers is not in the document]
nelsonsilva: I understand but I feel these should be part of the page objects, otherwise we'll keep on adding `waitForVisible` throughout the codebase, no?

packages/nuxeo-web-ui-ftest/pages/ui/browser.js
```diff
@@ -206,10 +209,12 @@ export default class Browser extends BasePage {
     this.waitForChildren();
     const { rows } = this;
     let i;
-    for (i = 0; i < rows.value.length; i++) {
+    for (i = 1; i < rows.length; i++) {
       if (
-        rows.value[i].isVisible('nuxeo-data-table-cell a.title') &&
```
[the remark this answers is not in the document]
nelsonsilva: Webdriver commands normally require elements to be visible and not disabled so I think this is redundant

packages/nuxeo-web-ui-ftest/pages/ui/browser.js
```diff
@@ -244,12 +248,11 @@ export default class Browser extends BasePage {
    * Results might vary with the viewport size as only visible items are taken into account.
    */
   waitForNbChildren(nb) {
-    this.waitForVisible('nuxeo-data-table nuxeo-data-table-row');
```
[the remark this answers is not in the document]
nelsonsilva: Same thing .. the new API is all around element commands. I introduced a driver level `waitForVisible` command for compatibility but I'm currently doing `this.$(selector).waitForVisible()` perhaps I should change to use a more complex `waitUntil` so I can do `this.$$(selector).some(isVisible)` or something 🤔

packages/nuxeo-web-ui-ftest/pages/ui/browser/document_comment_thread.js
nelsonsilva: IIRC the step definition when using `{word}` was already capturing the `'s` which was meant to be optional... will check

packages/nuxeo-web-ui-ftest/pages/ui/browser/document_comment_thread.js
nelsonsilva: Going with `No comment authored by "${user}" with the text "${text}" was found`

package.json
```diff
@@ -10,20 +10,33 @@
...
+    "@rpii/wdio-html-reporter": "^6.3.2",
+    "@wdio/cli": "^7.2.0",
+    "@wdio/cucumber-framework": "^7.2.0",
+    "@wdio/junit-reporter": "^7.2.0",
+    "@wdio/local-runner": "^7.2.0",
+    "@wdio/selenium-standalone-service": "^7.2.0",
+    "@wdio/spec-reporter": "^7.2.0",
+    "@wdio/sync": "^7.2.0",
     "@webcomponents/webcomponentsjs": "^2.0.0",
     "babel-preset-env": "^1.7.0",
```
Gabez0r, line 33: Do we still need this?

packages/nuxeo-web-ui-ftest/wdio.conf.js
```diff
@@ -73,23 +84,19 @@ require('babel-register')({
 });
```
Gabez0r, line 84: Do we need still need this? Was unsure after reading https://webdriver.io/docs/babel.

packages/nuxeo-web-ui-ftest/wdio.conf.js
nelsonsilva: don't remember why i commented this out tbh. I guess those are just the default anyway otherwise it wouldn't work

nelsonsilva: ok, those are note valid w3c capabilities, see https://w3c.github.io/webdriver/#capabilities

packages/nuxeo-web-ui-ftest/features/step_definitions/comments.js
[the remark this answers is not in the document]
nelsonsilva: Optional text isn't working as far as I can tell so I switched to regex

packages/nuxeo-web-ui-ftest/features/step_definitions/video.js
```diff
@@ -7,9 +7,20 @@ Then('I can see the video conversions panel', function() {
...
+  driver.waitUntil(() => {
+    const page = this.ui.browser.documentPage(this.doc.type);
+    if (!page.isVisible()) {
+      return false;
+    }
+    const videoViewer = page.el.element('nuxeo-video-viewer');
+    if (!videoViewer.isVisible()) {
+      return false;
+    }
+    if (videoViewer.element('#storyboard').isVisible() !== true) {
```
nelsonsilva, line 19: since this is the only element that may not yet be ready shouldn't we just keep this bit in the `driver.waitUntil` block?

packages/nuxeo-web-ui-ftest/pages/ui/browser.js
```diff
@@ -227,9 +227,10 @@ export default class Browser extends BasePage {
   sortContent(field, order) {
     driver.waitUntil(() => {
       this.waitForChildren();
-      const idx = this.currentPage
-        .elements('nuxeo-data-table nuxeo-data-table-column')
-        .map((col) => col.getAttribute('sort-by'))
+      const dataTable = this.currentPage.element('nuxeo-data-table');
+      const columns = dataTable.elements('nuxeo-data-table-column');
+      const idx = columns
+        .map((col) => browser.execute((el) => el.sortBy, col))
```
nelsonsilva, line 233: the `elements()` bit was working properly, right ? it's just the `sortBy` which is not being reflected to the attribute and thus we need to retrieve the property, yes ?

packages/nuxeo-web-ui-ftest/wdio.conf.js
```diff
@@ -1,54 +1,63 @@
...
-      capability.chromeOptions.args.push('--disable-gpu');
-      capability.chromeOptions.args.push('--disable-dev-shm-usage');
+      options.args.push('--window-size=1920,1080');
+      options.args.push('--single-process');
+      options.args.push('--headless');
+      options.args.push('--disable-gpu');
+      options.args.push('--disable-dev-shm-usage');
     }
     if (process.env.BROWSER_BINARY) {
       capability.chromeOptions.binary = process.env.BROWSER_BINARY;
```
richardsd, line 48: Don't we need to change this line too, to something like `options.binary = process.env.BROWSER_BINARY`, similar to what you did for `firefox`? (see below https://github.com/nuxeo/nuxeo-web-ui/pull/1183/files#diff-55b30c5d9c6eb640a9a5539c52cf56daee72c2eb0539ad5ff613e7a826ac2617L50)

packages/nuxeo-web-ui-ftest/wdio.conf.js
```diff
@@ -182,38 +193,37 @@ exports.config = {
...
+    profile: [],
+    // <boolean> fail if there are any undefined or pending steps
+    strict: true,
     // <string> (expression) only execute the features or scenarios with tags matching the expression
-    // Details can be found here: https://cucumber.io/docs/cucumber/api/#tag-expressions
     tagExpression: process.env.TAG_EXPRESSION,
-    timeout: process.env.DEBUG ? 24 * 60 * 60 * 1000 : TIMEOUT + 5000, // <number> timeout for step definitions
-    ignoreUndefinedDefinitions: false, // <boolean> Enable this config to treat undefined definitions as warnings.
+    // <number> timeout for step definitions
+    timeout: process.env.DEBUG ? 24 * 60 * 60 * 1000 : TIMEOUT + 500,
```
richardsd, line 224: Why did we change from `TIMEOUT + 5000` to `TIMEOUT + 500`?

richardsd (approved):
I just left you some minor comments, because I already reviewed most of the code. Other than that, it lgtm 👍

packages/nuxeo-web-ui-ftest/wdio.conf.js
Gabez0r: I think that makes sense! Thanks.

packages/nuxeo-web-ui-ftest/wdio.conf.js
Gabez0r: I'm not sure. 🤔  I guess in the end it doesn't make much of a difference. We just wanted this to be larger than the   waitforTimeout.

richardsd: Fair enough to me 👍 thanks.

packages/nuxeo-web-ui-ftest/features/step_definitions/create_dialog.js
```diff
@@ -11,7 +11,10 @@ When('I click the Create Document button', function() {
 Then('I click the Create button to finish the import', function() {
   const importButton = this.ui.createDialog.importCreateButton;
   importButton.waitForVisible();
+  importButton.waitForEnabled();
   importButton.click();
+  driver.waitForExist('iron-overlay-backdrop', driver.options.waitForTimeout, true);
```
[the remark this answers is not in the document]
Gabez0r: We are waiting for the overlay to not exists. The true inverts it.

packages/nuxeo-web-ui-ftest/features/step_definitions/video.js
Gabez0r: Not sure I follow. It is inside the waitUntil block already. 🤔

packages/nuxeo-web-ui-ftest/pages/ui/browser.js
Gabez0r: Yes, the sortBy is the only issue here.

richardsd (dismissed):

packages/nuxeo-web-ui-ftest/features/step_definitions/create_dialog.js
```diff
@@ -11,7 +11,10 @@ When('I click the Create Document button', function() {
...
+  driver.pause(3000); // XXX just give it some time to the server to do the convertions
```
mnixo, line 17:
```suggestion
  driver.pause(3000); // XXX just give it some time to the server to do the conversions
```

mnixo (approved):

mnixo (dismissed):

richardsd (dismissed):

richardsd (approved):

mnixo (approved):

PR #1184, opened 2021-03-23
WEBUI-204: add alt attribute to image elements

semisse:


addons/nuxeo-platform-3d/elements/nuxeo-3d-render-views.js
```diff
@@ -69,7 +69,7 @@ Polymer({
       <div class="render-views">
         <template is="dom-repeat" items="[[document.properties.threed:renderViews]]" as="renderView">
           <div class="item">
-            <img src="[[renderView.thumbnail.data]]" on-tap="_setCoords" />
+            <img src="[[renderView.thumbnail.data]]" on-tap="_setCoords" alt="[[document.title]]" />
```
Gabez0r, line 72: I think we should use $= here instead, since we are binding to an attribute and not a property. Same applies to the changes below.

elements/nuxeo-sardine.js
```diff
@@ -117,7 +117,7 @@ Polymer({
       <div id="wrapper">
         <span id="feeling">[[feeling]]</span>
         <template is="dom-if" if="[[!hidden]]">
-          <img src="images/sardine.png" />
+          <img src="images/sardine.png" alt="nuxeo sardine" />
```
mnixo, line 120: 🏆

mnixo (approved):

Gabez0r (approved):

richardsd (approved):

PR #1218, opened 2021-04-26
WEBUI-333: set ui config from functional tests

Gabez0r:
Depends on https://github.com/nuxeo/nuxeo-elements/pull/407.

- Adds `@config(key, value)` cucumber tag to enable a given configuration property for a specific scenario or feature file.
- Adds a hook to the application where it waits for an event `automation-ready` to be fired before defining the `nuxeo-app` element, IF automation is enabled (i.e, the browser was launched by webdriver).
- Adds new helper methods `url` and `refresh`, which encapsulate the respective webdriver commands but also sends in the properties defined in the tests using the `@cofig` tag and then fire the `automation-ready` event. This means that the application will only load after the configuration has been loaded.

Note that the diff for nuxeo-app looks huge, but I just wrapped the nuxeo-app element definition with:
```
prom.then(() => {
  ...
});
```
### Experiments

Experiment A: Adding a fallback to the get helper, which tries to evaluate the result based on the fallback data type (see https://github.com/nuxeo/nuxeo-web-ui/pull/1218/commits/ec0d3f4ec7d8dc5ca2c3598efb1199b4911c32b7).
Experiment B: Moving the automation hook to index.js, which requires some refactoring (see https://github.com/nuxeo/nuxeo-web-ui/pull/1218/commits/cfa16025f54ae5cd4078c78619cfb37a92f337be).

### DIFF
--- a/elements/nuxeo-document-creation/nuxeo-document-creation-behavior.js
+++ b/elements/nuxeo-document-creation/nuxeo-document-creation-behavior.js
@@ -14,8 +14,9 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-import { FormatBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-format-behavior.js';
 
+import { config } from '@nuxeo/nuxeo-elements';
+import { FormatBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-format-behavior.js';
 import { RoutingBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-routing-behavior.js';
 
 let schemaFetcher = null;
@@ -92,7 +93,7 @@ export const DocumentCreationBehavior = [
         'fetch-document': 'properties',
         'translate-directoryEntry': 'label',
       };
-      schemaFetcher.enrichers = (Nuxeo.UI && Nuxeo.UI.config && Nuxeo.UI.config.enrichers) || {};
+      schemaFetcher.enrichers = config.get('enrichers', {});
       return schemaFetcher.get().then((doc) => {
         if (properties) {
           Object.keys(properties).forEach((prop) => {
--- a/packages/nuxeo-web-ui-ftest/features/step_definitions/support/ui_config.js
+++ b/packages/nuxeo-web-ui-ftest/features/step_definitions/support/ui_config.js
@@ -1,0 +1,11 @@
+import { Before } from 'cucumber';
+
+Before((e) => {
+  const { tags } = e.pickle;
+  global.config = tags
+    .map((tag) => {
+      const res = tag.name.match(/@config\('(.*)','(.*)'\)/);
+      return res && { key: res[1], value: res[2] };
+    })
+    .filter(Boolean);
+});
--- a/packages/nuxeo-web-ui-ftest/pages/helpers.js
+++ b/packages/nuxeo-web-ui-ftest/pages/helpers.js
@@ -1,0 +1,18 @@
+const _flushProperties = () => {
+  driver.execute((conf) => {
+    conf.forEach(({ key, value }) => Nuxeo.UI.config.set(key, value));
+    document.dispatchEvent(new CustomEvent('automation-ready'));
+  }, global.config);
+};
+
+const refresh = () => {
+  driver.refresh();
+  _flushProperties();
+};
+
+const url = (...args) => {
+  driver.url(...args);
+  _flushProperties();
+};
+
+export { refresh, url };
--- a/elements/nuxeo-app.js
+++ b/elements/nuxeo-app.js
@@ -19,6 +19,7 @@ import '@polymer/polymer/polymer-legacy.js';
 import '@nuxeo/nuxeo-elements/nuxeo-document.js';
 import '@nuxeo/nuxeo-elements/nuxeo-operation.js';
 import '@nuxeo/nuxeo-elements/nuxeo-resource.js';
+import { config } from '@nuxeo/nuxeo-elements';
 import '@nuxeo/nuxeo-ui-elements/nuxeo-layout.js';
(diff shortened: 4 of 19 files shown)


Gabez0r:
Pipeline check https://qa.nuxeo.org/jenkins/view/All/job/web-ui-pipeline-gbarata/374/ 🔵

elements/nuxeo-document-creation/nuxeo-document-creation-behavior.js
```diff
@@ -92,7 +93,7 @@ export const DocumentCreationBehavior = [
         'fetch-document': 'properties',
         'translate-directoryEntry': 'label',
       };
-      schemaFetcher.enrichers = (Nuxeo.UI && Nuxeo.UI.config && Nuxeo.UI.config.enrichers) || {};
+      schemaFetcher.enrichers = config.get('enrichers') || {};
```
nelsonsilva, line 96: we can still do `config.enrichers` right ? or we should stop doing it?

package.json
```diff
@@ -103,8 +103,8 @@
     "@nuxeo/chart-elements": "^4.0.0-nx.1",
     "@nuxeo/moment": "^2.24.0-nx.0",
     "@nuxeo/nuxeo-dataviz-elements": "^3.1.0-rc.0",
-    "@nuxeo/nuxeo-elements": "^3.1.0-rc.0",
-    "@nuxeo/nuxeo-ui-elements": "^3.1.0-rc.0",
+    "@nuxeo/nuxeo-elements": "file:../nuxeo-elements/core/nuxeo-nuxeo-elements-3.1.0-SNAPSHOT.tgz",
```
nelsonsilva, line 106: guess this was unintended

packages/nuxeo-web-ui-ftest/features/step_definitions/support/ui_config.js
```diff
@@ -0,0 +1,11 @@
+import { Before } from 'cucumber';
+
+Before((e) => {
+  const { tags } = e.pickle;
+  global.config = tags
+    .map((tag) => {
+      const res = tag.name.match(/@config\('(.*)','(.*)'\)/);
```
nelsonsilva, line 7: nit: `const [_, key, value] = ...` allows you to return `{key, value}`

packages/nuxeo-web-ui-ftest/pages/helpers.js
```diff
@@ -0,0 +1,18 @@
+const _flushProperties = () => {
+  driver.execute((conf) => {
+    conf.forEach(({ key, value }) => Nuxeo.UI.config.set(key, value));
+    document.dispatchEvent(new CustomEvent('automation-ready'));
```
nelsonsilva, line 4: not sure I understand what we're doing with the `automation-ready` event 🤔

nelsonsilva (commented):
This will require people to use the new `url` and `refresh` helpers for the config properties to be properly set, right ?

elements/nuxeo-document-creation/nuxeo-document-creation-behavior.js
Gabez0r: You can still do `Nuxeo.UI.config.enrichers`. I left it on the addons, for example. But of course, using the get helper is shorter and cleaner.

package.json
Gabez0r: yup

Gabez0r: Will clean it up. Thanks.

packages/nuxeo-web-ui-ftest/pages/helpers.js
Gabez0r: nuxeo-app won't be defined until this event is fired, if the browser was launched in automation mode.

Gabez0r:
> This will require people to use the new url and refresh helpers for the config properties to be properly set, right ?
@nelsonsilva: Yup. The alternative would be to override the default wdio commands, but it would be tricky as we would have to be able to refer to the original ones. But perhaps it might be better if we intent to reuse the page objects outside the framework?

packages/nuxeo-web-ui-ftest/features/step_definitions/support/ui_config.js
Gabez0r: but if the tag does not match, it will return null which will result in an error, right?

packages/nuxeo-web-ui-ftest/pages/helpers.js
richardsd: Is it this Gabe https://github.com/nuxeo/nuxeo-web-ui/pull/1218/files#diff-af6c1d346048ac2ffccfb6dff23c31fe30da8d343dae3ddde3d1880b5b59bd5dR99?

richardsd (dismissed):
Please coordinate this is @nelsonsilva because of the webdriverio upgrade 👍 but it lgtm 🙂

richardsd (approved):

Gabez0r:
Pipeline https://qa.nuxeo.org/jenkins/view/All/job/web-ui-pipeline-gbarata/378/ 🔵

Gabez0r:
Pipeline ok 🔵 : https://qa.nuxeo.org/jenkins/view/All/job/web-ui-pipeline-gbarata/381/

mnixo (approved):

packages/nuxeo-web-ui-ftest/pages/helpers.js
Gabez0r: It's being used here, now: https://github.com/nuxeo/nuxeo-web-ui/pull/1218/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R45

richardsd (approved):

semisse (approved):

mnixo (approved):

mnixo (dismissed):

mnixo (approved):

semisse (approved):

PR #1225, opened 2021-05-07
WEBUI-321: fix aria-label on nuxeo-document-create-button

semisse:


Gabez0r (approved):

mnixo (approved):

richardsd (approved):

semisse:
Looks like AxeDevTools is picking this as a false positive because it can't access the nuxeo-tooltip even though it is present. However, voice-over on mac can pick it up and read the proper aria-label as you can check in the video:
https://user-images.githubusercontent.com/1880896/119638484-9dba8b80-be0e-11eb-8a64-e4de086ea8ee.mov

AxeDevTools report:

Fix at least one (1) of these issues:
Element does not have text that is visible to screen readers

aria-label attribute does not exist or is empty

aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty

Element has no title attribute

I will close this PR since the problem is already fixed.

PR #1230, opened 2021-05-11
WEBUI-237: Upgrade threejs and add GLTF 1.0 loaders (three@v81)

mnixo:
This is the proposed (short) solution to upgrade `threejs`:
- `threejs` version is bumped from `0.81.0` to `0.125.0`.
- The GLTF v1.0 loaders that were deprecated and removed from `threejs` are added to the 3D addon codebase (~same contents and structure~ with minor changes so that they can be imported via ESM). The controls were also added.

Luckily, everything seems to be running smoothly.
~Still looking into `webpack` warnings that are showing up when Web UI is started in development mode: `webpack` can no longer resolve the methods that are imported from the loader files.~ Fixed.


semisse (approved):

richardsd (dismissed):

richardsd:
> This is the proposed (short) solution to upgrade `threejs`:
> 
> * `threejs` version is bumped from `0.81.0` to `0.125.0`.
> * The GLTF v1.0 loaders that were deprecated and removed from `threejs` are added to the 3D addon codebase (same contents and structure).
> 
> Luckily, everything seems to be running smoothly.
> Still looking into `webpack` warnings that are showing up when Web UI is started in development mode: `webpack` can no longer resolve the methods that are imported from the loader files.

@mnixo just out of curiosity, did they deprecate and remove the loaders in favor of anything else?

mnixo:
> @mnixo just out of curiosity, did they deprecate and remove the loaders in favor of anything else?

Yes, they replaced it with a very improved new implementation ([GLTFLoader v2](https://github.com/mrdoob/three.js/pull/9786)) which initially included "all of the features of the legacy glTF code". Later they ended up doing further changes which resulted in dropping the support for GLTF v1.0 models.

Gabez0r (commented):
LGTM. I think we should add `addons/nuxeo-platform-3d/loaders/README.md` explaining what version of THREE.js we have copied these from, and perhaps why we need to keep them (or we could just add this part to the ticket). I think this also deserves a follow up ticket to actually do the upgrade to be able to use gltf 2.0,  particularly because we will be blind to any potential security threats these loads might represent. Not sure we will tackle them any time soon, though 😬

mnixo:
> we should add `addons/nuxeo-platform-3d/loaders/README.md` explaining what version of THREE.js we have copied these from

I'll do that.

> this also deserves a follow up ticket to actually do the upgrade to be able to use gltf 2.0, particularly because we will be blind to any potential security threats these loads might represent.

That would still be [TL-372: Assess upgrading the internal 3D model representation to GLTF 2.0](https://jira.nuxeo.com/browse/TL-372).

Gabez0r (approved):

richardsd (approved):

semisse (approved):

mnixo:
⏳ https://qa.nuxeo.org/jenkins/job/web-ui-pipeline-mnixo/50/

PR #1242, opened 2021-06-08
WEBUI-360: fix getAttribute checks

Gabez0r:


packages/nuxeo-web-ui-ftest/features/step_definitions/support/fixtures/layouts.js
```diff
@@ -159,11 +159,11 @@ global.fieldRegistry.register(
...
 global.fieldRegistry.register(
   'paper-checkbox',
-  (element) => element.getAttribute('checked'),
+  (element) => element.getAttribute('checked') !== null,
   (element, value) => {
     if (
-      (value === false && element.getAttribute('checked') === 'true') ||
-      (value === true && !element.getAttribute('checked'))
+      (value === false && element.getAttribute('checked') !== null) ||
+      (value === true && element.getAttribute('checked') === null)
```
mnixo, line 166: Is this one really needed? If we're expecting `element.getAttribute('checked')` to be `null`, wouldn't `!element.getAttribute('checked')` be enough? `!null` should be `true` too.
Is it because it might evaluate to `undefined` or `0` under certain conditions? (`!undefined` and `!0` are `true` as well)

mnixo (approved):

richardsd (approved):

PR #1260, opened 2021-07-20
WEBUI-362: fix tabs height

semisse:


elements/nuxeo-browser.html
```diff
@@ -40,12 +40,16 @@
 
       paper-tabs {
         --paper-tabs-selection-bar-color: transparent;
-        height: auto;
+        max-height: 42px;
```
Gabez0r, line 43: Does it need to be a `max-height` instead of a `height`?

Gabez0r (commented):
looking good

elements/nuxeo-browser.html
semisse: I was trying to avoid having a hardcoded height here, but it won't make a huge difference in this case.

Gabez0r: I'm ok with either way, was just trying to understand if there was any reason in particular. For example, if there are no tabs being contributed, what happens with one or the other?

mnixo (approved):

elements/nuxeo-browser.html
semisse: In that edge case, we should go with max-height and make sure height is set to auto so the tab bar won't be visible. Just updated the PR to reflect this.

mnixo (dismissed):

richardsd (dismissed):

Gabez0r (approved):

semisse:
I suggest we go with min-height for tabs to ensure they have space to increase their height if needed. Since having no tabs is probably an edge case, let's keep it as it is. If needed, later we can always check if the `paper-tabs` have children being contributed and make it display none in the UI.

mnixo (approved):

richardsd (approved):

PR #1264, opened 2021-08-04
WEBUI-278: upgrade webdriver.io (3.0.x)

Gabez0r:
Backport of https://github.com/nuxeo/nuxeo-web-ui/pull/1183.

Ended up having to backport [WEBUI-10](https://jira.nuxeo.com/browse/WEBUI-10) as well in order to minimize conflicts. It also improves test coverage, so I think it's a good addition. WDYT?


richardsd:
> Backport of #1183.
> 
> Ended up having to backport [WEBUI-10](https://jira.nuxeo.com/browse/WEBUI-10) as well in order to minimize conflicts. It also improves test coverage, so I think it's a good addition. WDYT?

Sounds good to me 👍

richardsd (dismissed):
Should we wait to merge this PR and the master until legacy CI is live again?

mnixo (dismissed):

richardsd (approved):

Gabez0r:
Sorry for the force push. Ended up dropping the commit that disabled the Authorized Applications functional tests (see https://github.com/nuxeo/nuxeo-web-ui/pull/1265).

packages/nuxeo-web-ui-ftest/features/step_definitions/support/fixtures/layouts.js
```diff
@@ -82,8 +82,7 @@ global.fieldRegistry.register(
   (element, value) => {
     element.element('.input-element input').click();
     element.waitForExist('paper-item');
-    const item = element.elementByTextContent('paper-item', value);
-    item.waitForExist();
+    const item = element.elements('paper-item').find((e) => e.getText() === value);
```
semisse, line 85: wouldn't be safer to wait for item to exist?

Gabez0r: we already have `element.waitForExist('paper-item');` in the previous line, so I'm not sure adding another one would help?

semisse: I'm asking this because in the older version we still check if its value exist. If we find that this is no longer needed, I'm fine with it

Gabez0r: I'm not even sure why we had it... We check for it to exist, we get element, and we check for it to exist again...

Gabez0r: Perhaps it made sense with the `elementByTextContent ` helper, which we don't have anymore? 🤔

semisse: In the older version, we check if its value exists, which might be different, right?

Gabez0r: In the older version, we are checking if a paper-item element with a given text exists. Here, we check if if we have an existing paper-item, and we iterate on all of them and retrieve one with the given text. But if we can iterate on them and get their text, then it is because they exist, right?

semisse: Makes sense 👍

packages/nuxeo-web-ui-ftest/pages/base.js
```diff
@@ -4,8 +4,7 @@ export default class BasePage {
   }
 
   get el() {
-    driver.waitForExist(this._selector);
-    return driver.element(this._selector);
+    return $(this._selector);
```
semisse, line 7: Same here. Don't we need to check for its existence anymore?

packages/nuxeo-web-ui-ftest/features/step_definitions/support/fixtures/layouts.js
Gabez0r: Cool. Let's give it a shot. We might have to revisite some of these tests and improve their stability later...

semisse (approved):
I'm a little bit worried that by dropping a couple of `waitForVisible` we might get flaky tests. But let's give it a try and if that's the case we might fix them later.

Other than that, lgtm! Thank you for your work!

packages/nuxeo-web-ui-ftest/pages/base.js
Gabez0r: If I understood the API changes correctly, I believe this is not needed because on [v4](http://v4.webdriver.io/api/protocol/element.html), we would get a `NoSuchElement` exception if the element didn't exist whereas on [v7](https://webdriver.io/docs/api/element/waitForExist) this is not the case, and the `waitForExist` can only be invoked after fetching the element.

mnixo (approved):

PR #1268, opened 2021-08-10
WEBUI-278: improve compat with legacy v4 API

Gabez0r:
Note that https://github.com/nuxeo/nuxeo-web-ui/pull/1268/commits/7e6605ca69a52ef339f6240bd59ab69b310ed98b adds support to a couple of features, which got broken with https://github.com/nuxeo/nuxeo-web-ui/pull/1183/:
1) projects relying on `babel-preset-env`, usually via a local `.babelrc` file, which can now be deleted
2) projects relying on cucumber v4, by using `babel-plugin-transform-rename-import` to rewrite imports such as follows:
    ```
    import { Given, When, Then } from 'cucumber';
    ```
    to
    
    ```
    import { Given, When, Then } from '@cucumber/cucumber';
    ```
On the other hand, b803d69252b4b13b55c77a2cba9200a3bf8accb8 adds support to accessing the results of the `findElementes` command via `const res = elements(...).value;`, which now should be written `const res = elements(...);`

This means that 7e6605ca69a52ef339f6240bd59ab69b310ed98b and b803d69252b4b13b55c77a2cba9200a3bf8accb8 can be reverted later once we no longer need the compatibility layer.

Also updated the supported node version of node for the functional test framework, which does not work with node v16. In order to make it compatible with node v16 we need to update more dependencies, which will probably break compatibility with node v10.


richardsd (dismissed):
Looks good to me 👍

semisse (dismissed):

package.json
```diff
@@ -10,30 +10,17 @@
...
-    "@wdio/cli": "^7.2.0",
-    "@wdio/cucumber-framework": "7.7.7",
-    "@wdio/junit-reporter": "^7.2.0",
-    "@wdio/local-runner": "^7.2.0",
-    "@wdio/selenium-standalone-service": "^7.2.0",
-    "@wdio/spec-reporter": "^7.2.0",
-    "@wdio/sync": "^7.2.0",
     "@webcomponents/webcomponentsjs": "^2.0.0",
-    "babel-preset-env": "^1.7.0",
+    "babel-plugin-transform-rename-import": "^2.3.0",
```
richardsd, line 21: The use of the rename plugin is very clever 👍 Nice

packages/nuxeo-web-ui-ftest/wdio.conf.js
```diff
@@ -11,7 +11,7 @@ if (process.env.CUCUMBER_REQUIRES) {
   cucumberRequires.push(process.env.CUCUMBER_REQUIRES);
 }
 
-const reporters = [];
+const reporters = ['spec'];
```
richardsd, line 14: Thank you for this one 🙇

richardsd (dismissed):

nelsonsilva (approved):

nelsonsilva (approved):

richardsd (approved):

PR #1274, opened 2021-08-18
WEBUI-394: convert dates to ISO in analytics

semisse:


semisse:
If the team agrees with this solution, maybe we could extract it to a helper method?

richardsd (approved):
I would be ok with this 👍 does it have any impact on anything else?

semisse:
@richardsd I would have to investigate. Since this legacy, we might not want to invest more on it. I will wait for more answers.

mnixo (approved):

Gabez0r (approved):

PR #1275, opened 2021-08-19
WEBUI-261: Rely on document specific schemas instead of type schemas for diff

mnixo:


mnixo:
I ended up not extending on error handling when getting the schemas information because:
- We were not doing any additional error handling before (when getting the schemas information for the types of each of the documents).
- There isn't anything we can do if the server fails to provide the schemas information.
- When the request fails, the UI already shows a nice error message with details about what went wrong.
Screenshot 2021-08-20 at 15 18 01

elements/diff/nuxeo-diff.js
```diff
@@ -468,28 +468,28 @@ Polymer({
...
-    // if both have the same type, only do a single fetch
-    if (left && right && left.type === right.type) {
-      return this._fetchSchemas(left).then((response) => response.schemas);
-    }
-    return this._fetchSchemas(left).then((response1) =>
-      this._fetchSchemas(right).then((response2) =>
-        response1.schemas.filter((schema1) => !!response2.schemas.find((schema2) => schema1.name === schema2.name)),
-      ),
+    const commonSchemas = left.schemas.filter(
+      (schema1) => !!right.schemas.find((schema2) => schema1.name === schema2.name),
```
Gabez0r, line 484: nit: if we are just turning this into a boolean, perhaps we could just `Array.prototype.includes`.

elements/diff/nuxeo-diff.js
```diff
@@ -468,28 +468,28 @@ Polymer({
...
     );
+    return this._fetchSchemas().then((schemas) => {
+      // populate the common schemas with the fields information
+      commonSchemas.forEach((commonSchema) => {
+        commonSchema.fields = schemas.find((schema) => schema.name === commonSchema.name).fields;
```
Gabez0r, line 489: Keep in mind that `Array.prototype.filter` creates a new array, but it does not do a deep copy, so we will likely be modifying the original schemas on the `schemas` property of the left document. I don't think this is an issue here, though.

elements/diff/nuxeo-diff.js
mnixo: Doesn't that work by reference? Wouldn't it be unsafe to assume that we have the same schema objects (not strings) on both sides?

elements/diff/nuxeo-diff.js
mnixo: Should be safe assuming that there is no manipulation going on. Or we can just force the deep copy, if need be.

elements/diff/nuxeo-diff.js
Gabez0r: You are right. I though it would accept a comparer function but it doesn't, so I guess it is only useful for string, boolean and numbers. Please ignore my comment.

Gabez0r (approved):

richardsd (commented):
Don't we need this [logic](https://github.com/nuxeo/nuxeo-web-ui/pull/1283/files#diff-c8196ee71265c5b2dbf29843e7122970b9269ea99fe523af9df93e0bcaa3defcR507) here?

mnixo:
> We don't need this type of [logic](https://github.com/nuxeo/nuxeo-web-ui/pull/1283/files#diff-c8196ee71265c5b2dbf29843e7122970b9269ea99fe523af9df93e0bcaa3defcR507) here?

We need to have 2 implementations for `10.10` because the `schemas` object was only made available after HF14. That is important because we know of instances where customers are using forked writers (pre HF14) that don't provide the `schemas` object. Thankfully, this is not a problem for LTS 2021 and `master`.

PR #1282, opened 2021-08-24
WEBUI-261: Rely on document specific schemas instead of type schemas for diff (3.0.x)

mnixo:
Backport of #1275.

### DIFF
--- a/elements/diff/nuxeo-diff.js
+++ b/elements/diff/nuxeo-diff.js
@@ -33,7 +33,7 @@ import { importHref } from '@nuxeo/nuxeo-ui-elements/import-href.js';
 import * as jsondiffpatch from 'jsondiffpatch/dist/jsondiffpatch.esm.js';
 
 let _customLoadPromise;
-const typeDataCache = {};
+let schemasCache;
 
 /**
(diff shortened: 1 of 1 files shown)


Gabez0r (approved):

richardsd (commented):
We don't need this type of [logic](https://github.com/nuxeo/nuxeo-web-ui/pull/1283/files#diff-c8196ee71265c5b2dbf29843e7122970b9269ea99fe523af9df93e0bcaa3defcR507) here?

mnixo:
> We don't need this type of [logic](https://github.com/nuxeo/nuxeo-web-ui/pull/1283/files#diff-c8196ee71265c5b2dbf29843e7122970b9269ea99fe523af9df93e0bcaa3defcR507) here?

We need to have 2 implementations for `10.10` because the `schemas` object was only made available after HF14. That is important because we know of instances where customers are using forked writers (pre HF14) that don't provide the `schemas` object. Thankfully, this is not a problem for LTS 2021 and `master`.

richardsd (approved):

semisse (approved):

PR #1283, opened 2021-08-25
WEBUI-261: Rely on document specific schemas instead of type schemas for diff (10.10)

mnixo:
Backport of #1275.

### DIFF
--- a/elements/diff/nuxeo-diff.html
+++ b/elements/diff/nuxeo-diff.html
@@ -231,6 +231,7 @@ limitations under the License.
     (function() {
       var _customLoadPromise;
       var schemasCache;
+      var typesCache = {};
       Polymer({
         is: 'nuxeo-diff',
         behaviors: [Nuxeo.I18nBehavior, Nuxeo.FiltersBehavior, Polymer.IronResizableBehavior],
@@ -436,6 +437,10 @@ limitations under the License.
           }
         },
 
+        /**
+         * Returns a list of all the available schemas and fields.
+         * After the first fetch, the list is cached.
+         */
         _fetchSchemas: function() {
           if (schemasCache) {
             return Promise.resolve(schemasCache);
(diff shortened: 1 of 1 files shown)


elements/diff/nuxeo-diff.html
```diff
@@ -436,34 +436,32 @@
...
+            return schemasCache;
           });
         },
 
         _fetchCommonSchemas: function(left, right) {
-          // if both have the same type, only do a single fetch
-          if (left && right && left.type === right.type) {
-            return this._fetchSchemas(left).then(function(response) {
-              return response.schemas;
+          var commonSchemas = left.schemas.filter(function(schema1) {
```
Gabez0r, line 451: On 10.10, `schemas` can be undefined if people are using a forked writer pre HF14.

mnixo: I propose:
- Null check `left.schemas` to ensure that scenario doesn't result in an uncontrolled failure.
- If caught by the check, return an empty list of schemas, resulting in an empty diff screen.
- Log the occurrence (browser console? elsewhere?) to ensure it can be picked up and fixed.

The alternative to that would be to avoid the use of the `schemas` object entirely at the cost of having to use a much more complicated and inefficient approach to determine the common schemas and fields (👎 ).

Gabez0r: Couldn't we just have a fallback on the old version of `_fetchCommonSchemas` if `left.schemas == null`?

richardsd (approved):

semisse (approved):

elements/diff/nuxeo-diff.html
```diff
@@ -436,36 +437,86 @@
...
         },
 
         _fetchCommonSchemas: function(left, right) {
-          // if both have the same type, only do a single fetch
-          if (left && right && left.type === right.type) {
-            return this._fetchSchemas(left).then(function(response) {
-              return response.schemas;
-            });
+          if (!left.schemas || !right.schemas) {
+            // the schemas objects might be undefined when using a (forked) writer pre HF14
```
Gabez0r, line 473: We could add the ticket for future reference
```suggestion
            // XXX: the schemas objects might be undefined when using a (forked) writer pre HF14 (see https://jira.nuxeo.com/browse/NXP-27995)
```

elements/diff/nuxeo-diff.html
```diff
@@ -436,36 +437,86 @@
...
+            return this._fetchCommonSchemasFromTypes(left, right);
```
Gabez0r, line 474: If this happens in production, and the client is expecting to get schemas from on dynamic facets, it would be interesting for us to know that this happened. It could help with support. We could probably use a console.warn here?

Gabez0r (commented):
Fallback solution looks great. Please add to the release notes of WEBUI-261 (and maybe on the comments too), that the solution will not work if people have custom Document JSON witters that do not expose the schemas property, and that for 10.10, will added a compat layer that won't be able to take into account dynamic facets.

mnixo:
Added a [comment on the ticket](https://jira.nuxeo.com/browse/WEBUI-261?focusedCommentId=567376&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-567376) regarding the fallback mechanism.

richardsd (approved):

Gabez0r (approved):

PR #1296, opened 2021-09-07
WEBUI-426: control the display of select all selection checkmark for each view

richardsd:


mnixo (dismissed):

elements/nuxeo-app.js
```diff
@@ -1116,6 +1116,7 @@ Polymer({
     if (!this._isEmpty(e.detail)) {
       this._toast(this.i18n(e.detail.docIds ? 'app.documents.addedToCollection' : 'app.document.addedToCollection'));
     }
+    this.fire('document-updated');
```
Gabez0r: Why do we need to fire `document-updated` here? Is it still related to WEBUI-426?

Gabez0r (commented):
Looking good! 👌

elements/nuxeo-app.js
richardsd: 💩 no, you're right, let me put that in another commit, because it is related with [WEBUI-427](https://jira.nuxeo.com/browse/WEBUI-427)

semisse (dismissed):

mnixo (dismissed):

semisse (dismissed):

andreacornaglia (dismissed):

elements/nuxeo-app.js
Gabez0r: Do we need to fire this here? This means that even actions that fire the `added-to-collection` event, such as `nuxeo-add-to-collection-button`, which is not bulk, will perform an additional update that is not required, right?

Gabez0r: Do we need to fire this here? This means that even actions that fire the added-to-collection event, such as nuxeo-add-to-collection-button, which is not bulk, will perform an additional update that is not required, right?

Gabez0r (commented):
Please see https://github.com/nuxeo/nuxeo-web-ui/pull/1296#discussion_r713931765

elements/nuxeo-app.js
richardsd: 🤔 For some reason, the `document-updated` listener is not triggered when bulk actions fire the event, which happens for any other action (not in bulk).

richardsd: Just dropped this. In fact it was not really needed to solve the problem. @Gabez0r thank you for really taking the time to question this.

Gabez0r (dismissed):

Gabez0r (dismissed):

semisse (dismissed):

semisse (approved):

Gabez0r (approved):

PR #1298, opened 2021-09-14
WEBUI-418: use search form visibility to display saved search actions

richardsd:
Problem: when clicking the **Save As** button, we rely on the search form to display the save dialog, however, when the search form is not visible we cannot display the dialog. 

Solution: as requested, we should hide the **Save As** button if we navigate away from the search form.

### DIFF
--- a/elements/search/nuxeo-saved-search-actions.js
+++ b/elements/search/nuxeo-saved-search-actions.js
@@ -52,14 +52,14 @@ Polymer({
     <div class="horizontal layout center">
       <paper-button
         on-tap="_saveSearchAs"
-        hidden$="[[!_showSaveAs(searchDoc, isSavedSearch, _dirty)]]"
+        hidden$="[[!_showSaveAs(searchDoc, isSavedSearch, _dirty, _isSearchFormVisible)]]"
         class="primary small"
       >
         <iron-icon icon="nuxeo:filter-edit"></iron-icon>[[i18n('app.saveNewSearch')]]
       </paper-button>
       <paper-button
         on-tap="_saveSearch"
-        hidden$="[[!_showSave(searchDoc, isSavedSearch, _dirty)]]"
+        hidden$="[[!_showSave(searchDoc, isSavedSearch, _dirty, _isSearchFormVisible)]]"
         class="secondary small"
       >
         <iron-icon icon="nuxeo:filter-add"></iron-icon>[[i18n('app.savedSearch')]]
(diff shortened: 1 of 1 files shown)


elements/search/nuxeo-saved-search-actions.js
```diff
@@ -101,9 +101,16 @@ Polymer({
...
     _dirty: Boolean,
+    _isSearchFormVisible: Boolean,
   },
 
-  _searchFormChanged() {
+  ready() {
+    this._searchFormVisibilityChanged = (event) => {
+      this._isSearchFormVisible = event.target.visible;
+    };
+  },
```
mnixo, line 111: Can we get away with not having to add a `ready` method just for this or is it strictly necessary to do it like this? (could probably mess up the `this` inside?)
```suggestion
  _searchFormVisibilityChanged(event) {
    this._isSearchFormVisible = event.target.visible;
  }
```

mnixo (dismissed):

elements/search/nuxeo-saved-search-actions.js
richardsd: Ah yes, you are absolutely correct. I tried that, but the problem is the `this` inside. With that we would end up having to use the `bind(this)` which would create a new function. 
That said, my first approach was to define the method inside the `searchFormChanged`, wdyt?

elements/search/nuxeo-saved-search-actions.js
```diff
@@ -116,6 +123,13 @@ Polymer({
         }
       });
 
+      if (oldForm) {
```
richardsd, line 126: What I had previously was something like:

```
if (oldForm && this._searchFormVisibilityChanged) {
  oldForm.removeEventListener('visible-changed', this._searchFormVisibilityChanged);
}

this._searchFormVisibilityChanged = (event) => {
  this._isSearchFormVisible = event.target.visible;
};

if (form) {
  form.addEventListener('visible-changed', this._searchFormVisibilityChanged);
  this._isSearchFormVisible = form.visible;
}
```

My main problem with this approach is that I'm constantly defining the the `this._searchFormVisibilityChanged` method. I could hide behind a condition block, but don't know what the best option is.

richardsd: @mnixo

mnixo: Understood. The current approach seems to be the cleanest.

richardsd: Thanks for the feedback 👍

semisse (dismissed):

semisse (approved):

mnixo (approved):

PR #1299, opened 2021-09-14
WEBUI-222: bump to TP 2021.9

guirenard:


semisse (dismissed):

richardsd (dismissed):

semisse (dismissed):

pom.xml
```diff
@@ -3,7 +3,7 @@
   <parent>
     <groupId>org.nuxeo</groupId>
     <artifactId>nuxeo-parent</artifactId>
-    <version>2021.6</version>
+    <version>2021.7</version>
```
richardsd: Couldn't we set the version to **2021.8** (at least until the 2021.9 is promoted)?

package.json
```diff
@@ -180,6 +180,7 @@
     "dotenv": "^8.0.0",
     "es6-promise": "^4.0.0",
     "extend": "^3.0.2",
+    "inquirer": "8.1.2",
```
richardsd: Why do we need to add this dependency? Do you know if it was coming from another package already?

richardsd (dismissed):

package.json
Gabez0r: What do we need this package for?

guirenard: Please see https://jira.nuxeo.com/browse/WEBUI-474 referenced in the associated commit -> https://github.com/SBoudrias/Inquirer.js/issues/1046
It is now fixed and they released a newer version but at the time I couldn't get the build passing. I am gonna revert this commit now

semisse (approved):

doulba (approved):

PR #1307, opened 2021-10-04
WEBUI-188: move to github actions

richardsd:


.github/workflows/main.yaml
```diff
@@ -0,0 +1,139 @@
+name: Build
+
+on:
+  push:
+    branches:
+      - maintenance-3.0.x
+  pull_request:
```
nelsonsilva, line 7: I think we need to limit the PR to the `maintenance-3.0.x` base branch too, right ?

.github/workflows/main.yaml
```diff
@@ -0,0 +1,139 @@
...
+    name: Packages
+    runs-on: self-hosted
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: actions/setup-java@v1
+      with:
+        java-version: 11
+
+    - name: Cache local Maven repository
```
nelsonsilva: question: how does the caching work for self hosted runners ? 🤔

.github/workflows/main.yaml
```diff
@@ -0,0 +1,139 @@
...
+  CONNECT_URL: https://nos-preprod-connect.nuxeocloud.com/nuxeo
+
+jobs:
+  build:
+    name: Packages
+    runs-on: self-hosted
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: actions/setup-java@v1
```
nelsonsilva: does it make sense to setup java and node versions when this is a self hosted runner running an image we built with the right tools already in place ?

.github/workflows/main.yaml
```diff
@@ -0,0 +1,139 @@
...
+    - name: Archive cucumber reports
+      if: always()
+      uses: actions/upload-artifact@v2
+      with:
+        name: cucumber-reports
+        path: |
+          ftest/target/cucumber-reports/
+
+    - name: Archive logs
+      if: always()
```
nelsonsilva: forgotten if for debugging ?

.github/workflows/main.yaml
```diff
@@ -0,0 +1,139 @@
...
+      run: mvn install -DskipInstall
+
+    - name: Functional tests
+      run: |
+        mvn -B -nsu -f plugin/itests/addon install
+        mvn -B -nsu -f plugin/itests/marketplace install
+        mvn -B -nsu -f ftest install
+
+    - name: Archive cucumber reports
+      if: always()
```
nelsonsilva: forgotten if for debugging ?

.github/workflows/main.yaml
richardsd: Yup, you're totally right.

.github/workflows/main.yaml
richardsd: Probably not, since as you pointed out https://github.com/nuxeo/nuxeo-web-ui-gh-runner/blob/WEBUI-188-automate-runner-actions/Dockerfile#L39 is already defining jdk11.

.github/workflows/main.yaml
richardsd: Looking at the pods, and since we are the ones hosting the runners, the dependencies are stored by us locally (differently from what happens with GH runnes see [here](https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-java-with-maven#caching-dependencies)). 
In the end, when the pods are destroyed, because of our strategy for horizontal scaling, I'm assuming that we lose the cache (except for the one that doesn't get destroyed).

richardsd: Is this what you were thinking? It seems that maybe caching won't work or us :/

richardsd: (Btw, according to GH documentation `When using GitHub-hosted runners, you can cache your dependencies to speed up your workflow runs.`)

.github/workflows/main.yaml
richardsd: Well, it seems that either we set up the default versions directly in the runner container, or if we don't use the action to change the version we end up with these errors: https://github.com/nuxeo/nuxeo-web-ui/pull/1307/checks?check_run_id=3904762833#step:8:2370

richardsd: Done. We don't need this anymore. Thanks for the feedback 🙂

.github/workflows/main.yaml
```diff
@@ -0,0 +1,134 @@
...
+
+jobs:
+  build:
+    name: Packages
+    runs-on: self-hosted
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Cache local Maven repository
+      uses: actions/cache@v2
```
Gabez0r: Do we even need to cache here? Do we know if our pods will live long enough for it to be used?

.github/workflows/preview.yaml
```diff
@@ -0,0 +1,121 @@
...
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled]
+    branches:
+      - maintenance-3.0.x
+
+env:
+  DOCKER_REGISTRY: us-east1-docker.pkg.dev
+  DOCKER_IMAGE: ${{ secrets.GKE_PROJECT }}/webui/nuxeo-web-ui
```
Gabez0r, line 11: We have a variable for this, which is only used once, but the server image is hardcoded further down.

Gabez0r (dismissed):

.github/workflows/main.yaml
richardsd: I don't think so. Since we are relying on the hpa to create/delete pods when necessary, only if the pod remains up this would be valuable.

.github/workflows/preview.yaml
richardsd: The variable is going to be used [here](https://github.com/nuxeo/nuxeo-web-ui/blob/fix-WEBUI-188-github-actions-3.0.x/charts/preview/values.yaml#L38) as well, because of this [step](https://github.com/nuxeo/nuxeo-web-ui/blob/fix-WEBUI-188-github-actions-3.0.x/.github/workflows/preview.yaml#L119-L121).

.github/workflows/preview.yaml
```diff
@@ -0,0 +1,141 @@
...
+      id: preview_ns
+      run: |
+        echo ::set-output name=exists::$(kubectl get namespace ${PREVIEW_NAMESPACE})
+
+    # Previous preview deployment needs to be scaled to 0 to be replaced correctly
+    # - name: Cleanup preview namespace
+    #   if: ${{ steps.preview_ns.outputs.exists }}
+    #   continue-on-error: true
+    #   run: |
+    #     kubectl -n ${PREVIEW_NAMESPACE} scale deployment preview --replicas=0
```
semisse: do we want to keep this code?

semisse (dismissed):

.github/workflows/preview.yaml
richardsd: Yet to be decided. I still need to confirm that for every push into the PR/branch, triggers an update in the preview image being pulled.

richardsd: Sam, trying to use this strategy to scale down the deployment and make sure it pulls the new image https://github.com/nuxeo/nuxeo-web-ui/pull/1307/files#diff-22da830e02ef5bed54ecad981722fd9c940aec797740c09927d4a904df0bccd6R122-R123.

Although we build a new image every time we push to the branch, because the version/tag is still the same, the pod is not recreated/updated, even though we set the `imagePullPolicy` to always.

This code actually solves the problem, but I could try setting a new tag/version for the preview image.

semisse (dismissed):

semisse (approved):

.github/workflows/main.yaml
```diff
@@ -0,0 +1,133 @@
...
+    - name: Archive packages
+      uses: actions/upload-artifact@v2
+      with:
+        name: packages
+        path: |
+          plugin/web-ui/marketplace/target/nuxeo-web-ui-marketplace-*.zip
+          plugin/itests/marketplace/target/nuxeo-web-ui-marketplace-itests-*.zip
+
+    - name: Tag
+      if: github.event_name == 'push'
```
Gabez0r, line 112: If by any change this fails and we retry, is `github.event_name` still `push`?

Gabez0r (approved):

.github/workflows/main.yaml
richardsd: It is a good question, let me look in the docs. I'll ping you when I have more details 
(🤔 if not we might need to review the step)

richardsd: Yes, so if the job fails while doing a push and you retry, the `github.event_name` is still `push` (I tested it in another repo - see [here](https://github.com/richardsd/gh_actions_tests/actions/runs/1416046316/attempts/1)).

mnixo (approved):

PR #1312, opened 2021-10-19
WEBUI-488: update Safari to v15 in Saucelabs

richardsd:
Notes: 
- Updated the MacOSx version to Big Sur (once Saucelabs update the Bigsur's Safari version, we will use it)

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -38,7 +38,7 @@ if (process.env.SAUCE_USERNAME && process.env.SAUCE_ACCESS_KEY) {
     sl_latest_safari: {
       base: 'SauceLabs',
       browserName: 'safari',
-      platform: 'macOS 10.13',
+      platform: 'macOS 11.00',
       version: 'latest',
     },
   };


karma.conf.js
```diff
@@ -38,8 +38,8 @@ if (process.env.SAUCE_USERNAME && process.env.SAUCE_ACCESS_KEY) {
     sl_latest_safari: {
       base: 'SauceLabs',
       browserName: 'safari',
-      platform: 'macOS 10.13',
-      version: 'latest',
+      platform: 'macOS 11.00',
+      version: 'latest', // TODO - check if v15 is available
```
Gabez0r: If we merge it like this, we would eventually just get v15 once it is available, right?

richardsd: For sure. We just ned to update the platform version, and once they update the platform on their side, we'll get the v15 seamlessly.

richardsd: I propose to remove the comment and close the ticket merging the PR, wdyt?

Gabez0r: I agree!

richardsd: Proceeding with that, thanks.

Gabez0r (approved):

mnixo (approved):

PR #1316, opened 2021-10-22
WEBUI-388/WEBUI-389/WEBUI-390: Bulk Edit (3.0.x)

mnixo:
Pairs with [nuxeo-elements/pull/460](https://github.com/nuxeo/nuxeo-elements/pull/460).

~Please disregard the changes on [the first commit](https://github.com/nuxeo/nuxeo-web-ui/pull/1316/commits/51a16a43506d6a728c9755c9f0edc52db17fc7d6) (added to enable Select All locally).~
~Commits identified with `[squash]` will eventually be squashed before merging.~

Currently missing:
* Functional tests.


elements/nuxeo-web-ui-bundle.html
```diff
@@ -790,6 +790,7 @@
 </nuxeo-slot-content>
 
 <!-- PUBLISHING -->
+
```
richardsd, line 801: You're adding empty lines here.

elements/nuxeo-web-ui-bundle.html
```diff
@@ -812,8 +813,21 @@
 </nuxeo-slot-content>
 
 <!-- SPREADSHEET-->
+
```
richardsd, line 824: Same here.

i18n/messages.json
```diff
@@ -182,6 +182,9 @@
   "browser.publications": "Publishing",
   "browser.trash": "Trash",
   "browser.view": "View",
+  "bulkWidget.error.replaceWithEmpty": "Please enter the value to replace this field with",
```
richardsd, line 185: Please don't forget to create the ticket to handle the translation of the labels.

elements/nuxeo-document-bulk-actions/nuxeo-bulk-widget.js
```diff
@@ -0,0 +1,192 @@
...
+          width: 180px;
+        }
+        paper-item[disabled] {
+          @apply --nx-button-text-disabled;
+        }
+      </style>
+      <div class="header">
+        <label>[[label]]</label>
+        <nuxeo-select selected="{{updateMode}}" attr-for-selected="id">
+          <paper-item id="keep" label="Don't change value(s)">Don't change value(s)</paper-item>
```
richardsd: Should the labels for the three paper-items be i18n too?

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
```diff
@@ -0,0 +1,501 @@
...
+    `;
+  }
+
+  static get properties() {
+    return {
+      /**
+       * The representation of the selected documents. Can be a list of documents or a view (for select all).
+       */
+      documents: {
+        type: Array,
```
richardsd: Shouldn't this be a more generic type, like object? Or in the end it doesn't really matter? Asking because I've seen we using the same strategy somewhere 🤔

elements/nuxeo-document-bulk-actions/nuxeo-bulk-widget.js
```diff
@@ -0,0 +1,192 @@
...
+          --paper-input-container-input: {
+            font-size: 12px;
+            text-align: right;
+          }
+          /* tweak text inside paper-items */
+          --paper-item: {
+            font-size: 12px;
+          }
+          /* set the width to make sure the text of the options doesn't wrap */
+          width: 180px;
```
Gabez0r: Is this doing anything?

elements/nuxeo-document-bulk-actions/nuxeo-bulk-widget.js
```diff
@@ -0,0 +1,192 @@
...
+
+  /**
+   * Dispatches an event upwards when `updateMode` changes, containing `this` instance.
+   */
+  _updateModeChanged() {
+    this.dispatchEvent(
+      new CustomEvent('update-mode-changed', {
+        bubbles: true,
+        composed: true,
+        detail: this,
```
Gabez0r: We could consider sending the widget as a property inside the details in case we later want to add more information to the event.

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
```diff
@@ -0,0 +1,501 @@
...
+        if (el.tagName.toLowerCase() === 'nuxeo-bulk-widget') {
+          // validate the element as bulk widget
+          // if the bulk widget is not tagged with a `boundPath`, it's safe to assume it has no value
+          const value = el.boundPath ? this._model[el.boundPath] : null;
+          // check if trying to replace with empty value
+          const replacingWithEmptyValue = el.updateMode === 'replace' && this._isValueEmpty(value);
+          if (replacingWithEmptyValue) {
+            el._setError(this.i18n('bulkWidget.error.replaceWithEmpty'));
+          }
+          valid = !replacingWithEmptyValue && valid;
```
Gabez0r: nit:
```suggestion
          valid &&= !replacingWithEmptyValue;
```
(assuming it works with linting)

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
```diff
@@ -0,0 +1,501 @@
...
+        } else {
+          // validate the element as any other widget
+          valid = (el.validate ? el.validate() : el.checkValidity()) && valid;
```
Gabez0r: nit:
```suggestion
          valid &&= el.validate ? el.validate() : el.checkValidity();
```
(assuming it works with linting)

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
```diff
@@ -0,0 +1,501 @@
...
+    }
+    let properties = '';
+    // go through each of the bound paths that exist in the `_model`
+    Object.keys(this._model).forEach((boundPath) => {
+      // get the bulk widget wrapper element
+      const bulkWidget = this._getWidgetForBoundPath(boundPath).parentNode;
+      // get the path without the `document.properties.` part
+      const path = boundPath.replace(/^(document\.properties\.)/, '');
+      if (bulkWidget.updateMode === 'replace') {
+        properties = `${properties}${path}=${this._model[boundPath]}\n`;
```
Gabez0r: I think this works fine if `this._model[boundPath]}` is a simple types or ar array of simple types, but if it is an object or an array of objects, we will likely get a bunch of `[object Object]`. Could you please confirm? If this is the case, we will have to do something like a `JSON.stringify`, for example. Perhaps we could even always stringify it?

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
```diff
@@ -0,0 +1,501 @@
...
+   * Loads the bulk edit layout.
+   */
+  _loadLayout(documents, layout, hrefFunction, hrefBase) {
+    // force layout restamp
+    this._set_href(null);
+    const base =
+      hrefBase ||
+      // XXX Since this button is one level deeper than the directory of the layouts (`bulk/`), we need to remove this
+      // part of the path to be able to locate them. Ideally we wouldn't need to do this manipulation.
+      pathFromUrl(this.__dataHost.importPath || this.importPath).replace(/(nuxeo-document-bulk-actions\/)$/, '');
```
Gabez0r: This is indeed a problem, that seems to be related to https://github.com/nuxeo/nuxeo-elements/pull/427, right? If so, we should find a common strategy for this. 🤔

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
```diff
@@ -0,0 +1,501 @@
...
+    // initialize the model
+    this._model = {};
+    // fill in the model with the properties that we can already know that exist in the layout
+    layout.element.__templateInfo.nodeInfoList.forEach((node) => {
+      node.bindings
+        .filter((binding) => binding.kind === 'property')
+        .forEach((binding) => {
+          binding.parts
+            .filter((part) => part.mode === '{' && part.source.startsWith('document.properties.'))
+            .filter((part) => !this._model[part.source])
```
Gabez0r: I think this won't filter out a property that is initialized with null. We could check instead if the key exists?

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
```diff
@@ -0,0 +1,501 @@
...
+    // set the observer in the layout so that when a property gets updated, the callback is executed
+    layout.element._createMethodObserver('_propertiesObserver(document.properties.*)', true);
+    // replace all the widgets in the layout with bulk widget wrappers
+    const widgets = Array.from(layout.element.shadowRoot.querySelectorAll('[role="widget"]'));
+    widgets.forEach((widget) => {
+      const { parentNode } = widget;
+      const bulkWidget = document.createElement('nuxeo-bulk-widget');
+      parentNode.replaceChild(bulkWidget, widget);
+      bulkWidget.appendChild(widget);
+      bulkWidget.element = widget;
```
Gabez0r: Where is this `element` used?

elements/nuxeo-document-bulk-actions/nuxeo-bulk-widget.js
```diff
@@ -0,0 +1,192 @@
...
+limitations under the License.
+*/
+import { mixinBehaviors } from '@polymer/polymer/lib/legacy/class.js';
+import '@polymer/paper-item/paper-item.js';
+import { I18nBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-i18n-behavior.js';
+import '@nuxeo/nuxeo-ui-elements/widgets/nuxeo-select.js';
+import { html } from '@polymer/polymer/lib/utils/html-tag.js';
+
+/**
+`nuxeo-bulk-widget`
```
Gabez0r: We should document this element and say that it is only for internal use of the `nuxeo-edit-documents-button`.

i18n/messages.json
```diff
@@ -398,6 +401,7 @@
   "documentEdit.layoutNotFound": "Failed to find edit layout for {0}.",
   "documentEdit.saveError": "Failed to save.",
   "documentEditAction.tooltip": "Edit document",
+  "documentEditBulkAction.tooltip": "Bulk Edit documents",
```
Gabez0r, line 408: Just a reminder that we need to add the `translation` tag to the respective ticket and that, once merged, we need to create the respective NXDOC ticket, so that we can get the translations.

elements/nuxeo-document-bulk-actions/nuxeo-bulk-widget.js
richardsd: 🤔 additionally, this element is going to appear in Designer's catalog by default

elements/nuxeo-document-bulk-actions/nuxeo-bulk-widget.js
richardsd: ➕1 on what Gabe said, could be useful

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
```diff
@@ -0,0 +1,501 @@
...
+    this.operation = 'Document.Update';
+    // set up a listener to act on changes of the update modes of the bulk widget wrappers in the layout
+    this.addEventListener('update-mode-changed', (e) => this._updateBulkWidget(e.detail));
+  }
+
+  /**
+   * Control the visibility of the button.
+   */
+  _isHidden() {
+    // TODO compare to the _isHidden on nuxeo-add-to-collections-documents-button
```
richardsd: The only difference I see is the `this.documents.every((doc) => !this.hasFacet(doc, 'NotCollectionMember'))` that I don't think we need here.

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
richardsd: 🤔 It is not part of the public API of the element

elements/nuxeo-document-bulk-actions/nuxeo-bulk-widget.js
```diff
@@ -0,0 +1,192 @@
...
+          align-items: center;
+          justify-content: space-between;
+        }
+        .warning,
+        .error {
+          /* ensures that the bulk widget takes the same vertical space regardless of having a message or not */
+          min-height: 20px;
+        }
+        .warning {
+          color: var(--nuxeo-text-default);
```
semisse: ideally we should have default values in every single CSS variable

elements/nuxeo-document-bulk-actions/nuxeo-bulk-widget.js
```diff
@@ -0,0 +1,192 @@
...
+            display: none;
+          }
+          /* remove padding on the paper-dropdown-menu */
+          --nuxeo-select-dropdown-menu-padding: 0;
+          /* remove padding on the paper-input */
+          --nuxeo-select-input-container-padding: 0;
+          /* tweak text inside the paper-input element */
+          --paper-input-container-input: {
+            font-size: 12px;
+            text-align: right;
```
semisse:
```suggestion
            text-align: end;
```

semisse: This suggestion makes this particular styling compatible with RTL.

elements/nuxeo-document-bulk-actions/nuxeo-bulk-widget.js
```diff
@@ -0,0 +1,192 @@
...
+          /* set the width to make sure the text of the options doesn't wrap */
+          width: 180px;
+        }
+        paper-item[disabled] {
+          @apply --nx-button-text-disabled;
+        }
+      </style>
+      <div class="header">
+        <label>[[label]]</label>
+        <nuxeo-select selected="{{updateMode}}" attr-for-selected="id">
```
semisse:
```suggestion
        <label for="options">[[label]]</label>
        <nuxeo-select selected="{{updateMode}}" attr-for-selected="id" id="options">
```

semisse: For better a11y

elements/nuxeo-document-bulk-actions/nuxeo-bulk-widget.js
```diff
@@ -0,0 +1,192 @@
...
+          <paper-item id="keep" label="Don't change value(s)">Don't change value(s)</paper-item>
+          <paper-item id="replace" label="Replace with">Replace with</paper-item>
+          <paper-item id="remove" label="Remove value(s)" disabled$="[[_required]]">Remove value(s)</paper-item>
+        </nuxeo-select>
+      </div>
+      <slot></slot>
+      <label class$="[[_messageClass]]">[[_message]]</label>
```
semisse: We should consider to wrap these form inputs in a fieldset element, in order for this label to make sense in this context

semisse: Label requires always a for attribute. In this case, we need to replace this element with a simple `span`
```suggestion
      <span class$="[[_messageClass]]">[[_message]]</span>
```

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
```diff
@@ -0,0 +1,501 @@
...
+
+  static get importMeta() {
+    return import.meta;
+  }
+
+  static get template() {
+    return html`
+      <style include="nuxeo-action-button-styles nuxeo-styles">
+        nuxeo-dialog {
+          height: 100%;
```
semisse: I'm wondering if this line is actually doing anything here 🤔

semisse (dismissed):

elements/nuxeo-document-bulk-actions/nuxeo-bulk-widget.js
mnixo: I'm sure that this was doing something at some point... Apparently, not anymore 👍

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
mnixo: It is used in the `_updateBulkWidget` method below, so that this button can know which widget to enable/disable according to mode changes on the bulk widget that wraps it.

elements/nuxeo-web-ui-bundle.html
mnixo: This line was added because this was one of the 2 entries that didn't have the line between the comment and the contribution in the whole bundle file, so it normalizes the format.

elements/nuxeo-web-ui-bundle.html
mnixo: This line was added because this was one of the 2 entries that didn't have the line between the comment and the contribution in the whole bundle file, so it normalizes the format.

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
mnixo: I checked the existing bulk action elements and while the majority uses `Array` for `documents`:
- `<nuxeo-add-to-collection-documents-button>`
- `<nuxeo-clipboard-documents-button>`
- `<nuxeo-download-documents-button>`

A couple use `Object` instead:
- `<nuxeo-delete-documents-button>`
- `<nuxeo-untrash-documents-button>`

So I picked `Array`. But ultimately I don't think it makes any relevant difference.

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
mnixo: https://user-images.githubusercontent.com/4223240/140541165-07a19b65-3598-4b86-961f-11420993bc21.mp4

i18n/messages.json
mnixo: Added the `translation` tag on the following tickets:
- [WEBUI-388: Add Bulk Edit widget wrapper element](https://jira.nuxeo.com/browse/WEBUI-388)
- [WEBUI-390: Add Bulk Edit default action](https://jira.nuxeo.com/browse/WEBUI-390)

Also created: [NXDOC-2393: Translate Bulk Edit labels](https://jira.nuxeo.com/browse/NXDOC-2393).

i18n/messages.json
mnixo: Added the `translation` tag on the following tickets:
- [WEBUI-388: Add Bulk Edit widget wrapper element](https://jira.nuxeo.com/browse/WEBUI-388)
- [WEBUI-390: Add Bulk Edit default action](https://jira.nuxeo.com/browse/WEBUI-390)

Also created: [NXDOC-2393: Translate Bulk Edit labels](https://jira.nuxeo.com/browse/NXDOC-2393).

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
mnixo:
```
error [parse-error] - This experimental syntax requires enabling the parser plugin: 'logicalAssignment'
```
🤷

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
mnixo:
```
error [parse-error] - This experimental syntax requires enabling the parser plugin: 'logicalAssignment'
```
🤷

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
mnixo: Yes, that is correct. In the meantime this mechanism has changed entirely but we ended up going with an approach that doesn't stringify any of the values, unless the value is an array of objects (`blob[]` & `complex[]`). More information about this can be found in [this JIRA ticket comment](https://jira.nuxeo.com/browse/WEBUI-390?focusedCommentId=582936&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-582936).

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
mnixo: This `filter` got removed because it ended up not serving any purpose with the new mechanism.

elements/nuxeo-document-bulk-actions/nuxeo-bulk-widget.js
mnixo: Done!

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
```diff
@@ -0,0 +1,613 @@
...
+   * Flattens the provided `data` structure (that can have multiple levels) into a single map-like structure, mapping
+   * bound paths to values. Accepts a `prefix` string to be added on all keys of the map. A `flattenedProperties`
+   * parameter can also be passed so that the entries are added to it incrementally.
+   */
+  _flattenProperties(data, prefix, flattenedProperties = {}) {
+    // XXX Avoid flattening batch upload blob values (for blob properties)
+    if (data['upload-batch'] && data['upload-fileId']) {
+      flattenedProperties[prefix] = data;
+      return flattenedProperties;
+    }
```
Gabez0r: This is something we could improve IF we got the filed information from the server, right? If so, we could mention it on the XXX, so we don't forget to refactor it.

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
```diff
@@ -0,0 +1,613 @@
...
+
+  /**
+   * Flattens the provided `data` structure (that can have multiple levels) into a single map-like structure, mapping
+   * bound paths to values. Accepts a `prefix` string to be added on all keys of the map. A `flattenedProperties`
+   * parameter can also be passed so that the entries are added to it incrementally.
+   */
+  _flattenProperties(data, prefix, flattenedProperties = {}) {
+    // XXX Avoid flattening batch upload blob values (for blob properties)
+    if (data['upload-batch'] && data['upload-fileId']) {
+      flattenedProperties[prefix] = data;
```
Gabez0r: nit: The `prefix` parameter is actually the current level path, right? Found the name a bit misleading.

Also, is it possible that this case can repeat in other situations? What happens if `data` is not an `Object` nor an `Array`? Can this happen?

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
```diff
@@ -0,0 +1,613 @@
...
+  }
+
+  /**
+   * Returns a boolean value according to whether the provided property `value` should be stringified or not.
+   */
+  _shouldStringifyValue(value) {
+    // check if value is an array
+    if (Array.isArray(value)) {
+      // should stringify the array if all the entries in it are objects
+      return value.every((entry) => entry instanceof Object);
```
Gabez0r: If we have an array with mixed objects and primitive types I guess we won't know what to do, right? Perhaps this is not even possible or expected? If so, using `some` instead would be more efficient I think:
```suggestion
      return value.some((entry) => entry instanceof Object);
```

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
```diff
@@ -0,0 +1,613 @@
...
+          .filter((binding) => binding.kind === 'property')
+          .forEach((binding) => {
+            binding.parts
+              .filter((part) => part.mode === '{' && part.source.startsWith('document.properties.'))
+              .forEach((part) => {
+                const sourceParts = part.source.split('.');
+                sourceParts.pop();
+                // create the path for each bound property (necessary for complex properties)
+                createNestedObject(bulkLayout, sourceParts);
+                bulkLayout.set(part.source, null);
```
Gabez0r: nit: This is where we would plug the schema information from the server, right? If so, it would be worth mentioning in a comment.

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
```diff
@@ -0,0 +1,613 @@
...
+  _propertiesObserver(detail) {
+    const bulkLayout = this.$$('nuxeo-layout').element;
+    if (detail.path === 'document.properties') {
+      // properties object was reset, do nothing
+    } else if (detail.path.endsWith('.length') && Array.isArray(bulkLayout.get(detail.path.replace(/\.length$/, '')))) {
+      // path ends in `.length` and the path without it matches an array property, do nothing
+    } else if (detail.path.match(/\.\d+$/) && Array.isArray(bulkLayout.get(detail.path.replace(/\.\d+$/, '')))) {
+      // path ends in `.[\d]` and the path without it matches an array property, do nothing
+    } else if (detail.path.startsWith('document.properties')) {
+      const boundPath = detail.path.endsWith('.splices') ? detail.path.replace('.splices', '') : detail.path;
```
Gabez0r: To safeguard from a case where we can actually have a property named `splices`, perhaps we could also check the `value` for a property named `indexSplices`?

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
Gabez0r: Wouldn't it make our lives easier if we moved `nuxeo-edit-documents-button` and `nuxeo-bulk-widget` to the `bulk` folder?

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
```diff
@@ -0,0 +1,613 @@
...
+
+  /**
+   * Gets the bulk widget that is structurally above the provided `node`, recursively.
+   */
+  _getBulkWidget(node) {
+    if (!node.parentNode) {
+      return null;
+    }
+    const parent = node.parentNode;
+    return parent.tagName.toLowerCase() === 'nuxeo-bulk-widget' ? parent : this._getBulkWidget(parent);
```
Gabez0r: Perhaps we could easily have an iterative solution here, which should be a bit more efficient?

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
```diff
@@ -0,0 +1,613 @@
...
+
+  /**
+   * Gets the element bound to a property that resides inside the provided `node` structure, recursively.
+   */
+  _getBoundElement(node, layoutTemplateInfo) {
+    // if there's no boundNodes information yet, create it by finding all the nodes that have bindings
+    if (!layoutTemplateInfo.boundNodes) {
+      layoutTemplateInfo.boundNodes = [];
+      for (let i = 0; i < layoutTemplateInfo.nodeList.length; i++) {
+        if (layoutTemplateInfo.nodeInfoList[i].bindings) {
```
Gabez0r: nit: we could save `layoutTemplateInfo.nodeInfoList[i]` in a variable, since we are using it twice.

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
```diff
@@ -0,0 +1,613 @@
...
+    // XXX The data table selector is needed because it sets `role="table"` on itself (see ELEMENTS-1346)
+    const selector = '[role="widget"], nuxeo-data-table[role="table"]';
+    const widgets = Array.from(bulkLayout.shadowRoot.querySelectorAll(selector));
+    widgets.forEach((widget) => {
+      const { parentNode } = widget;
+      const bulkWidget = document.createElement('nuxeo-bulk-widget');
+      parentNode.replaceChild(bulkWidget, widget);
+      bulkWidget.appendChild(widget);
+      // get the element that is bound to a property
+      const boundElement = this._getBoundElement(widget, bulkLayout.__templateInfo);
```
Gabez0r: It is a bit obscure to me, just by looking at the code, why we need such an elaborate `_getBoundElement`. I believe this might be related to less trivial cases, where the widgets represent divs (and possibly something else?) I think we should document this.

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
```diff
@@ -0,0 +1,613 @@
...
+    }
+    const bulkLayout = this.$$('nuxeo-layout').element;
+    const value = bulkLayout.get(boundPath);
+    if (Array.isArray(value)) {
+      if (value.length !== 0) {
+        bulkLayout.set(boundPath, []);
+      }
+    } else {
+      bulkLayout.set(boundPath, null);
+    }
```
Gabez0r: Again, this is something that we should update once we get schema information from the server, right?

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
```diff
@@ -0,0 +1,613 @@
...
+      return;
+    }
+    if ('disabled' in widget) {
+      widget.disabled = true;
+    } else if ('readonly' in widget) {
+      widget.readonly = true;
+    } else if ('editable' in widget) {
+      widget.editable = false;
+    }
+  }
```
Gabez0r: nit: We could have instead a single helper to set the "disabled" state to prevent duplicating a bunch of ifs, right? E.g.:
```suggestion
  _setWidgetDisable(widget, value) {
    if (!widget) {
      return;
    }
    if ('disabled' in widget) {
      widget.disabled = !!value;
    } else if ('readonly' in widget) {
      widget.readonly = !!value;
    } else if ('editable' in widget) {
      widget.editable = !value;
    }
  }
```

Gabez0r (commented):
Great job! We are almost there! 👏 💪

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
mnixo: We can rename `prefix` if it makes it easier to understand. Yes, it represents the current (or base) path, so I'll try using `currentPath` instead.
As to the other question, `_flattenProperties` is currently only being called in two scenarios:
- In the `_save` method, when `data` is `bulkLayout.document.properties`, which should be at least an empty object.
- Inside `flattenProperties` itself, but only when we know we're dealing with an `Object` that is not an `Array`.

Based on this I don't think we have reason to believe that `data` can be a non `Object`, but let me know if you think we should handle that case either way.

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
mnixo: I don't think there's a scenario where an `Array` can have entries of more than one data type, so yes, it really doesn't make sense to check every single entry.

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
mnixo: Absolutely, this was a slip up. Even the `replace` call should be only replacing the end of the string.

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
mnixo: Sure, we can go with something like this instead:
```
_getBulkWidget(node) {
  let current = node;
  while (current && current.tagName.toLowerCase() !== 'nuxeo-bulk-widget') {
    current = current.parentNode;
  }
  return current;
}
```
I assumed the depth of the DOM structure wouldn't be deep enough to blow up the call stack size 😅

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
mnixo: In this case one is from `nodeInfoList` and the other is from `nodeList`.

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
mnixo: That's exactly it. There's a few (unusual) concepts at hand here:
- Widget: The element tagged with `role="widget"`. It can be anything.
- Bulk Widget: The injected element that wraps around the widget element.
- Bound Element: The element that is bound to a property. Could be the widget element or could live somewhere inside the widget element.

I tried to expand a bit on this a few lines above:
```
// replace all the widgets in the layout with bulk widget wrappers
// a widget is a node identified with the `role="widget"` attribute, it can be something as simple an input element
// bound to a property, or a complex DOM structure with an element somewhere inside bound to a property
```

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
mnixo: Yes, adding a comment.

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
mnixo: Yes, we can.

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
mnixo: It would mean that we wouldn't need to do this trimming, but it would also make its path inconsistent with the other bulk actions paths. Would that be worth it?

richardsd (dismissed):

index.html
```diff
@@ -126,9 +126,9 @@
             }, */
             // pageprovider: { method: 'post', }, // uncomment to enable search using POST (see NXP-29126)
             // router: { key: { document: 'uid', }, }, // uncomment to enable navigation by UID (see WEBUI-14)
-            /* selection: { // uncomment to enable select all items in listings
+            selection: { // uncomment to enable select all items in listings
```
Gabez0r: Don't forget to drop this change.

elements/nuxeo-web-ui-bundle.html
```diff
@@ -705,6 +705,14 @@
   </template>
 </nuxeo-slot-content>
 
+<!-- BULK EDIT -->
+
+<nuxeo-slot-content name="bulkEditResultsAction" slot="RESULTS_SELECTION_ACTIONS" order="10">
```
Gabez0r: Don't forget to disable the feature:
```suggestion
<nuxeo-slot-content name="bulkEditResultsAction" slot="RESULTS_SELECTION_ACTIONS" order="10" disabled>
```

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
Gabez0r: So, the question here is, do we prefer to:
1) have the action on a common bulk actions folder, but have the feature split in two folders
2) have the feature in the same folder and have the bulk action in a folder that is not the common bulk actions folder

Please note that 2) would be consistent with what we did for `nuxeo-publish-button`, which is a complex bulk actions that depends on other elements, and resides in its own folder. I personally would be ok with option 2), but I would like others to chip in. @semisse @mnixo

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
Gabez0r: Great explanation. I would even go as far as to mention the example of field consisting of a div with a separate label and widget element inside.

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
Gabez0r: Ah, you are right! Sorry for the noise. 🔍

test/nuxeo-edit-documents-button.test.js
```diff
@@ -0,0 +1,434 @@
...
+      expect(isElementVisible(dialog)).to.be.true;
+      expect(isElementVisible(cancelButton)).to.be.true;
+    });
+
+    test('Should disable the save button if there are no properties to update', async () => {
+      button = await buildButton();
+      const { save } = button.$;
+      const expiredWidget = getWidget('dc:expired');
+      // open the bulk edit dialog
+      button._execute();
```
Gabez0r: We should avoid relying on private methods as much as possible, given that they expose internal implementation details. I would expect us to be able to achieve the same thing with:
```suggestion
      button.click();
```

Gabez0r (commented):
Great job on the tests!

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
mnixo: Ok, so in this case we'd be looking at a directory inside `elements/` that would include:
- `nuxeo-edit-documents-button.js` (the bulk action)
- `nuxeo-bulk-widget.js` (the wrapper)
- `nuxeo-bulk-default-layout.html` (the default bulk layout)
- Future custom bulk layouts.

How does `elements/nuxeo-bulk-edit/` sound?

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
mnixo: Will do.

test/nuxeo-edit-documents-button.test.js
mnixo: Unfortunately `button.click()` doesn't map to `button._execute()`, but `button.$$('.action').click()` does.
I'm going to use that instead!

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
Gabez0r: I could go with that or just plain "bulk", like we have now. The "search" folder, for example, also has a mix of layouts and elements.

test/nuxeo-edit-documents-button.test.js
```diff
@@ -0,0 +1,434 @@
...
+    test('Should open, close and reopen the bulk edit action dialog', async () => {
+      button = await buildButton();
+      const { dialog } = button.$;
+      const cancelButton = button.$$('paper-button.secondary');
+      // check that dialog is closed
+      expect(dialog.opened).to.be.false;
+      expect(isElementVisible(dialog)).to.be.false;
+      expect(isElementVisible(cancelButton)).to.be.false;
+      // open the bulk edit dialog
+      button.$$('.action').click();
```
richardsd, line 114: Shouldn't we click the `button` directly? (or is it executing any logic we don't want here?)

test/nuxeo-edit-documents-button.test.js
```diff
@@ -0,0 +1,434 @@
...
+      // check that save button is disabled
+      expect(save.disabled).to.be.true;
+      // set update mode to remove
+      expiredBulkWidget.updateMode = 'remove';
+      // check that save button is enabled
+      expect(save.disabled).to.be.false;
+      // set update mode to keep
+      expiredBulkWidget.updateMode = 'keep';
+      // check that save button is disabled
+      expect(save.disabled).to.be.true;
```
richardsd, line 236: I'm a bit confused: the test says ' Should enable the save button', but here we are doing an assert that the `save` button is disabled 🤔

richardsd (commented):
Nice work on the unit tests 💪

test/nuxeo-edit-documents-button.test.js
mnixo: We should, but it seems that actions extending `<nuxeo-operation-button>` don't really map the `.click()` method to anything. Tried it with a few of the other bulk ones and nothing happens. The way the `_execute()` is called is through clicking the `<div class="action">` inside of them.

richardsd: Well, it is weird in fact, but it is fine by me 👍  thanks

test/nuxeo-edit-documents-button.test.js
mnixo: It starts by checking that it is disabled, then checks that it gets enabled, and then checks that it goes back to disabled.

richardsd: 😄  ah, I see, it makes sense, sorry for the noise.

elements/nuxeo-document-bulk-actions/nuxeo-edit-documents-button.js
mnixo: Done!

Gabez0r (dismissed):

richardsd (dismissed):

Gabez0r (approved):

richardsd (approved):

semisse (approved):

PR #1319, opened 2021-10-27
WEBUI-552: better total count display

guirenard:


i18n/messages.json
```diff
@@ -1215,6 +1215,7 @@
   "repositoryAnalytics.topDownloads.heading": "Top Downloads",
   "repositoryAnalytics.topNCreators.heading": "Top {0} Creators",
   "results.heading.count": "{0} result(s)",
+  "results.heading.count.unknonw": "Many results",
```
Gabez0r, line 1218: This label needs to be validated with @bchauvin and @carolinagouveia.

Also, we need to replicate this label on all other messages files, even if keeping the english label, otherwise the label will be missing.

elements/nuxeo-results/nuxeo-results.html
```diff
@@ -592,7 +596,15 @@
...
-        return this.nxProvider && this.resultsCount;
+        if (this.nxProvider && this.resultsCount) {
+          if (this.resultsCount < 0) {
+            this._resultsCountLabel = this.i18n('results.heading.count.unknonw');
+          } else {
+            this._resultsCountLabel = this.i18n('results.heading.count', this.resultsCount);
+          }
+          return true;
+        }
+        return false;
```
Gabez0r, line 607: I think we are mixing two things here: we are using the same function to control visibility and also to set the label, wheres both can be independent. The logic for computing the label, for example, only depends on `this.resultsCount`. `_resultsCountLabel` could be a computed property.

Gabez0r (commented):
Thank you so much for the fix, @guirenard! Could you please upload a screenshot of how it looks so we can share with @carolinagouveia and @bchauvin ? Thank you.

i18n/messages.json
Gabez0r: As a matter of fact, whenever we introduce new labels, we can no longer push code without translations for other languages as well. We usually merge first on maintenance-3.0.x, wait for translations, backport for 10.10 and merge on 10.10.

guirenard:
> Thank you so much for the fix, @guirenard! Could you please upload a screenshot of how it looks so we can share with @carolinagouveia and @bchauvin ? Thank you.

I added before_NXP-30633.png in https://jira.nuxeo.com/browse/WEBUI-552

i18n/messages.json
guirenard: Ok I am gonna close this PR, let's focus on the `maintenance-3.0.x` PR then

PR #1322, opened 2021-10-29
ELEMENTS-1348: compute hrefBase on nuxeo-search-page and nuxeo-note-editor

Gabez0r:
This is another take on the problem described by ELEMENTS-1348 (see https://github.com/nuxeo/nuxeo-elements/pull/427). The goal is to redefined the hrefBase property of the nuxeo-results-view (even if nested inside the html editor) on Web UI, as this is the only place where we know where the default layouts live (the search/ folder).

Notes: I don't particularly like that we need to expose the hrefBase on the `nuxeo-html-editor` element. The alternative would be to pierce the shadow dom on the note editor element or use a different approach. Suggestions are more than welcome.

Note that I have used an ELEMENTS ticket but I will create a WEBUI one if we agree that doing a fix on both ends makes sense.

### DIFF
--- a/elements/nuxeo-note-editor/nuxeo-note-editor.js
+++ b/elements/nuxeo-note-editor/nuxeo-note-editor.js
@@ -100,7 +100,11 @@ Polymer({
             >[[_computeHtmlEditLabel(_viewMode, i18n)]]</paper-tooltip
           >
           <template is="dom-if" if="[[_viewMode]]">
-            <nuxeo-html-editor value="{{_value}}" read-only="[[!_canEdit(document)]]"></nuxeo-html-editor>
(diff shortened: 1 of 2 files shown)


richardsd (commented):
Both approaches aren't perfect, but passing the href around looks cleaner  than piercing through the shadow dom.

Gabez0r:
Another alternative to this would be to provide global configuration properties for either hrefBase or hrefTemplate of the `nuxeo-search-form-layout` and `nuxeo-search-results-layout` elements.

Gabez0r:
Closed in favor of https://github.com/nuxeo/nuxeo-web-ui/pull/1323.

PR #1331, opened 2021-11-05
WEBUI-188: fix upload of padded mp version to connect preprod

richardsd:


semisse (dismissed):

.github/workflows/main.yaml
```diff
@@ -123,10 +123,20 @@ jobs:
...
-        curl -i -u "${{ secrets.CONNECT_PREPROD_AUTH }}" -F package=@$PACKAGE "$CONNECT_PREPROD_URL/site/marketplace/upload?batch=true"
+        PACKAGE_NAME="nuxeo-web-ui-marketplace-$VERSION"
+        PACKAGE="plugin/web-ui/marketplace/target/$PACKAGE_NAME"
+
+        unzip -d $PACKAGE $PACKAGE.zip && rm $PACKAGE.zip
+        pushd $PACKAGE
+        sed -i 's/'"$VERSION"'/'"$PADDED_VERSION"'/g' package.xml
+        for f in install/bundles/*.jar; do mv $f $(echo $f | sed 's/'"$VERSION"'/'"$PADDED_VERSION"'/g'); done
+        zip -r ../$PACKAGE_NAME.zip .
+        popd
```
Gabez0r: Can't we just update the pom version with the padded version so that we don't have to repack the zip? We could do it in the "Update versions" stage, like we did on the old [Jenkinsfile](https://github.com/nuxeo/nuxeo-web-ui/blob/master/Jenkinsfile#L123).

richardsd: You're right, just pushed another version of the solution. It seems simple and cleaner. Thanks 👍

Gabez0r (approved):

semisse (approved):

mnixo (approved):

PR #1337, opened 2021-11-12
Web UI 573 add functional tests to bulkedit

semisse:
This draft pull request is intended to show the progress of functional tests in the bulk edit feature. Please keep in mind there is still some work to be done, namely, to test the options `Don't change value(s)`, `Replace with`, `Remove value(s)`.

Either way, we can start looking into file structure, nomenclature, and overall step definitions.

Thank you


ftest/features/bulk_edit.feature
```diff
@@ -0,0 +1,37 @@
...
+    And I can see the selection toolbar
+    When I click the edit button
+    And I see a dialog with editing options
+    Then I can edit multiple properties:
+      | name         | value                |
+      | nature       | Internship report    |
+      | subjects     | Medicine             |
+      | coverage     | Canada               |
+    And I click the save button
+    # Then I see a toast notification with the following message "Bulk Edit documents completed successfully on 3 document(s)"
```
richardsd, line 30: Just wondering if we really should check for the notification 🤔  Isn't this notification sticky? I believe you said that you had troubles asserting the context of the snackbar

ftest/features/bulk_edit.feature
```diff
@@ -0,0 +1,37 @@
...
+    And I click the dismiss button
```
richardsd, line 31: If you don't assert the notification, do you need to dismiss it?

packages/nuxeo-web-ui-ftest/pages/ui/bulk_edit.js
```diff
@@ -0,0 +1,64 @@
...
+    table.rows().forEach((row) => {
+      const fieldName = row[0];
+      const fieldEl = this.getField(fieldName);
+      fieldEl.waitForVisible();
+      fieldEl.scrollIntoView();
+      return fixtures.layouts.setValue(fieldEl, row[1]);
+    });
+  }
+
+  clickSaveButton() {
```
richardsd, line 59: It feels a bit redundant to have the save button getter and this method here. Looking into other parts of the framework, we do it like this https://github.com/nuxeo/nuxeo-web-ui/blob/maintenance-3.0.x/packages/nuxeo-web-ui-ftest/features/step_definitions/create_dialog.js#L61 (not sure what the best solution should be though)

packages/nuxeo-web-ui-ftest/features/step_definitions/bulk_edit.js
```diff
@@ -0,0 +1,29 @@
...
+  this.ui.bulkEdit.clickEditDocumentsButton();
+});
+
+Then('I see a dialog with editing options', function() {
+  this.ui.bulkEdit.dialog.isVisible().should.be.true;
+  this.ui.bulkEdit.bulkDefaultLayout.isVisible().should.be.true;
+});
+
+Then('I can edit multiple properties:', function(table) {
+  this.ui.bulkEdit.editMultipleOptions(table);
```
richardsd, line 13: I wonder if we could fill in the bulk edit layout just like we do for other layouts (see [here](https://github.com/nuxeo/nuxeo-web-ui/blob/maintenance-3.0.x/packages/nuxeo-web-ui-ftest/features/step_definitions/create_dialog.js#L57)). This way we wouldn't need the `getField` and `getFieldValue`, etc. in the `BulkEdit` page.

packages/nuxeo-web-ui-ftest/pages/ui/bulk_edit.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  setFieldValue(field, value) {
+    const fieldEl = this.getField(field);
+    fieldEl.waitForVisible();
+    return fixtures.layouts.setValue(fieldEl, value);
+  }
+
+  clickEditDocumentsButton() {
+    const el = this.editDocumentsButton;
```
richardsd, line 44: Same was bellow (for save button).

richardsd (commented):
@semisse left some comments for things we could discuss and improve (or not, consider them as food for though). Btw, it is very cool to have two people working on this subject 👍

ftest/features/bulk_edit.feature
semisse: Yes, the notifications are still a work in progress. I will push some code that will fix this.

ftest/features/bulk_edit.feature
semisse: Yes, the notifications are still a work in progress. I will push some code that will fix this.

packages/nuxeo-web-ui-ftest/pages/ui/bulk_edit.js
semisse: Makes sense, and it is coherent. I will change it. Thanks.

PR #1341, opened 2021-11-16
WEBUI-265: automate a11y checks with axe-core

Gabez0r:
This is a rebase of https://github.com/nuxeo/nuxeo-web-ui/pull/1169 on maintenance-3.0.x, to take into account the upgrade for wdio v7 and the move to GH actions.


plugin/a11y/getDriverVersion.js
```diff
@@ -0,0 +1,35 @@
...
+          .catch((e) => {
+            console.error('unable to parse ChromeDriver version: ', e);
+          });
+      }
+      console.error('unable to fetch ChromeDriver version: ', response);
+    });
+  } catch (e) {
+    console.error('unable to fetch ChromeDriver version: ', e);
+  }
+}
```
Gabez0r, line 35: This code is duplicated from nuxeo-web-ui-ftest. However, I didn't find a good way to expose this publicly, in a way that it could be shared and still make sense. Feedback is welcome, but we can also do this at a later stage.

plugin/a11y/README.md
```diff
@@ -0,0 +1,39 @@
...
+## Run tests
+
+To run the tests agains a running Nuxeo server:
+
+```
+npm run test
+```
+
+The server URL can be paremeterized using the `NUXEO_URL` environment variable. In case the instance of Web UI runs on
+a seprate server, it can be configured using the `NUXEO_WEB_UI_URL` environment variable. E.g.:
```
richardsd: Typo: `seprate` should be `separate server`

plugin/a11y/getDriverVersion.js
richardsd: I think we can do it in a later stage 👍

richardsd: Either way, could we put it in the testing-helpers? Or it doesn't make sense?

Gabez0r:
@richardsd 
> Either way, could we put it in the testing-helpers? Or it doesn't make sense?

That means exposing public API, which is what I want to avoid if we are not sure this is the best way to do it.

plugin/a11y/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+```
+NUXEO_WEB_UI_URL=http://localhost:5000/ NUXEO_URL=http://localhost:8080/nuxeo/ npm run test
+```
+
+The tests can be run in development mode, which assume a nuxeo server running locally (http://localhost:8080/nuxeo/) and
+Web UI being server in development mode from port 5000 (http://localhost:5000/):
```
mnixo:
```suggestion
Web UI being served in development mode from port 5000 (http://localhost:5000/):
```
?

plugin/a11y/pom.xml
```diff
@@ -0,0 +1,101 @@
...
+                </resource>
+              </resources>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+
+</project>
```
mnixo:
```suggestion
</project>

```

plugin/a11y/getDriverVersion.js
Gabez0r: https://github.com/nuxeo/nuxeo-web-ui/pull/1341#issuecomment-973119786

richardsd (approved):

mnixo (approved):

PR #1344, opened 2021-11-22
WEBUI-573: add functional tests to bulk edit

semisse:


ftest/features/bulk_edit.feature
```diff
@@ -0,0 +1,50 @@
...
+    Then I see a dialog with editing options
+    And I can edit multiple properties:
+      | name         | value                |
+      | nature       | Internship report    |
+      | subjects     | Medicine             |
+      | coverage     | Canada               |
+    Then I click the save button
+    # Then I see a toast notification with the following message "Bulk Edit documents is scheduled. You will be notified when it starts."
+    And I see a toast notification with the following message "Bulk Edit documents completed successfully on 3 document(s)"
+    When I click the dismiss button
```
richardsd: Is it really relevant to dismiss the snackbar? (just wondering)

packages/nuxeo-web-ui-ftest/pages/ui.js
```diff
@@ -170,4 +171,22 @@ export default class UI extends BasePage {
...
     driver.waitUntil(() => driver.elements('mwc-snackbar').every((toast) => !toast.getAttribute('open')));
   }
+
+  getToastDismissButton() {
+    return this.el.element('#snackbarPanel mwc-snackbar[open] #dismiss');
+  }
+
+  getToastMessage(message) {
+    let snackBar;
+    driver.waitUntil(() => {
```
richardsd, line 181: Ok, so you have to wait until the message is updated 👍

richardsd (approved):

ftest/features/bulk_edit.feature
```diff
@@ -0,0 +1,50 @@
...
+      | Workspace  | Test_Workspace   | booklet | sciences/astronomy      | europe/Portugal      | JSmith  | /default-domain                             |                  |        |            |
+      | Folder     | Test_Folder      | booklet | sciences/astronomy      | europe/Portugal      | JSmith  | /default-domain/Test_Workspace/             |                  |        |            |
+      | File       | First_File       | booklet | art/culture             | europe/France        | JSmith  | /default-domain/Test_Workspace/Test_Folder/ |                  |        | sample.png |
+      | File       | Second_File      | booklet | art/culture             | europe/France        | JSmith  | /default-domain/Test_Workspace/Test_Folder/ |                  |        | sample.png |
+      | File       | Third_File       | booklet | art/culture             | europe/France        | JSmith  | /default-domain/Test_Workspace/Test_Folder/ |                  |        | sample.png |
+    And I login as "Administrator"
+
+  @config('selection.selectAllEnabled','true')
+  Scenario: Select all documents and replace them
```
mnixo: `replace them`? Or `edit them`?
EDIT: Looking below I'm noticing that `them` is in reference to the properties and not the documents, although it seems that you're saying you're going to replace documents at first glance.

mnixo (approved):
Thank you!

ftest/features/bulk_edit.feature
```diff
@@ -0,0 +1,50 @@
+@watch
```
Gabez0r: Please don't forget to remove this.

ftest/features/bulk_edit.feature
Gabez0r: I agree that "and replace them" is a little ambiguous. Perhaps something like: "Select all documents and replace properties" or "Bulk replace properties"?

ftest/features/bulk_edit.feature
semisse: I don't think it hurts 🤷‍♂️

ftest/features/bulk_edit.feature
semisse: I agree, let's make it more clear.

ftest/features/bulk_edit.feature
```diff
@@ -0,0 +1,50 @@
...
+    And I login as "Administrator"
+
+  @config('selection.selectAllEnabled','true')
+  Scenario: Select all documents and replace them
+    When I browse to the document with path "/default-domain/Test_Workspace/Test_Folder"
+    And I can see the "Test_Folder" document
+    Then I select all the documents
+    And I can see the selection toolbar
+    When I click the edit button
+    Then I see a dialog with editing options
```
Gabez0r: I would expect these to be a single step.

ftest/features/bulk_edit.feature
```diff
@@ -0,0 +1,50 @@
...
+    And I can edit multiple properties:
+      | name         | value                |
+      | nature       | Internship report    |
+      | subjects     | Medicine             |
+      | coverage     | Canada               |
+    Then I click the save button
```
Gabez0r: I would expect these to be a single step too. Being able to edit requires me to click on the save the button. This would also be more consistent with other edition steps, such as `I can edit the following properties in the (.+) metadata`.

packages/nuxeo-web-ui-ftest/features/step_definitions/bulk_edit.js
```diff
@@ -0,0 +1,44 @@
+import { Then } from '@cucumber/cucumber';
+
+Then('I click the edit button', function() {
```
Gabez0r: I think we should make this description a bit more specific. It not just any edit button, but the bulk edit button, and we have other edit buttons.

packages/nuxeo-web-ui-ftest/features/step_definitions/bulk_edit.js
```diff
@@ -0,0 +1,44 @@
...
+  this.ui.bulkEdit.editDocumentsButton.click();
+});
+
+Then('I see a dialog with editing options', function() {
+  this.ui.bulkEdit.dialog.isVisible().should.be.true;
+  this.ui.bulkEdit.bulkDefaultLayout.isVisible().should.be.true;
```
Gabez0r: Would recommend for stability:
```suggestion
  this.ui.bulkEdit.dialog.waitForDisplayed();
  this.ui.bulkEdit.bulkDefaultLayout.waitForDisplayed();
```

packages/nuxeo-web-ui-ftest/features/step_definitions/bulk_edit.js
```diff
@@ -0,0 +1,44 @@
...
+  this.ui.bulkEdit.saveButton.click();
+});
+
+Then('I see a toast notification with the following message {string}', function(message) {
+  const notificationMessage = this.ui.getToastMessage(message);
+  const trimmedMessage = message.trim().replace(/"/g, '');
+  notificationMessage.should.be.equals(trimmedMessage);
+});
+
+Then('I click the dismiss button', function() {
```
Gabez0r: Perhaps we should make it more specific, so that it pertains to the toast?

packages/nuxeo-web-ui-ftest/features/step_definitions/bulk_edit.js
```diff
@@ -0,0 +1,44 @@
...
+  this.ui.getToastDismissButton().click();
+  this.ui.waitForToastNotVisible();
+});
+
+Then('I remove {string} values from its field', function(field) {
```
Gabez0r: Could we make it more specific to bulk? Also, we could probably have a generic step to pick an option from the dropdown instead?

ftest/features/bulk_edit.feature
```diff
@@ -0,0 +1,50 @@
...
+  Scenario: Select all documents and remove values
+    When I browse to the document with path "/default-domain/Test_Workspace/Test_Folder"
+    Then I can see the "Test_Folder" document
+    When I select all the documents
+    And I click the edit button
+    Then I remove "Nature" values from its field
+    And I click the save button
+    Then I see a toast notification with the following message "Bulk Edit documents completed successfully on 3 document(s)"
+    When I browse to the document with path "/default-domain/Test_Workspace/Test_Folder/First_File"
+    Then I see "Nature" value was removed from its field
```
Gabez0r: Can't we reuse the `I can see File metadata with the following properties` step here?

semisse: That was my first option but I don't think the method is prepared to get empty/hidden fields. Should we improve it to accept those?

Gabez0r: I would expect the step definition should work with empty fields (see [metadata.feature#L65-L73](https://github.com/nuxeo/nuxeo-web-ui/blob/maintenance-3.0.x/ftest/features/metadata.feature#L65-L73)). But we might have to do some tweaks.

mnixo (approved):

packages/nuxeo-web-ui-ftest/features/step_definitions/bulk_edit.js
```diff
@@ -0,0 +1,42 @@
+import { Then } from '@cucumber/cucumber';
+
+Then('I click the bulk edit button I see a dialog with editing options', function() {
+  this.ui.bulkEdit.editDocumentsButton.click();
+  this.ui.bulkEdit.dialog.waitForDisplayed();
+  this.ui.bulkEdit.bulkDefaultLayout.waitForDisplayed();
+  this.ui.bulkEdit.dialog.isVisible().should.be.true;
+  this.ui.bulkEdit.bulkDefaultLayout.isVisible().should.be.true;
```
Gabez0r: We don't need these likes because we are already waiting for they to be displayed.

packages/nuxeo-web-ui-ftest/features/step_definitions/bulk_edit.js
```diff
@@ -0,0 +1,42 @@
...
+});
+
+Then('I can edit multiple properties:', function(table) {
```
Gabez0r: I think that this step (and any other step using the dialog and the bulk edit form, such as the `I can bulk edit all documents and {string} from {string} field` step), should be the ones waiting for the dialog to be visible, instead of assuming it will. This would make them a bit more reusable. This would also turn the `I click the bulk edit button I see a dialog with editing options` step into something simpler, like `I click the bulk edit button`.

Gabez0r: "I can edit multiple properties" is still a bit too generic. What about:
```suggestion
Then('I can bulk edit multiple properties:', function(table) {
```

ftest/features/bulk_edit.feature
```diff
@@ -0,0 +1,49 @@
...
+      | subjects     | Medicine             |
+      | coverage     | North-america/Canada |
+
+  @config('selection.selectAllEnabled','true')
+  Scenario: Select all documents and remove values
+    When I browse to the document with path "/default-domain/Test_Workspace/Test_Folder"
+    Then I can see the "Test_Folder" document
+    When I select all the documents
+    When I click the bulk edit button I see a dialog with editing options
+    Then I can bulk edit all documents and "Remove value(s)" from "Nature" field
```
Gabez0r: I know that it was me who suggested turning the action into a parameter, but now I'm wondering how much reusable this actually is. 🤔 

I am also wondering if we need to have a separate edition steps for removing and setting values. What if we could have something like:
```
    Then I can bulk edit multiple properties:
      | name         | value                | action    |
      | nature       | Internship report    | append    |
      | subjects     | Medicine             |           |
      | coverage     |                      | remove    |
```

richardsd (commented):
Gabriel suggestions look good and worth checking, other than that it looks good to me 👍

ftest/features/bulk_edit.feature
```diff
@@ -0,0 +1,34 @@
...
+      | subjects     | Medicine             | replace   |
+      | coverage     | Canada               | replace   |
+    And I see a toast notification with the following message "Bulk Edit documents completed successfully on 3 document(s)"
+    When I click the toast notification dismiss button
+    And I browse to the document with path "/default-domain/Test_Workspace/Test_Folder/First_File"
+    Then I can see File metadata with the following properties:
+      | name         | value                |
+      | nature       |                      |
+      | subjects     | Medicine             |
+      | coverage     | North-america/Canada |
```
mnixo: The tiniest nitpick:
```suggestion
      | coverage     | North-america/Canada |

```

mnixo (approved):

ftest/features/bulk_edit.feature
```diff
@@ -0,0 +1,34 @@
...
+  @config('selection.selectAllEnabled','true')
+  Scenario: Select all documents and replace properties
+    When I browse to the document with path "/default-domain/Test_Workspace/Test_Folder"
+    And I can see the "Test_Folder" document
+    Then I select all the documents
+    And I can see the selection toolbar
+    When I click the bulk edit button
+    Then I can bulk edit multiple properties:
+      | name         | value                | action    |
+      | nature       | booklet              | remove    |
```
Gabez0r: I guess we don't need to pass a value, right?
```suggestion
      | nature       |                      | remove    |
```

packages/nuxeo-web-ui-ftest/pages/ui/bulk_edit.js
```diff
@@ -0,0 +1,69 @@
...
+    const fieldEl = this.getField(field);
+    fieldEl.waitForVisible();
+    return fixtures.layouts.setValue(fieldEl, value);
+  }
+
+  editMultipleOptions(table) {
+    table.rows().forEach((row) => {
+      const fieldName = row[0];
+      const fieldValue = row[1];
+      const action = row[2];
```
Gabez0r: nit: can be conciser with object destructuring:
```suggestion
      const [ fieldName, fieldValue, action ] = row;
```

packages/nuxeo-web-ui-ftest/pages/ui/bulk_edit.js
```diff
@@ -0,0 +1,69 @@
...
+      const fieldEl = this.getField(fieldName);
+      fieldEl.waitForVisible();
+      if (action === 'remove') {
+        this.getBulkEditOptions(fieldName).click();
+        this.bulkEditOptionsList('Remove value(s)').click();
+        return;
+      }
+      if (action === 'replace') {
```
Gabez0r: What happens if action is empty? Should we assume "replace" is the default or should we always want it to be made explicit?

semisse: Also thought about it, in my opinion, the default should be `Don't change values`, which, in practice, is what we already do in the interface. Making it explicit also makes it more understandable when reading the test.
Of course, I'm open to listening to more opinions about this.

Gabez0r: Sounds good to me.

ftest/features/bulk_edit.feature
```diff
@@ -0,0 +1,35 @@
+@watch
```
Gabez0r: Please don't forget to remove this.

ftest/features/bulk_edit.feature
```diff
@@ -0,0 +1,35 @@
...
+      | Workspace  | Test_Workspace   | booklet | art/culture             | europe/Portugal      | JSmith  | /default-domain                             |                  |        |            |
+      | Folder     | Test_Folder      | booklet | art/culture             | europe/Portugal      | JSmith  | /default-domain/Test_Workspace/             |                  |        |            |
+      | File       | First_File       | booklet | art/culture             | europe/Portugal      | JSmith  | /default-domain/Test_Workspace/Test_Folder/ |                  |        | sample.png |
+      | File       | Second_File      | booklet | art/culture             | europe/Portugal      | JSmith  | /default-domain/Test_Workspace/Test_Folder/ |                  |        | sample.png |
+      | File       | Third_File       | booklet | art/culture             | europe/Portugal      | JSmith  | /default-domain/Test_Workspace/Test_Folder/ |                  |        | sample.png |
+    And I login as "Administrator"
```
Gabez0r, line 13: I am assuming you are going to use it next, because having a background only makes sense if you have more than on scenario.

packages/nuxeo-web-ui-ftest/pages/ui/bulk_edit.js
richardsd: It sounds good to me as well.

richardsd: Just out of curiosity: why not an `else if` without the return? It's by design?

richardsd (approved):

richardsd (approved):

Gabez0r (approved):

PR #1348, opened 2021-12-03
Fix web UI 470 maca issues

Gabez0r:
Tried to fix WEBUI-478 and WEBUI-434 without breaking a11y, replacing anchors with buttons. I did not backport the a11y fixing for 10.10 (https://github.com/nuxeo/nuxeo-web-ui/pull/1349). Reviews are welcome.

### DIFF
--- a/elements/document/nuxeo-document-import.js
+++ b/elements/document/nuxeo-document-import.js
@@ -351,6 +351,20 @@ Polymer({
       paper-progress {
         margin-top: 8px;
       }
+
+      button.link {
+        color: var(--nuxeo-link-color, #3a3a54);
+        padding: 0;
+        background: none;
+        border: none;
+        cursor: pointer;
+        font: inherit;
+      }
+
+      button.link:hover {
+        color: var(--nuxeo-link-hover-color, #0066ff);
+        font: inherit;
+      }
     </style>
 
     <nuxeo-connection id="nx"></nuxeo-connection>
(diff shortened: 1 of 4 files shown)


richardsd (commented):
So, in short, we replace the anchors with buttons and add the `modulePathPrefix` to the service worker, right? Btw, the a11y tests are not passing, but it doesn't seem related with your changes 🤔

Gabez0r:
@richardsd:
> So, in short, we replace the anchors with buttons and add the modulePathPrefix to the service worker, right?

Exactly! The buttons part is only for 3.0.x, so that we don't fix and issue and introduce another (a11y) one.

> Btw, the a11y tests are not passing, but it doesn't seem related with your changes 🤔

Thanks. Going to re-run it to see if it goes through.

elements/document/nuxeo-document-import.js
```diff
@@ -351,6 +351,21 @@ Polymer({
       paper-progress {
         margin-top: 8px;
       }
+
+      button.link {
+        color: var(--nuxeo-secondary-color, #0066ff);
+        text-decoration: underline;
```
semisse: We should use the `nuxeo-link-color` and remove the underline to match the style we have currently.
```suggestion
        color: var(--nuxeo-link-color, #3a3a54);
```
Original:
Screenshot 2021-12-07 at 10 22 04

Change:
Screenshot 2021-12-07 at 10 21 56

Gabez0r: Note that this means changing the way it looks today, and I would consider this outside the scope of this ticket, which is about fixing CSP issues. I'd rather have a separate ticket open to address this.

elements/nuxeo-app/nuxeo-expired-session.js
```diff
@@ -54,20 +54,24 @@ Polymer({
         -webkit-transform: translateY(0);
       }
 
-      a,
-      a:active,
-      a:visited,
-      a:focus {
+      button.link {
+        text-decoration: underline;
```
semisse: Same here, we should remove the underline style to match what we have currently.

```suggestion
   color: var(--nuxeo-link-color, #3a3a54)
```

elements/document/nuxeo-document-import.js
Gabez0r: Sorry, you are right. I misunderstood the original vs change screenshots. Will do the change, thanks.

semisse (approved):

richardsd (approved):

PR #1351, opened 2021-12-09
WEBUI-615: pass resultsCount to nuxeo-selection-toolbar for select all is active

richardsd:


mnixo (approved):

semisse (approved):

elements/nuxeo-selection/nuxeo-selection-toolbar.js
```diff
@@ -116,7 +116,7 @@ Polymer({
 
           <!-- else if all the elements are selected -->
           <template is="dom-if" if="[[selectAllActive]]">
-            <span class="count">[[i18n('selectionToolbar.selected.all', selectedItems.length)]]</span>
+            <span class="count">[[i18n('selectionToolbar.selected.all', _resultsCount)]]</span>
```
Gabez0r, line 119: Won't this break the current behavior if not doing select all? We would be displaying the whole result count even if we are only selecting a couple of items, right?

richardsd: It shouldn't because it would be hidden as this is only displayed if `selectAllActive` is true (see [here](https://github.com/nuxeo/nuxeo-web-ui/pull/1351/files/ff69ad2e0590c5d35099863f64d9fac50bff9e23#diff-114c49ef84a663fa043e5f5c9707f7e381d92130200720a4b0fbfbb2b97a71d8L118)).

richardsd: If `selectAllActive` is false, then we use the old behavior (see [here](https://github.com/nuxeo/nuxeo-web-ui/blob/ff69ad2e0590c5d35099863f64d9fac50bff9e23/elements/nuxeo-selection/nuxeo-selection-toolbar.js#L110)).

Gabez0r: Ah, you are right! Overlooked the if on top. Looks good to me then!

Gabez0r (approved):

elements/nuxeo-selection/nuxeo-selection-toolbar.js
richardsd: 👍  no problem, I'm glad you asked. At least we double checked it. Thanks 😉

PR #1356, opened 2021-12-23
Web UI 599 10.10 ondemand pipeline

richardsd:
Initial work on the 10.10 ondemand pipeline (single job alternative to https://github.com/nuxeo/nuxeo-web-ui/pull/1359).

Feedback is very welcome 🙂

### DIFF
--- a/.github/actions/get-branch/action.yaml
+++ b/.github/actions/get-branch/action.yaml
@@ -12,10 +12,10 @@ inputs:
     required: true
 outputs:
   branch:
-    description: "The name of the branch to checkout"
+    description: 'The name of the branch to checkout'
     value: ${{ steps.pick_branch_to_checkout.outputs.branch }}
 runs:
-  using: "composite"
+  using: 'composite'
   steps:
     - id: pick_branch_to_checkout
       shell: bash
--- a/.github/actions/get-branch/action.yaml
+++ b/.github/actions/get-branch/action.yaml
@@ -1,0 +1,27 @@
+name: 'Get branch'
+description: 'Get branch to checkout in repository or the default one'
+inputs:
+  repository:
+    description: 'github repository name'
+    required: true
+  branch:
+    description: 'branch name'
+    required: true
+  default-branch:
+    description: 'default branch name'
+    required: true
+outputs:
+  branch:
+    description: "The name of the branch to checkout"
+    value: ${{ steps.pick_branch_to_checkout.outputs.branch }}
+runs:
+  using: "composite"
+  steps:
+    - id: pick_branch_to_checkout
+      shell: bash
+      run: |
+        if git ls-remote --exit-code --heads https://github.com/${{ inputs.repository }} ${{ inputs.branch }}; then
+          echo ::set-output name=branch::${{ inputs.branch }}
+        else
+          echo ::set-output name=branch::${{ inputs.default-branch }}
+        fi
--- a/.github/workflows/ondemand-main.yaml
+++ b/.github/workflows/ondemand-main.yaml
@@ -127,8 +127,11 @@ jobs:
         pushd nuxeo-elements
         npm install && bower install
         popd
+
         pushd nuxeo-ui-elements
         npm install && bower install
+        rm -rf bower_components/nuxeo-elements
+        rsync -av --exclude='node_modules' --exclude='.git' --exclude='bower_components' ../nuxeo-elements bower_components
         popd
 
     - name: Lint Elements
--- a/.github/workflows/ondemand-main.yaml
+++ b/.github/workflows/ondemand-main.yaml
@@ -150,26 +150,25 @@ jobs:
         npm run test
         popd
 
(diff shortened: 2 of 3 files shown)


.github/actions/get-branch/action.yaml
```diff
@@ -0,0 +1,27 @@
...
+    required: true
+  branch:
+    description: 'branch name'
+    required: true
+  default-branch:
+    description: 'default branch name'
+    required: true
+outputs:
+  branch:
+    description: "The name of the branch to checkout"
```
mnixo: In these `.yaml` files, when are we supposed to use `""` or `''`?

mnixo (dismissed):

.github/workflows/ondemand-main.yaml
```diff
@@ -0,0 +1,307 @@
...
+
+    - name: Determine nuxeo-elements branch to use
+      id: pick_nuxeo_elements_branch
+      uses: ./.github/actions/get-branch
+      with:
+        repository: nuxeo/nuxeo-elements
+        branch: ${{ github.event.inputs.branch_name }}
+        default-branch: ${{ env.ELEMENTS_REFERENCE_BRANCH }}
+
+    - name: Determine nuxeo-ui-elements branch to use
```
Gabez0r, line 86: We are missing [nuxeo/nuxeo-dataviz-elements](https://github.com/nuxeo/nuxeo-dataviz-elements/tree/maintenance-2.4.x). I see that this is also the case for the current legacy pipeline, so I'd be ok with having a dedicated ticket for this if it causes too much burden.

.github/actions/get-branch/action.yaml
Gabez0r: Indeed, we see to be mixing both.

.github/workflows/ondemand-main.yaml
```diff
@@ -0,0 +1,307 @@
...
+      run: |
+        pushd nuxeo-web-ui
+        sed -i 's/10.10-HF[0-9]\{2\}-SNAPSHOT/10.10-HF${{ github.event.inputs.hf_version }}/g' pom.xml
+        popd
+
+    - name: Nuxeo package build
+      run: |
+        pushd nuxeo-web-ui
+        mvn install -DskipInstall
+        popd
```
Gabez0r: We are repeatedly doing `pushd nuxeo-web-ui` and `popd` on these steps. Do we really need it?

.github/workflows/ondemand-main.yaml
```diff
@@ -0,0 +1,307 @@
...
+      with:
+        repository: nuxeo/nuxeo-web-ui-itests
+        path: nuxeo-web-ui-itests
+        ref:  ${{ steps.pick_plugin_nuxeo_web_ui_itests_branch.outputs.branch }}
+
+    # HACK - this is a workaround that will be removed. Necessary because SNAPSHOT version might not be published yet
+    - name: Update nuxeo-web-ui-itests SNAPSHOT version with sed
+      run: |
+        pushd nuxeo-web-ui-itests
+        grep -rl '10.10-HF[0-9]\{2\}-SNAPSHOT' . --exclude-dir={.git,node_modules,bower_components} | xargs sed -i 's/10.10-HF[0-9]\{2\}-SNAPSHOT/10.10-HF${{ github.event.inputs.hf_version }}/g'
```
Gabez0r: We could also try to filter for the files we are expecting to affect using `--include`. E.g.: `--include=**/pom.xml`.

Gabez0r (commented):
Great job! 👏 💪  Did you try this on legacy QA with the seds done manually? Does it run?

.github/workflows/ondemand-main.yaml
richardsd: You are totally right, we are missing the dataviz-elements in both the new and old pipelines. I'll add it, don't think it is that hard to add as well. Thanks

.github/actions/get-branch/action.yaml
richardsd: 🤔  You are partially right, I need to review the use of both `''` and `""`, but as a general rule, we need to use `""` in places where we need to perform variable substitution (which is not the case in this particular line).

.github/workflows/ondemand-main.yaml
richardsd: We don't really need it. I'm doing this to keep each step as an atomic unit, but we can try to simplify the workflow.

.github/actions/get-branch/action.yaml
richardsd: I'm reviewing it 👍 Thanks

.github/workflows/ondemand-main.yaml
richardsd: Done. Let me know what you think about it.

mnixo (approved):

richardsd:
Closing this one in favor of https://github.com/nuxeo/nuxeo-web-ui/pull/1359.

PR #1367, opened 2022-01-13
Lit poc

Gabez0r:
Depends on https://github.com/nuxeo/nuxeo-elements/pull/505.

### DIFF
--- a/.github/workflows/ftest.yaml
+++ b/.github/workflows/ftest.yaml
@@ -35,10 +35,11 @@ jobs:
         with:
           ref: ${{ env.BRANCH_NAME }}
 
-      - uses: actions/setup-node@v1
+      - uses: actions/setup-node@v3
         with:
           registry-url: ${{ env.NPM_REPOSITORY }}
           scope: '@nuxeo'
+          node-version: 12
 
       - uses: actions/setup-java@v2
         with: 
--- a/.github/workflows/ftest.yaml
+++ b/.github/workflows/ftest.yaml
@@ -123,6 +123,13 @@ jobs:
           name: cucumber-reports
           path: ftest/target/cucumber-reports/
 
+      - name: Archive screenshots
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v2
+        with:
+          name: screenshots
+          path: ftest/target/screenshots/
+
       - name: Archive logs
         if: ${{ always() }}
         uses: actions/upload-artifact@v2
--- a/elements/bulk/nuxeo-edit-documents-button.js
+++ b/elements/bulk/nuxeo-edit-documents-button.js
@@ -202,7 +202,7 @@ class NuxeoEditDocumentsButton extends mixinBehaviors([I18nBehavior, FiltersBeha
   }
 
   static get observers() {
-    return ['_loadLayout(layout, hrefFunction, hrefBase)'];
+    return ['_loadLayout(layout, hrefFunction, hrefBase, documents)'];
   }
 
   constructor() {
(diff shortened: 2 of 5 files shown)


Gabez0r:
Build broke initially because of incompatible versions of `@lit/reactive-element` between `mwc-snackbar` (which used an old one) and elements. This [PR](https://github.com/material-components/material-web/pull/3057) landed and a new release of `mwc-snackbar` was done, which fixed the issue. This points again to the **first limitation** reported in https://github.com/nuxeo/nuxeo-elements/pull/505: combining different major versions of lit, even if these come via transitive dependencies, might prevent Web UI from being built successfully.

Gabez0r:
Please ignore 8b68ad0f09ef60c4842ade12363d73159ec7bcf4, a3beb615a2bff62b769686e8ad5588f3f4bccf83 and 0127b757e95cc41fb6c4f49333fc31bf7554bb3d.

Gabez0r:
Requirements:
- Update `vaadin-date-picker`, which is already planned (see [WEBUI-799](https://jira.nuxeo.com/browse/WEBUI-799)).
- Update nodejs version, which is already planned (see [WEBUI-792](https://jira.nuxeo.com/browse/WEBUI-792)).
- Update `mwc-snackbar` (4e4e1ce17af37bf92607ca7a3f3d49df4a5d5648). The API of the element seems to have changed, so we will have to do some changes to our integration of the element with Web UI (see 4e4e1ce17af37bf92607ca7a3f3d49df4a5d5648).

Gabez0r:
Beyond the requirements, no further effort is needed on the Web UI side to make sure an element that was converted on `nuxeo-elements` can be safely used.

elements/bulk/nuxeo-edit-documents-button.js
```diff
@@ -430,14 +430,10 @@ class NuxeoEditDocumentsButton extends mixinBehaviors([I18nBehavior, FiltersBeha
    * Loads the bulk edit layout.
    */
   _loadLayout(layout, hrefFunction, hrefBase) {
-    const { href } = this.$$('nuxeo-layout');
+    this._set_href(null);
```
richardsd, line 433: @Gabez0r Do we really need this change? Asking because this might cause regression of WEBUI-739, where we'll end up loading and stamping the layout every time the `documents` change.

.github/workflows/ftest.yaml
```diff
@@ -123,6 +124,13 @@ jobs:
           name: cucumber-reports
           path: ftest/target/cucumber-reports/
 
+      - name: Archive screenshots
```
richardsd, line 127: Created https://jira.nuxeo.com/browse/WEBUI-820 to add this to the pipeline.

elements/nuxeo-app.js
```diff
@@ -1395,6 +1395,9 @@ Polymer({
         }
         defaultAction();
       };
+      // XXX: needs refactoring, just copes with API changes
```
richardsd, line 1398: You mean a bug in the `mwc-snackbar` element?

richardsd (commented):
Besides some minor comments, it looks pretty straightforward

elements/bulk/nuxeo-edit-documents-button.js
Gabez0r: @richardsd The intention was to explicitly revert [WEBUI-739](https://jira.nuxeo.com/browse/WEBUI-739) so I could get the tests passing. This is to be removed as soon as the issue is fixed. It is not a requirement or related to the lit migration at all.

elements/nuxeo-app.js
Gabez0r: It is either a bug or they simply changed/broke the API.

elements/bulk/nuxeo-edit-documents-button.js
richardsd: Ah ok, I'm relieved then 😅

sgammon:
really excited for this :) if you need help or would like contributions from end-developers, let me know. i'd love to contribute.

PR #1372, opened 2022-01-31
WEBUI-245: browser title update

richardsd:


elements/nuxeo-app.js
```diff
@@ -632,7 +632,7 @@ Polymer({
     'nuxeo-diff-documents': '_diffDocuments',
   },
 
-  observers: ['_computeSharedActionContext(currentUser)', '_updateTitle(page, i18n)'],
```
Gabez0r, line 635: Why wasn't this working?

richardsd: The solution introduced, that removes the call to **_updateTitle()** when the **show()** method is triggered, and relies only on the observer **_updateTitle(page, i18n)** to update the title, falls short because the page doesn't change in subsequent actions (event though the currentDocument/search form, etc. has changed).

richardsd: When browsing, the page is `browse`, regardless of the document where you are. So:

```
this.page = page;
```
Doesn't do anything and the observer is not triggered.

Gabez0r: But this way we might be breaking the api, right? One can now switch the page property and now the title won't update, right? Wound we be able to fix this by adding the missing dependencies to this computed binding? Looking at the `_updateTitle`, we see to be missing a few: the currentDocument, searchForm, currentTask, selectedAdminTab. E.g. (didn't test):
```
observers: ['_computeSharedActionContext(currentUser)', '_updateTitle(page, i18n, currentDocument, searchForm, currentTask, selectedAdminTab)'],
```

richardsd: We broke the API [here](https://github.com/nuxeo/nuxeo-web-ui/commit/9477681ba75818833bb483a3b42ff63344753ae3), a long time ago, not by my code. In fact, what I'm doing is just adding the code we had in the past with minimal changes. So my main approach is to keep the amount of changes, as short as possible. Btw, this is exactly as 10.10 is working too.

ftest/features/browser.feature
```diff
@@ -38,3 +38,13 @@ Feature: Browser
...
+
+  Scenario: Document title update
+    When I click the "browser" button
+    And I can see the browser tree
+    And I can see the "Domain" browser tree node
+    And I click "Domain" in the browser tree
+    Then I can see the browser title as "Domain - Nuxeo Platform"
+    And I can see the "Workspaces" browser tree node
+    When I click "Workspaces" in the browser tree
+    Then I can see the browser title as "Workspaces - Nuxeo Platform"
```
semisse: Small nit:

```suggestion
    Then I click "Domain" in the browser tree
    and I can see the browser title as "Domain - Nuxeo Platform"
    And I can see the "Workspaces" browser tree node
    When I click "Workspaces" in the browser tree
    Then I can see the browser title as "Workspaces - Nuxeo Platform"
```

richardsd: makes sense, thank you

richardsd:
~~Requires https://github.com/nuxeo/nuxeo-web-ui/pull/1374 to fix the `a11y` checks.~~ PR merged.

elements/nuxeo-app.js
Gabez0r: > We broke the API here, a long time ago, not by my code.

Can you please explain how [that](https://github.com/nuxeo/nuxeo-web-ui/commit/9477681ba75818833bb483a3b42ff63344753ae3) break the API? It adds new API, but I don't see how it breaks any previous contract.

Right now, the `_updateTitle` runs whenever the `page` property, which is public, is updated. And your change removes this behavior, which means that there are situations now where the title won't update anymore. Try navigating to a document and then updating the `page` property of the `nuxeo-app` to see this in action (e.g. `Nuxeo.UI.app.page = 'home'`).

Gabez0r: Btw, my suggestion was the following: 
```diff
diff --git a/elements/nuxeo-app.js b/elements/nuxeo-app.js
index a8ccab60..b3fc3036 100644
--- a/elements/nuxeo-app.js
+++ b/elements/nuxeo-app.js
@@ -632,7 +632,7 @@ Polymer({
     'nuxeo-diff-documents': '_diffDocuments',
   },
 
-  observers: ['_computeSharedActionContext(currentUser)', '_updateTitle(page, i18n)'],
+  observers: ['_computeSharedActionContext(currentUser)', '_updateTitle(page, i18n, currentDocument, searchForm, currentTask, selectedAdminTab)'],
 
   ready() {
     this.$.drawerPanel.closeDrawer();

```

richardsd: > Can you please explain how [that](https://github.com/nuxeo/nuxeo-web-ui/commit/9477681ba75818833bb483a3b42ff63344753ae3) break the API? It adds new API, but I don't see how it breaks any previous contract.
> Right now, the `_updateTitle` runs whenever the `page` property, which is public, is updated. And your change removes this behavior, which means that there are situations now where the title won't update anymore. Try navigating to a document and then updating the `page` property of the `nuxeo-app` to see this in action (e.g. `Nuxeo.UI.app.page = 'home'`).

I say it broke the API because whenever you directly called the `show()` method, `_updateTitle()` was called as well to update the title. Nuno extended the API contract by having an observer to trigger the `_updateTitle()` based on `page` and `i18n` update, but since `page` doesn't change, the contract is broken.

semisse (approved):

elements/nuxeo-app.js
Gabez0r: I understand your point. So I guess we know have to keep both behaviors. :P

Gabez0r (approved):

PR #1374, opened 2022-02-01
WEBUI-666: update the number of a11y violations

richardsd:


Gabez0r (dismissed):

semisse:
Always good to see it decreasing!

semisse (dismissed):

richardsd:
> Always good to see it decreasing!

It seems that some are increasing, but I don't think this is our fault, because latest commits didn't change anything. Possibly coming from a dependency?

Gabez0r (approved):

semisse (approved):

PR #1375, opened 2022-02-02
WEBUI-392: remove unused d3 dependency

richardsd:


richardsd:
After some investigation, it looks that `d3` is no longer a dependency used in Web UI, so the best option is to remove the dependency at all from our code.
There's no import at all in Web UI and no reference in the code. As for the elements, they do their own imports with appropriate reference https://github.com/nuxeo/nuxeo-elements/blob/4cc6116fe896d2c6100b6b347ddc8389c26efe24/ui/package.json#L77-L79.

semisse (dismissed):

richardsd:
~~Requires https://github.com/nuxeo/nuxeo-web-ui/pull/1374 to be merged first.~~ Merged.

Gabez0r (dismissed):

semisse (approved):

Gabez0r (approved):

PR #1380, opened 2022-02-07
WEBUI-572: update docker action to v2

richardsd:


.github/workflows/preview.yaml
```diff
@@ -62,27 +62,34 @@ jobs:
...
+
+    - name: Login to DockerHub
+      uses: docker/login-action@v1
       with:
-        path: server
         registry: ${{ env.DOCKER_REGISTRY }}
         username: _json_key
         password: ${{ secrets.GKE_SA_KEY }}
+
+    - name: Build and push server image v2
```
Gabez0r: Do we need to keep this "v2" in the name?

.github/workflows/preview.yaml
```diff
@@ -62,27 +62,34 @@ jobs:
...
+      uses: docker/build-push-action@v2
+      with:
+        context: server
+        pull: true
         push: true
-        repository: ${{ secrets.GKE_PROJECT }}/webui/nuxeo-web-ui/server
-        tags: ${{ env.VERSION }}
+        file: server/Dockerfile
+        tags: ${{ env.DOCKER_REGISTRY }}/${{ secrets.GKE_PROJECT }}/webui/nuxeo-web-ui/server:${{ env.VERSION }}
```
Gabez0r: Isn't this the same as?
```suggestion
        tags: ${{ env.DOCKER_REGISTRY }}/${{ env. DOCKER_IMAGE }}/server:${{ env.VERSION }}
```

.github/workflows/preview.yaml
```diff
@@ -62,27 +62,34 @@ jobs:
...
 
-    - name: Build and push image
-      uses: docker/build-push-action@v1
+    - name: Build and push image v2
```
Gabez0r: Do we need to have this v2 here?

.github/workflows/preview.yaml
richardsd: Nope, forgot to remove the v2 😬

.github/workflows/preview.yaml
richardsd: Same as the other, to remove.

.github/workflows/preview.yaml
richardsd: It is in fact. I'll update this one too. Thanks

Gabez0r (approved):

semisse (approved):

PR #1382, opened 2022-02-10
WEBUI-600: support cross repo check

richardsd:
This one pairs with https://github.com/nuxeo/nuxeo-elements/pull/510, but the improvements in Web UI should be merged first.
Main changes:
- rename of the ondemand wf to `cross-repo`
- adds id job for remote triggering from `nuxeo-elements` cross-repo check
- update reference to `get-branch` action

### DIFF
--- a/.github/workflows/ondemand-main.yaml
+++ b/.github/workflows/ondemand-main.yaml
@@ -102,6 +102,9 @@ jobs:
           npm install
           popd
 
+      - name: Lint Web UI
+        run: npm run lint
+
       - name: Checkout the nuxeo-elements repo
         uses: actions/checkout@v2
         with:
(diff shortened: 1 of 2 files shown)


semisse (dismissed):

.github/workflows/ondemand-main.yaml
```diff
@@ -156,17 +137,6 @@ jobs:
...
-        if: ${{ github.event.inputs.skip_unit_tests == 'false' && github.event.inputs.sauce_labs == 'false' }}
-        run: npm run test
-
-      - name: Web UI Unit tests (Sauce Labs)
-        if: ${{ github.event.inputs.skip_unit_tests == 'false' && github.event.inputs.sauce_labs == 'true' }}
-        env:
-          SAUCE_USERNAME: nuxeo-web-ui
-          SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
-        run: npm run test
-
```
Gabez0r: Shouldn't we keep doing the unit tests? We might unit test an element that depends nuxeo-elements, which could break.

.github/workflows/ondemand-main.yaml
```diff
@@ -86,9 +102,6 @@ jobs:
           npm install
           popd
 
-      - name: Lint Web UI
-        run: npm run lint
-
```
Gabez0r: I think we should still lint web ui. There could be changes on elements that could trigger linting errors (renaming or moving methods among methods, for example).

Gabez0r (commented):
looking good! 💪

.github/workflows/ondemand-main.yaml
richardsd: You're right, totally forgot about that use case :/ I'll add the unit tests and linting back. Thanks

.github/workflows/ondemand-main.yaml
richardsd: ⬆️  I'll add the linting back.

semisse (dismissed):

Gabez0r (approved):

semisse (approved):

PR #1385, opened 2022-02-11
WEBUI-665: fix saved search title validation

Gabez0r:
Because the input was marked as `required` and `auto-validate`, it would be marked as invalid when empty (it just doesn't happen on the first time, not sure why). So, the fix consists of removing the `required`. However, this means that the user would be able to click the "Save" button even if the input value is empty, so I also made it disabled in this case. Please let me know what you think.

### DIFF
--- a/elements/search/nuxeo-search-form.js
+++ b/elements/search/nuxeo-search-form.js
@@ -377,12 +377,14 @@ Polymer({
         auto-validate
         error-message="[[i18n('searchForm.savePopup.error')]]"
         pattern="^[^/]*$"
-        required
+        value="{{_savedSearchTitle}}"
       >
       </paper-input>
       <div class="buttons">
         <paper-button dialog-dismiss class="secondary">[[i18n('command.cancel')]]</paper-button>
-        <paper-button noink class="primary" on-tap="_saveSearch">[[i18n('command.save')]]</paper-button>
+        <paper-button noink class="primary" on-tap="_saveSearch" disabled$="[[!_savedSearchTitle]]"
+          >[[i18n('command.save')]]</paper-button
+        >
       </div>
     </nuxeo-dialog>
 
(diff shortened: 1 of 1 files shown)


richardsd:
> Because the input was marked as `required` and `auto-validate`, it would be marked as invalid when empty (it just doesn't happen on the first time, not sure why). So, the fix consists of removing the `required`. However, this means that the user would be able to click the "Save" button even if the input value is empty, so I also made it disabled in this case. Please let me know what you think.

Your solution sounds reasonable to me 👍

richardsd (approved):
In practice, the only difference, is that we lose the highlighting when the input is empty, right?

Gabez0r:
> In practice, the only difference, is that we lose the highlighting when the input is empty, right?

@richardsd: yes, and the error message, which is a weird error message to have when the value is empty :P

richardsd:
> > In practice, the only difference, is that we lose the highlighting when the input is empty, right?
> 
> @richardsd: yes, and the error message, which is a weird error message to have when the value is empty :P

Ah you're right :P ltgm then

semisse (approved):

elements/search/nuxeo-search-form.js
```diff
@@ -377,12 +377,14 @@ Polymer({
...
-        required
+        value="{{_savedSearchTitle}}"
       >
       </paper-input>
       <div class="buttons">
         <paper-button dialog-dismiss class="secondary">[[i18n('command.cancel')]]</paper-button>
-        <paper-button noink class="primary" on-tap="_saveSearch">[[i18n('command.save')]]</paper-button>
+        <paper-button noink class="primary" on-tap="_saveSearch" disabled$="[[!_savedSearchTitle]]"
+          >[[i18n('command.save')]]</paper-button
+        >
```
semisse, line 387: nit: why is this closing tag breaking into a new line?

Gabez0r: Prettier 🤷

PR #1391, opened 2022-02-18
WEBUI-218: manually update results view on browser

Gabez0r:
It is a bit hacky, but the nuxeo browser only provides the document and the clipboard in the context, and the second doesn't make much sense to be used in the context of a `nuxeo-results-view`, so I'd say that there is no good reason no to update it if it is currently visible. It fixes the issue (apparently), it seems pretty harmless, but it's hacky...


richardsd (commented):
It seems harmless and hacky, but if it works 🤷

Gabez0r:
I proposed it as a custom code on the ticket. If it accepted as a workaround we won't have to merge, which is probably the safest path.

richardsd:
> I proposed it as a custom code on the ticket. If it accepted as a workaround we won't have to merge, which is probably the safest path.

I'm ok with that. Lets wait then.

PR #1394, opened 2022-02-22
WEBUI-308: display tags for users and groups in document tasks

richardsd:
This seems like a hacky solution for me.

I'd be ok with the purposed changes for `nuxeo-document-task` as it wouldn't mess any public API and it is an easy fix. However, having to change the `nuxeo-task6d8-layout` or any layout if necessary, should be a no go.

I thought about adding some logic to the `nuxeo-tags` instead, but since this is a dumb element without any logic, it means we would probably have to update (and break) its API.

Feedback of this is very welcome 👋  Thanks.

### DIFF
--- a/elements/workflow/nuxeo-document-task.js
+++ b/elements/workflow/nuxeo-document-task.js
@@ -163,12 +163,22 @@ Polymer({
           </div>
           <div id="assignedActors" class="vertical spaced">
             <span>[[i18n('tasks.actors.assigned')]]</span>
-            <nuxeo-tags type="user" items="[[task.actors]]"></nuxeo-tags>
+            <template is="dom-if" if="[[_hasActorType(task.actors, 'group')]]">
+              <nuxeo-tags type="group" items="[[_getActorsByType(task.actors, 'group')]]"></nuxeo-tags>  
+            </template>
+            <template is="dom-if" if="[[_hasActorType(task.actors, 'user')]]">
+              <nuxeo-tags type="user" items="[[_getActorsByType(task.actors, 'user')]]"></nuxeo-tags>
+            </template>
           </div>
           <template is="dom-if" if="[[_delegatedActorsExist(task.delegatedActors)]]">
             <div id="delegatedActors" class="vertical spaced">
               <span>[[i18n('tasks.actors.delegated')]]</span>
-              <nuxeo-tags type="user" items="[[task.delegatedActors]]"></nuxeo-tags>
+              <template is="dom-if" if="[[_hasActorType(task.delegatedActors, 'group')]]">
+                <nuxeo-tags type="user" items="[[_getActorsByType(task.delegatedActors, 'group')]]"></nuxeo-tags>
+              </template>
+              <template is="dom-if" if="[[_hasActorType(task.delegatedActors,'user')]]">
+                <nuxeo-tags type="user" items="[[_getActorsByType(task.delegatedActors, 'user')]]"></nuxeo-tags>
+              </template>
             </div>
           </template>
           <div class="vertical spaced">
@@ -322,5 +332,13 @@ Polymer({
 
   _isTaskInEndState(task) {
     return task && task.state === 'ended';
+  },
+
+  _hasActorType(actors, type) {
+    return actors && Array.isArray(actors) && actors.findIndex((actor) => actor['entity-type'] === type) >= 0;
+  },
+
+  _getActorsByType(actors, type) {
+    return actors && Array.isArray(actors) && actors.filter((actor) => actor['entity-type'] === type);
   },
 });
--- a/elements/workflow/serialdocumentreview/nuxeo-task6d8-layout.html
+++ b/elements/workflow/serialdocumentreview/nuxeo-task6d8-layout.html
@@ -23,9 +23,19 @@ limitations under the License.
 -->
 <dom-module id="nuxeo-task6d8-layout">
   <template>
-    <div role="widget">
+    <style>
+      .participants {
+        display: flex;
+      }
+    </style>
+    <div role="widget" class="participants">
       <label>[[i18n('wf.serialDocumentReview.participants')]]</label>
-      <nuxeo-tags type="user" items="[[task.variables.participants]]"></nuxeo-tags>
+      <template is="dom-if" if="[[_hasActorType(task.variables.participants, 'group')]]">
+        <nuxeo-tags type="group" items="[[_getActorsByType(task.variables.participants, 'group')]]"></nuxeo-tags>
+      </template>
+      <template is="dom-if" if="[[_hasActorType(task.variables.participants, 'user')]]">
+        <nuxeo-tags type="user" items="[[_getActorsByType(task.variables.participants, 'user')]]"></nuxeo-tags>
+      </template>
     </div>
 
     <div role="widget">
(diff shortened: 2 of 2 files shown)


elements/workflow/nuxeo-document-task.js
```diff
@@ -163,12 +163,22 @@ Polymer({
           </div>
           <div id="assignedActors" class="vertical spaced">
             <span>[[i18n('tasks.actors.assigned')]]</span>
-            <nuxeo-tags type="user" items="[[task.actors]]"></nuxeo-tags>
+            <template is="dom-if" if="[[_hasActorType(task.actors, 'group')]]">
+              <nuxeo-tags type="group" items="[[_getActorsByType(task.actors, 'group')]]"></nuxeo-tags>  
+            </template>
+            <template is="dom-if" if="[[_hasActorType(task.actors, 'user')]]">
+              <nuxeo-tags type="user" items="[[_getActorsByType(task.actors, 'user')]]"></nuxeo-tags>
+            </template>
```
Gabez0r, line 171: Can't we go with a computed binding that computes the type based on `task.actors`? Something like:
```suggestion
            <nuxeo-tags type="[[_computeTagType(task.actors)]]" items="[[task.actors]]"></nuxeo-tags>
```

elements/workflow/nuxeo-document-task.js
```diff
@@ -163,12 +163,22 @@ Polymer({
...
           <template is="dom-if" if="[[_delegatedActorsExist(task.delegatedActors)]]">
             <div id="delegatedActors" class="vertical spaced">
               <span>[[i18n('tasks.actors.delegated')]]</span>
-              <nuxeo-tags type="user" items="[[task.delegatedActors]]"></nuxeo-tags>
+              <template is="dom-if" if="[[_hasActorType(task.delegatedActors, 'group')]]">
+                <nuxeo-tags type="user" items="[[_getActorsByType(task.delegatedActors, 'group')]]"></nuxeo-tags>
+              </template>
+              <template is="dom-if" if="[[_hasActorType(task.delegatedActors,'user')]]">
+                <nuxeo-tags type="user" items="[[_getActorsByType(task.delegatedActors, 'user')]]"></nuxeo-tags>
+              </template>
```
Gabez0r, line 181: Same here.

elements/workflow/serialdocumentreview/nuxeo-task6d8-layout.html
```diff
@@ -25,7 +25,12 @@
   <template>
     <div role="widget">
       <label>[[i18n('wf.serialDocumentReview.participants')]]</label>
-      <nuxeo-tags type="user" items="[[task.variables.participants]]"></nuxeo-tags>
+      <template is="dom-if" if="[[_hasActorType(task.variables.participants, 'group')]]">
+        <nuxeo-tags type="group" items="[[_getActorsByType(task.variables.participants, 'group')]]"></nuxeo-tags>
+      </template>
+      <template is="dom-if" if="[[_hasActorType(task.variables.participants, 'user')]]">
+        <nuxeo-tags type="user" items="[[_getActorsByType(task.variables.participants, 'user')]]"></nuxeo-tags>
+      </template>
```
Gabez0r, line 38: Same here.

Gabez0r (commented):
We need to document in the ticket that people might have to update their layouts if they want the complete fix.

elements/workflow/nuxeo-document-task.js
richardsd: 🤔 Doesn't work if you have assigned groups and individual users simultaneously. Should we care about this one?

elements/workflow/nuxeo-document-task.js
richardsd: Same as above.

elements/workflow/serialdocumentreview/nuxeo-task6d8-layout.html
richardsd: Same as above.

elements/workflow/nuxeo-document-task.js
Gabez0r: Ah, I understand now. We should care and this is a possibility, indeed. It seems that we are just fighting against the `nuxeo-tags` element. But I have no other suggestion, then. The alternative would be to add some sort of type resolver to the `nuxeo-tags`, but that by itself ins an improvement and has other implications to the public API of the element.

elements/workflow/nuxeo-document-task.js
```diff
@@ -323,4 +333,12 @@ Polymer({
   _isTaskInEndState(task) {
     return task && task.state === 'ended';
   },
+
+  _hasActorType(actors, type) {
+    return actors && actors.length > 0 && actors.findIndex((actor) => actor['entity-type'] === type) >= 0;
```
Gabez0r: nit:
```suggestion
    return actors && Array.isArray(actors.length) && actors.findIndex((actor) => actor['entity-type'] === type) >= 0;
```

elements/workflow/nuxeo-document-task.js
```diff
@@ -323,4 +333,12 @@ Polymer({
...
+  },
+
+  _getActorsByType(actors, type) {
+    return actors && actors.length > 0 && actors.filter((actor) => actor['entity-type'] === type);
```
Gabez0r: nit:
```suggestion
    return actors && Array.isArray(actors) && actors.filter((actor) => actor['entity-type'] === type);
```

Gabez0r (approved):

elements/workflow/nuxeo-document-task.js
richardsd: 👍 It makes sense.

elements/workflow/nuxeo-document-task.js
richardsd: Same as above here.

elements/workflow/nuxeo-document-task.js
richardsd: Exactly. In fact that was my first idea, to extend the public API of the `nuxeo-tags` element by adding a type resolver. What if we keep the same public API, but extend the `type` attribute with `auto` as a possible value, and  move all the logic to inside the `nuxeo-tags`?

richardsd: So `type` could be `user`, `group` and `auto`.

Gabez0r: I think that makes perfect sense, but perhaps we should do it as a follow up ticket? In the end, people will still be forced to update their layouts, right? If so, we won't get any immediate benefit to committing to this new API.

richardsd: No, you're totally right, we can open an improvement ticket to add this new API, as we'd still need to update the layouts.

elements/workflow/nuxeo-document-task.js
Gabez0r: This change didn't happen, right?

Gabez0r (dismissed):

elements/workflow/nuxeo-document-task.js
richardsd: No, I didn't update the code yet.

elements/workflow/nuxeo-document-task.js
richardsd: FYI https://jira.nuxeo.com/browse/ELEMENTS-1467

Gabez0r (approved):

semisse (approved):

PR #1398, opened 2022-02-22
WEBUI-592: introduce property to config headers used in user management

richardsd:
This pairs with https://github.com/nuxeo/nuxeo-elements/pull/515 and exposes a configuration property to control the headers used to fetch user data in the `nuxeo-user-management`.

With this approach we keep the same behavior we have right now, without having to update any layout (this means that if someone wants to take advantage of this, they'll have to update the layouts).

### DIFF
--- a/plugin/web-ui/addon/src/main/resources/OSGI-INF/web-ui-properties.xml
+++ b/plugin/web-ui/addon/src/main/resources/OSGI-INF/web-ui-properties.xml
@@ -44,5 +44,8 @@
 
     <!-- Control the enablement of select all -->
     <property name="org.nuxeo.web.ui.selection.selectAllEnabled">${nuxeo.selection.selectAllEnabled:=false}</property>
+
+    <!-- Properties to be fetched when loading the user object in user management, default is empty -->
+    <property name="org.nuxeo.web.ui.user.management.fetch.document" list="true"></property>
   </extension>
 </component>
--- a/index.html
+++ b/index.html
@@ -134,6 +134,13 @@ limitations under the License.
             /* selection: { // uncomment to enable select all items in listings
               selectAllEnabled: true
             }, */
+            /*user: {
+              management: {
+                fetch: {
+                  document: ['properties'],
+                },
+              },
+            },*/
           },
         },
       };


plugin/web-ui/addon/src/main/resources/OSGI-INF/web-ui-properties.xml
```diff
@@ -44,5 +44,8 @@
 
     <!-- Control the enablement of select all -->
     <property name="org.nuxeo.web.ui.selection.selectAllEnabled">${nuxeo.selection.selectAllEnabled:=false}</property>
+
+    <!-- Enable additional headers to enrich user in user management -->
+    <property name="org.nuxeo.web.ui.user.management.headers">${nuxeo.user.management.headers}</property>
```
Gabez0r: I think that what we want to do here is precisely the same we are doing for the `fetch` property (see [code](https://github.com/nuxeo/nuxeo-web-ui/blob/maintenance-3.0.x/plugin/web-ui/addon/src/main/resources/OSGI-INF/web-ui-properties.xml#L27)). And it would make sense because this is collection of headers, and we could want to contribute to them from different projects. Therefore, perhaps we don't even have (or should) expose it in the nuxeo.conf.

richardsd: Yes, probably you're right. ~~Can you point me where we load this?
I see a contribution in dev mode [here](https://github.com/nuxeo/nuxeo-web-ui/blob/aa606019bbd0fb4acca83662f35b0db972a439d6/index.html#L93-L95) but that's it.~~
Let me review it.

richardsd: I just updated the code with your suggestions Gabe.

Gabez0r (dismissed):
LGTM. Please add to the ticket an example of a contribution adding the properties fetcher. And please, also add to the update notes of the ticket that we are adding this new property.

richardsd:
An example of how to add `properties` to `fetch-document` using the new configuration property:

```
<?xml version="1.0" encoding="UTF-8"?>
<component name="org.nuxeo.web.ui.user.management.config.example">
    <property name="org.nuxeo.web.ui.user.management.fetch.document" list="true">properties</property>
  </extension>
</component>

```

richardsd:
WEBUI-592

plugin/web-ui/addon/src/main/resources/OSGI-INF/web-ui-properties.xml
```diff
@@ -44,5 +44,8 @@
 
     <!-- Control the enablement of select all -->
     <property name="org.nuxeo.web.ui.selection.selectAllEnabled">${nuxeo.selection.selectAllEnabled:=false}</property>
+
+    <!-- Properties to be fetched when loading user management, default is empty -->
```
Gabez0r: Probably something a bit more concrete regarding what we are loading?:
```suggestion
    <!-- Properties to be fetched when loading the user object in user management, default is empty -->
```

Gabez0r (approved):

semisse (approved):

PR #1399, opened 2022-02-23
WEBUI-351: add css variable to control html editor height

Gabez0r:
The height of the `nuxeo-html-editor` inside `nuxeo-note-editor` can now be controlled using a new css variable:
`--nuxeo-note-editor-html-height`

For example, you can set the following on a theme:
```css
--nuxeo-note-editor-html-height: calc(80vh - 132px);
```

Depends on https://github.com/nuxeo/nuxeo-elements/pull/516.

### DIFF
--- a/elements/nuxeo-note-editor/nuxeo-note-editor.js
+++ b/elements/nuxeo-note-editor/nuxeo-note-editor.js
@@ -67,10 +67,14 @@ Polymer({
       }
 
       paper-textarea,
-      nuxeo-document-preview,
-      nuxeo-html-editor {
+      nuxeo-document-preview {
         display: block;
         min-height: calc(80vh - 132px);
+      }
+
+      nuxeo-html-editor {
+        min-height: calc(80vh - 132px);
+        height: var(--nuxeo-note-editor-html-height);
       }
 
       paper-textarea {


richardsd (approved):

elements/nuxeo-note-editor/nuxeo-note-editor.js
```diff
@@ -67,12 +67,16 @@ Polymer({
...
-      nuxeo-document-preview,
-      nuxeo-html-editor {
+      nuxeo-document-preview {
         display: block;
         min-height: calc(80vh - 132px);
       }
 
+      nuxeo-html-editor {
+        min-height: calc(80vh - 132px);
+        height: var(--nuxeo-note-editor-html-height);
```
semisse, line 77: We should have a default value here

Gabez0r: We should only have a default value if is different than the height's default, right?

semisse: Yes, but having a default value helps readability

Gabez0r: `height: var(--nuxeo-note-editor-html-height, auto);` seems redundant to me when compared to `height: var(--nuxeo-note-editor-html-height);`. It's not like we would use `height: auto`  here if we weren't using the var. 🤔

semisse (approved):

PR #1405, opened 2022-03-03
WEBUI-706: prevent vocabulary creation with whitespaces only in id field

richardsd:
The pattern ensures that we have at least a non whitespace character for the `id` field.

Current behavior: we can have ids made of whitespaces only
New behavior: we need at least a non whitespace character

Feedback is very welcome 🙂

### DIFF
--- a/elements/directory/vocabulary/nuxeo-vocabulary-edit-layout.html
+++ b/elements/directory/vocabulary/nuxeo-vocabulary-edit-layout.html
@@ -31,6 +31,8 @@ limitations under the License.
       value="{{entry.properties.id::change}}"
       required
       readonly$="[[!new]]"
+      pattern=".*\S.*"
+      error-message="[[i18n('vocabularyManagement.edit.error')]]"
     >
     </nuxeo-input>
 
--- a/i18n/messages.json
+++ b/i18n/messages.json
@@ -1440,6 +1440,7 @@
   "vocabularyManagement.confirmDelete": "Delete vocabulary entry?",
   "vocabularyManagement.deleteEntry": "Delete Entry",
   "vocabularyManagement.edit.actions": "Actions",
+  "vocabularyManagement.edit.error": "ID cannot be empty",
   "vocabularyManagement.edit.id": "ID",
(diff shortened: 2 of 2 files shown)


richardsd:
WEBUI-706

elements/directory/vocabulary/nuxeo-vocabulary-edit-layout.html
```diff
@@ -31,6 +31,7 @@
       value="{{entry.properties.id::change}}"
       required
       readonly$="[[!new]]"
+      pattern=".*\S.*"
```
Gabez0r, line 34: Shouldn't we also have an `error-message` displaying what the error is about? It might not be clear to the user why the field is marked as invalid.

richardsd: I agree, it probably should have a message like: 

`ID should have at least a non whitespace character`

wydt? (should we ping the PO to decide the message?)

Gabez0r: Yes, something like that. Or "ID cannot be empty". Please confirm wit the PO.

richardsd: @bchauvin we need your input on the message content.

bchauvin: @richardsd "ID cannot be empty" seems good to me, simpler and clearer :-)

richardsd: Thanks, I seems good to me too 👍

richardsd:
WEBUI-706-vocabularies

Gabez0r (approved):
Once merged, please don't forget to follow the [translation process](https://nuxeowiki.atlassian.net/wiki/spaces/NuxEng/pages/3193340004/Translations+for+Web+UI+and+Elements#Ensuring-100%25-translation-rate).

semisse (approved):

PR #1418, opened 2022-03-24
WEBUI-721: fix document import tab

richardsd:


richardsd:
webui-721

elements/document/nuxeo-document-import.js
```diff
@@ -372,8 +372,6 @@ Polymer({
     <nuxeo-operation id="fileManagerImport" op="FileManager.Import" sync-indexing></nuxeo-operation>
     <nuxeo-document
       id="docRequest"
-      doc-path="[[targetPath]]"
```
Gabez0r, line 375: Please note that the targetPath is also changed by `Nuxeo.DocumentCreationBehavior`.

richardsd: It doesn't seem to have an effect in the import tab, thank you for the review 👍

elements/document/nuxeo-document-import.js
```diff
@@ -1266,18 +1279,22 @@ Polymer({
           .forEach((index) => {
             this.splice('remoteFiles', index, 1);
           });
+        /*
+         * Prevent this._selectDoc(0) from storing the previously selected file,
```
Gabez0r: nit: I would even put an "XXX" here, so we could find this later :P

Gabez0r (dismissed):
Great job! 👏 💪

elements/document/nuxeo-document-import.js
richardsd: Let me add that indeed. Thanks 😄

Gabez0r (approved):

semisse (approved):

PR #1420, opened 2022-03-28
WEBUI-721: display validation errors in document import - 10.10

richardsd:
Cross-repo check https://github.com/nuxeo/nuxeo-web-ui/actions/runs/2051730363 🔵 
Backport of https://github.com/nuxeo/nuxeo-web-ui/pull/1418 (read comment bellow).

### DIFF
--- a/elements/document/nuxeo-document-import.html
+++ b/elements/document/nuxeo-document-import.html
@@ -1205,7 +1205,10 @@ limitations under the License.
         }
         Promise.all(promises).then(function(results) {
           var errorFree = results.filter(function(result) {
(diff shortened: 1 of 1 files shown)


richardsd:
Unfortunately, `nuxeo-document-layout` doesn't provide the `reportValidation` method that was introduced in `maintenance-3.0.x`, which means we can't do better in this case than just displaying the generic error message.

To improve this behavior, we would need to backport the public API of the `nuxeo-document-layout` element.

Gabez0r (approved):

semisse (approved):

PR #1422, opened 2022-03-29
WEBUI-724: add pr checks for 10.10

richardsd:


.github/workflows/ftest.yaml
```diff
@@ -0,0 +1,320 @@
...
+    branches:
+      - '10.10'
+  pull_request:
+    branches:
+      - '10.10'
+
+env:
+  REFERENCE_BRANCH: '10.10'
+  ELEMENTS_REFERENCE_BRANCH: 'maintenance-2.4.x'
+  NPM_REPOSITORY: https://packages.nuxeo.com/repository/npm-public/
```
semisse: don't we need to pass this value as a string?

.github/workflows/ftest.yaml
```diff
@@ -0,0 +1,320 @@
...
+          bower install && npm install
+          popd
+
+      - name: Lint Web UI
+        run: |
+          pushd nuxeo-web-ui
+          npm run lint
+          popd
+
+      - name: Link elements to Web UI
```
semisse:
```suggestion
      - name: Link Nuxeo Elements to Web UI
```

semisse (dismissed):

.github/workflows/ftest.yaml
richardsd: No, everything is a string, we should probably try to remove all the '' in here.

.github/workflows/ftest.yaml
richardsd: I didn't use `Nuxeo Elements` just to make it clear that we are linking all the core, ui and dataviz elements. 
Do you think that using `Nuxeo Elements` is clear enough? 
Thanks Sam

.github/workflows/ftest.yaml
```diff
@@ -0,0 +1,320 @@
...
+      - name: Archive packages
+        if: always()
+        uses: actions/upload-artifact@v2
+        with:
+          name: packages
+          path: |
+            # HACK - removed the SNAPSHOT because of sed
+            # plugin-nuxeo-web-ui/marketplace/target/nuxeo-web-ui-marketplace-*-SNAPSHOT.zip
+            plugin-nuxeo-web-ui/marketplace/target/nuxeo-web-ui-marketplace-*.zip
+            plugin-nuxeo-web-ui/marketplace-itests/target/nuxeo-web-ui-marketplace-itests-*.zip
```
richardsd: Missing a new line here.

semisse (approved):

.github/workflows/ftest.yaml
```diff
@@ -0,0 +1,309 @@
...
+                <server>
+                  <id>maven-internal</id>
+                  <username>${{ secrets.PACKAGES_AUTH_USER }}</username>
+                  <password>${{ secrets.PACKAGES_AUTH_TOKEN }}</password>
+                </server>
+              </servers>
+              </settings>' > ~/.m2/settings.xml
+
+      - name: Checkout the nuxeo-elements repo
+        uses: actions/checkout@v2
```
Gabez0r, line 121: Any particular reason not to be using v3? I'm ok with using v2 for consistency with other, btw.

Gabez0r (approved):
LGTM

.github/workflows/ftest.yaml
richardsd: No particular reason. It was released last month and the major difference is the nodejs runtime update to version 16 https://github.com/actions/checkout#whats-new. I'm ok with using v2 for consistency too.

PR #1432, opened 2022-04-21
WEBUI-209: improve a11y in nuxeo-document-tree

semisse:


richardsd (dismissed):

richardsd (approved):

elements/nuxeo-document-tree/nuxeo-document-tree.js
```diff
@@ -183,13 +183,15 @@ Polymer({
...
-            <paper-spinner active$="[[loading]]"></paper-spinner>
-            <iron-icon icon="[[_expandIcon(opened)]]" toggle hidden$="[[loading]]"></iron-icon>
-          </template>
-          <span class="node-name flex">
-            <a href$="[[urlFor(item)]]">[[_title(item)]]</a>
-          </span>
+          <div role="treeitem" aria-expanded="[[opened]]">
+            <template class="flex" is="dom-if" if="[[!isLeaf]]">
+              <paper-spinner active$="[[loading]]" aria-hidden="true"></paper-spinner>
+              <iron-icon icon="[[_expandIcon(opened)]]" toggle hidden$="[[loading]]" aria-hidden="true"></iron-icon>
```
Gabez0r, line 189: Why are these `aria-hidden`? Won't this prevent the icon from being clickable using assistive technology, to expand the node?

Gabez0r: I'm understand that this preserves a current limitation. So, I suggest we create an improvement ticket for this and we can merge as is.

Gabez0r (approved):

elements/nuxeo-document-tree/nuxeo-document-tree.js
semisse: Yes, so `aria-hidden` is a conservative approach to improve a11y in this element, since the `iron-icon` is not focusable. We need to address this problem in a different context since it is also related with other improvements we need to introduce in order to alert the user for the expanded level and status (opened/closed)

PR #1433, opened 2022-04-22
WEBUI-769: nuxeo-document-tree unit tests

richardsd:
Feedback is more than welcome, because there's a lot of space for improvements :)

### DIFF
--- a/test/nuxeo-document-tree.test.js
+++ b/test/nuxeo-document-tree.test.js
@@ -1,0 +1,346 @@
+/**
+@license
+(C) Copyright Nuxeo Corp. (http://nuxeo.com/)
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+import { fixture, flush, html, isElementVisible, login } from '@nuxeo/testing-helpers';
+import '@nuxeo/nuxeo-ui-elements/nuxeo-icons.js';
+import '@polymer/iron-icons/iron-icons.js';
+import '@polymer/iron-icons/hardware-icons.js';
+import '../elements/nuxeo-document-tree/nuxeo-document-tree.js';
+
+// mock the label for the document tree root
+window.nuxeo.I18n.language = 'en';
+window.nuxeo.I18n.en = window.nuxeo.I18n.en || {};
+window.nuxeo.I18n.en['browse.root'] = 'Root';
+
+const jsonHeader = { 'Content-Type': 'application/json' };
+
+// root document definition
+const rootDocument = {
+  'entity-type': 'document',
+  repository: 'default',
+  uid: '1',
+  path: '/',
+  type: 'Root',
+  parentRef: '/',
+  title: 'Root',
+  properties: {},
+  facets: ['Folderish', 'NXTag', 'NotCollectionMember'],
+  schemas: [
+    {
+      name: 'common',
+      prefix: 'common',
+    },
+    {
+      name: 'dublincore',
+      prefix: 'dc',
+    },
+    {
+      name: 'facetedTag',
+      prefix: 'nxtag',
+    },
+  ],
+  contextParameters: {
+    hasFolderishChild: true,
+  },
+};
+
+// root content children
+const rootChildrenResponse = {
+  'entity-type': 'documents',
+  isPaginable: true,
+  resultsCount: 3,
+  pageSize: 40,
+  maxPageSize: 40,
+  resultsCountLimit: 40,
+  currentPageSize: 3,
+  currentPageIndex: 0,
+  currentPageOffset: 0,
+  numberOfPages: 1,
+  isPreviousPageAvailable: false,
+  isNextPageAvailable: false,
+  isLastPageAvailable: false,
+  isSortable: true,
+  totalSize: 3,
+  pageIndex: 0,
+  pageCount: 1,
+  entries: [
+    {
+      'entity-type': 'document',
+      repository: 'default',
+      uid: '2',
+      path: '/default-domain/folder1',
+      type: 'Folder',
+      parentRef: '1',
+      title: 'Folder 1',
+      properties: {},
+      facets: ['Folderish', 'NXTag', 'SuperSpace', 'HiddenInCreation', 'NotCollectionMember'],
+      schemas: [
+        {
+          name: 'common',
+          prefix: 'common',
+        },
+        {
+          name: 'dublincore',
+          prefix: 'dc',
+        },
+        {
+          name: 'facetedTag',
+          prefix: 'nxtag',
+        },
+      ],
+    },
+    {
+      'entity-type': 'document',
+      repository: 'default',
+      uid: '3',
+      path: '/default-domain/folder2',
+      type: 'Folder',
+      parentRef: '1',
+      title: 'Folder 2',
+      properties: {},
+      facets: ['Folderish', 'NXTag', 'SuperSpace', 'HiddenInCreation', 'NotCollectionMember', 'MasterPublishSpace'],
+      schemas: [
+        {
+          name: 'common',
+          prefix: 'common',
+        },
+        {
+          name: 'dublincore',
+          prefix: 'dc',
+        },
+        {
+          name: 'facetedTag',
+          prefix: 'nxtag',
+        },
+      ],
+    },
+    {
+      'entity-type': 'document',
+      repository: 'default',
+      uid: '4',
+      path: '/default-domain/folder3',
+      type: 'Folder',
+      parentRef: '1',
+      title: 'Folder 3',
+      properties: {},
+      facets: ['Folderish', 'NXTag', 'SuperSpace', 'HiddenInCreation', 'NotCollectionMember'],
+      schemas: [
+        {
+          name: 'common',
+          prefix: 'common',
+        },
+        {
+          name: 'dublincore',
+          prefix: 'dc',
+        },
+        {
+          name: 'facetedTag',
+          prefix: 'nxtag',
+        },
+      ],
+      contextParameters: {
+        hasFolderishChild: true,
+      },
+    },
+  ],
+};
+
+// folder 3 children
+const folderChildrenResponse = {
+  'entity-type': 'documents',
+  isPaginable: true,
+  resultsCount: 2,
+  pageSize: 40,
+  maxPageSize: 40,
+  resultsCountLimit: 40,
+  currentPageSize: 2,
+  currentPageIndex: 0,
+  currentPageOffset: 0,
+  numberOfPages: 1,
+  isPreviousPageAvailable: false,
+  isNextPageAvailable: false,
+  isLastPageAvailable: false,
+  isSortable: true,
+  hasError: false,
+  errorMessage: null,
+  totalSize: 2,
+  pageIndex: 0,
+  pageCount: 1,
+  entries: [
+    {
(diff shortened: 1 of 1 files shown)


test/nuxeo-document-tree.test.js
```diff
@@ -0,0 +1,456 @@
...
+const getTreeNodeByUid = (el, uid) => el.shadowRoot.querySelector(`nuxeo-tree-node[data-uid="${uid}"]`);
+
+const waitForTreeNodeLoading = async (tree, node = null) => {
+  const rootNode = node || getTreeRoot(tree);
+  const spinner = getNodeLoadingSpinner(rootNode);
+  expect(isElementVisible(rootNode)).to.be.true;
+  if (rootNode.loading) {
+    await waitForAttrMutation(spinner, 'active', null);
+  }
+  expect(isElementVisible(rootNode)).to.be.true;
```
Gabez0r: Why are we doing `expect(isElementVisible(rootNode)).to.be.true;` before and after the spinner being active?

test/nuxeo-document-tree.test.js
```diff
@@ -0,0 +1,456 @@
...
+window.nuxeo.I18n.en['browse.root'] = 'Root';
+
+const jsonHeader = { 'Content-Type': 'application/json' };
+
+// generate a document for the tree
+const generateDocument = ({ uid, type, parentRef, parentPath, isFolderish }) => {
+  const doc = {
+    'entity-type': 'document',
+    repository: 'default',
+    uid,
```
Gabez0r, line 53: Could we make this automatic, with a counter, so that it is one less parameter to specify? I guess this wouldn't make sense if we have to keep parentRef.

test/nuxeo-document-tree.test.js
```diff
@@ -0,0 +1,456 @@
...
+    path: type === 'Root' ? '/' : `/${parentPath}/${type}${uid}`,
+    type,
+    parentRef,
```
Gabez0r, line 56: Do we need to have the parentRef?

test/nuxeo-document-tree.test.js
```diff
@@ -0,0 +1,456 @@
...
+    resultsCountLimit: 40,
+    currentPageSize: entries.length,
+    currentPageIndex: 0,
+    currentPageOffset: 0,
+    numberOfPages: 1,
+    isPreviousPageAvailable: false,
+    isNextPageAvailable: false,
+    isLastPageAvailable: false,
+    isSortable: true,
+    totalSize: entries,
```
Gabez0r: Shouldn't this be?:
```suggestion
    totalSize: entries.length,
```

test/nuxeo-document-tree.test.js
```diff
@@ -0,0 +1,456 @@
...
+      true,
+    );
+    documentTree.document = rootDocument;
+    await flush();
+    // wait for the tree to finish loading
+    await waitForTreeNodeLoading(documentTree);
+
+    // assert that the nodes have been rendered
+    const nodes = getTreeNodes(documentTree);
+    assert.equal(nodes.length, 4, 'Tree nodes not loaded yet.');
```
Gabez0r: We are mixing assert with expect. Perhaps we could use just one of them?

test/nuxeo-document-tree.test.js
```diff
@@ -0,0 +1,456 @@
...
+
+const getTreeRoot = (el) => el.$$('nuxeo-tree-node#root');
+const getNodeLoadingSpinner = (el) => el.querySelector('paper-spinner');
+const getTreeNodes = (el) => el.shadowRoot.querySelectorAll('nuxeo-tree-node');
+const getTreeNodeByUid = (el, uid) => el.shadowRoot.querySelector(`nuxeo-tree-node[data-uid="${uid}"]`);
+
+const waitForTreeNodeLoading = async (tree, node = null) => {
+  const rootNode = node || getTreeRoot(tree);
+  const spinner = getNodeLoadingSpinner(rootNode);
+  expect(isElementVisible(rootNode)).to.be.true;
```
Gabez0r, line 121: I find it weird to have a waitFor* helper method that does an assertion, particularly before actually waiting. I would expect it to return something, or throw something, which could then be asserted upon on the test method.

test/nuxeo-document-tree.test.js
```diff
@@ -0,0 +1,456 @@
...
+  expect(rootNode.loading).to.be.false;
+};
+
+const getDocumentByPath = (path) => {
+  let document;
+  const parts = path.split('/');
+  if (parts.length > 1) {
+    document = levelOneDocuments.find((doc) => doc.title === parts[1]);
+    expect(document).to.be.not.null;
```
Gabez0r: Same here. This is a getter method which actually does a bunch of assertions.

test/nuxeo-document-tree.test.js
```diff
@@ -0,0 +1,456 @@
...
+    await flush();
+    // wait for the tree to finish loading
+    await waitForTreeNodeLoading(documentTree);
+
+    // assert that the nodes have been rendered
+    const nodes = getTreeNodes(documentTree);
+    assert.equal(nodes.length, 4, 'Tree nodes not loaded yet.');
+    nodes.forEach((node) => {
+      expect(isElementVisible(node)).to.be.true;
+    });
```
Gabez0r: We usually don't do asserts on the setup. This means that we will always be asserting the same thing, since setup is not parameterizable. Consider moving this to a dedicated test.

test/nuxeo-document-tree.test.js
```diff
@@ -0,0 +1,456 @@
...
+      // icon name should not be empty
+      expect(iconName).to.be.not.empty;
+      // tap to open the node
+      tap(node.querySelector('iron-icon'));
+      // node should now be opened
+      expect(node.opened).to.be.true;
+      // icon should have been updated to something different
+      const openedIcon = node.querySelector('iron-icon');
+      expect(openedIcon.icon).to.be.not.empty;
+      expect(iconName).to.be.not.equal(openedIcon.icon);
```
Gabez0r: We should compare this to what we are expecting. And we should probably compare that we are using the correct one too before clicking.

test/nuxeo-document-tree.test.js
```diff
@@ -0,0 +1,456 @@
...
+      expect(nodes).to.be.not.null;
+      assert.equal(nodes.length, 8, 'There are nodes missing in the tree.');
+      nodes.forEach((n) => {
+        const anchor = n.querySelector('.node-name a');
+        anchor.setAttribute('href', anchor.href.substring(anchor.href.indexOf('javascript'), anchor.href.length));
+      });
+      const nodeToClick = getTreeNodeByUid(documentTree, '7');
+      expect(nodeToClick).to.be.not.null;
+      // click the anchor
+      nodeToClick.querySelector('a').click();
```
Gabez0r: Wouldn't `tap` work here?

test/nuxeo-document-tree.test.js
```diff
@@ -0,0 +1,456 @@
...
+      [...levelThreeDocuments, levelTwoDocuments[1]].forEach((document) => {
+        const n = getTreeNodeByUid(documentTree, document.uid);
+        expect(n).to.be.not.null;
+        expect(isElementVisible(n)).to.be.true;
+      });
+      // assert that we have one parent only and it is visible
+      expect(documentTree.parents).to.be.not.empty;
+      assert.equal(documentTree.parents.length, 1);
+      assert.equal(documentTree.parents[0].uid, '4');
+      isElementVisible(documentTree.$$('.parents'));
```
Gabez0r: we should be consistent and replace the `$$` with the querySelector, since this is what we have been doing in other places.

test/nuxeo-document-tree.test.js
```diff
@@ -0,0 +1,456 @@
...
+    ];
+
+    // document definition for: root -> documents -> documents -> documents
+    levelThreeDocuments = [
+      generateDocument({ uid: '7', type: 'File', parentRef: '6', parentPath: '/Folder6/Folder7', isFolderish: false }),
+      generateDocument({ uid: '8', type: 'File', parentRef: '6', parentPath: '/Folder6/Folder8', isFolderish: false }),
+    ];
+
+    // set the breadcrumb for some documents
+    levelTwoDocuments[1].contextParameters.breadcrumb = {
```
Gabez0r: We should consider assigning `levelTwoDocuments[1]` and `levelOneDocuments[2]` to variables, since they are reference several times in the code. This would keep things DRY.

test/nuxeo-document-tree.test.js
```diff
@@ -0,0 +1,456 @@
...
+      await waitForTreeNodeLoading(documentTree, node);
+
+      // XXX - update the href to prevent redirect
+      let nodes = getTreeNodes(documentTree);
+      expect(nodes).to.be.not.null;
+      assert.equal(nodes.length, 8, 'There are nodes missing in the tree.');
+      nodes.forEach((n) => {
+        const anchor = n.querySelector('.node-name a');
+        anchor.setAttribute('href', anchor.href.substring(anchor.href.indexOf('javascript'), anchor.href.length));
+      });
```
Gabez0r: Couldn't we try to make the router just return a # and then here intercept the click and do the `documentTree.currentDocument = getDocumentByPath(path);` logic? Perhaps it would simplify things and we don't need to test that the anchors work (we could still test the href on another test if we want, but we weren't testing it here either).

test/nuxeo-document-tree.test.js
richardsd: You're right 😬

test/nuxeo-document-tree.test.js
richardsd: Yes, it makes sense. I'll use the `expect` for all the use cases.

test/nuxeo-document-tree.test.js
richardsd: You mean to hardcode the actual icon string name here and compare to it?

test/nuxeo-document-tree.test.js
richardsd: It would indeed 😬  Thanks, I missed this one.

test/nuxeo-document-tree.test.js
richardsd: I can replace the `$$` with `querySelector` along with `shadowRoot`, otherwise, the result won't be the same, but notice that, in some cases, I used querySelector directly instead of `$$` because I'm not going through the shadow root to get the elements, but rather selecting them from the light DOM (or using `querySelectorAll`).

test/nuxeo-document-tree.test.js
richardsd: ~~I can try to review it yes and simplify the code. Let me check what I can do. Thanks.~~ Not sure we can just over simplify it like this because, we still to pass the path from the router to the method `getDocumentByPath(path)`.

test/nuxeo-document-tree.test.js
richardsd: Yes we need, to keep the structure like something `/Folder1/File2`.

test/nuxeo-document-tree.test.js
richardsd: Not sure, I first had individual documents hardcoded, and only refactored the code later, so I'll need to check what can I do to simplify it.

test/nuxeo-document-tree.test.js
richardsd: Ah, I probably forgot to remove it, before doing the refactoring.

test/nuxeo-document-tree.test.js
richardsd: Done 👍

test/nuxeo-document-tree.test.js
richardsd: Done, let me know what you think about this approach. It looks much cleaner for me, thank you so much for the feedback.

test/nuxeo-document-tree.test.js
richardsd: This part of the code was like this because it was part of a test, in fact we don't need it.

Gabez0r (approved):
LGTM

semisse (approved):

PR #1442, opened 2022-05-24
WEBUI-739: prevent bulk edit layout from loading on documents change

richardsd:
This PR proposes two changes to the edit documents button:
- prevent reloading the layout if the href is exactly the same that was previously used in `nuxeo-layout`
- prevent reloading the layout everytime a new document is selected (removing the `documents.splices` from the observer)


Gabez0r (approved):
LGTM

semisse (approved):

PR #1444, opened 2022-05-26
SUPINT-2107: fix date-picker ftest selector (v4)

semisse:


semisse:
Investigation is done and we will move forward with vaadin-date-picker v23

PR #1454, opened 2022-06-02
web ui development workshop exercise 1

swarnadipa-dev:


elements/document/collection/nuxeo-collection-create-layout.html
```diff
@@ -34,6 +34,9 @@
     >
     </nuxeo-input>
 
+    <nuxeo-input role="widget" label="[[i18n('source')]]" name="source" value="{{document.properties.dc:source}}">
```
Gabez0r, line 37: `source` is the key to the label to be translated. We are missing the respective key/value pair in the reference translation file.

elements/document/collection/nuxeo-collection-metadata-layout.html
```diff
@@ -30,6 +30,11 @@
       <div name="title">[[document.properties.dc:title]]</div>
     </div>
 
+    <div role="widget">
+      <label>[[i18n('source')]]</label>
+      <div name="source">[[document.properties.dc:source]]</div>
+    </div>
```
Gabez0r: What happens if `document.properties.dc:source` is empty? We will still see the field label with an empty value, right? But this won't fully meet the following AC:

> As a user, I want to be able to see dc:source on the metadata layout of a document of types Collection, Note, File, Folder and Workspace, **but only if that field has content**.

..., **but only if that field has content**

ftest/features/create_doc.feature
```diff
@@ -1,3 +1,4 @@
+
```
Gabez0r, line 1: Adding an unneeded empty line at the top of the file.

Gabez0r (commented):
Good job with the functional tests in [ftest/features/create_doc.feature](https://github.com/nuxeo/nuxeo-web-ui/pull/1454/commits/88447b0a3664c27e818a39e6116eb70b820fe3bb#diff-2277caa8a2f4641ffc95643eab86db5cf8983e879e2b5a13e9a8624ade34c56e). However, this only covers document creation. What about editing documents? Perhaps there's another feature file that needs changing?

elements/nuxeo-clipboard/nuxeo-clipboard.js
```diff
@@ -261,9 +261,7 @@ Polymer({
     if (!documents || documents.length === 0 || !this.hasFacet(doc, 'Folderish')) {
       return false;
     }
-    return doc.contextParameters && doc.contextParameters.subtypes
-      ? documents.every((entry) => doc.contextParameters.subtypes.map((e) => e.type).indexOf(entry.type) > -1)
-      : true;
+    return true;
```
Gabez0r, line 264: Well done. But this is causing the functional testes to fail now.

Gabez0r (commented):
Good job. However, I'm seeing on the commit log commits that are already on main branch and that are not needed to be there (from Ricardo, for example), as well as some duplicates commits (72e46ba25f223103de736267305c7886a72a371b duplicates 88447b0a3664c27e818a39e6116eb70b820fe3bb and e4ce74283dca175cc0d77caf079697127d497ae2 duplicates 8ebd07276b8be0a188001ef74da0b51fec793c67).

ftest/features/create_doc.feature
Gabez0r, line 1: Please don't forget to remove this line.

test/nuxeo-breadcrumb.test.js
```diff
@@ -99,7 +99,24 @@ suite('nuxeo-breadcrumb', () => {
...
+        path: '/default-domain/workspaces/my workspace/folder 1/folder 2/folder 3/my file',
+        title: 'my file',
+        type: 'File',
+        uid: '7ertyyy-hdjkdks-874fghd',
+        isTrashed: true,
+      }; */
+      document.isTrashed = true;
+      await flush();
+      // sinon.stub(breadcrumb, '_isTrashed').returns(true);
+      expect(isElementVisible(trashIcon)).to.be.true;
```
Gabez0r, line 172: Applying the following patch fixed the test for me:
```diff
diff --git a/test/nuxeo-breadcrumb.test.js b/test/nuxeo-breadcrumb.test.js
index cea4f95b..2e3a4b14 100644
--- a/test/nuxeo-breadcrumb.test.js
+++ b/test/nuxeo-breadcrumb.test.js
@@ -73,6 +73,63 @@ const document = {
   uid: '7ertyyy-hdjkdks-874fghd',
 };
 
+const trashedDocument = {
+  'entity-type': 'document',
+  contextParameters: {
+    breadcrumb: {
+      entries: [
+        {
+          path: '/default-domain',
+          title: 'Domain',
+          type: 'Domain',
+          uid: '1',
+        },
+        {
+          path: '/default-domain/workspaces',
+          title: 'Workspaces',
+          type: 'WorkspaceRoot',
+          uid: '2',
+        },
+        {
+          path: '/default-domain/workspaces/my workspace',
+          title: 'my workspace',
+          type: 'Workspace',
+          uid: '3',
+        },
+        {
+          path: '/default-domain/workspaces/my workspace/folder 1',
+          title: 'folder 1',
+          type: 'Folder',
+          uid: '4',
+        },
+        {
+          path: '/default-domain/workspaces/my workspace/folder 1/folder 2',
+          title: 'folder 2',
+          type: 'Folder',
+          uid: '5',
+        },
+        {
+          path: '/default-domain/workspaces/my workspace/folder 1/folder 2/folder 3',
+          title: 'folder 3',
+          type: 'Folder',
+          uid: '6',
+        },
+        {
+          path: '/default-domain/workspaces/my workspace/folder 1/folder 2/folder 3/my file',
+          title: 'my file',
+          type: 'File',
+          uid: '7',
+        },
+      ],
+    },
+  },
+  path: '/default-domain/workspaces/my workspace/folder 1/folder 2/folder 3/my file',
+  title: 'my file',
+  type: 'File',
+  uid: '7ertyyy-hdjkdks-874fghd',
+  isTrashed: true,
+};
+
 // Mock router
 const router = {
   browse: (path) => path.substring(1),
@@ -113,7 +170,7 @@ suite('nuxeo-breadcrumb', () => {
         uid: '7ertyyy-hdjkdks-874fghd',
         isTrashed: true,
       }; */
-      document.isTrashed = true;
+      breadcrumb.document = trashedDocument;
       await flush();
       // sinon.stub(breadcrumb, '_isTrashed').returns(true);
       expect(isElementVisible(trashIcon)).to.be.true;
```

Gabez0r (commented):
Looking great. But I still see some duplicated commits and by looking at the commit log, I can see that the branch was not rebased on the main branch, but instead the main branch was merged to the feature branch. We prefer rebasing to merging because it keeps the history clean and linear. Please see [Merging vs. Rebasing](https://www.atlassian.com/git/tutorials/merging-vs-rebasing) fore more information.

elements/nuxeo-browser/nuxeo-breadcrumb.js
```diff
@@ -129,9 +134,12 @@ import { microTask } from '@polymer/polymer/lib/utils/async.js';
               href$="[[urlFor(document)]]"
               class="current breadcrumb-item breadcrumb-item-current"
               aria-current="page"
-              title="[[_title(document)]]"
+              title="[[_title(document)]] [[_documentUID(document)]]"
```
richardsd: FYI, my comment on https://github.com/nuxeo/nuxeo-web-ui/pull/1473#discussion_r898089101

Gabez0r:
I also see some commits in triplicate (e.g. ".web ui development workshop exercise 1")

elements/nuxeo-browser/nuxeo-breadcrumb.js
```diff
@@ -129,9 +140,12 @@ import { microTask } from '@polymer/polymer/lib/utils/async.js';
...
```
Gabez0r: Where is this `_documentUID` coming from?

test/nuxeo-breadcrumb.test.js
```diff
@@ -99,7 +99,70 @@ suite('nuxeo-breadcrumb', () => {
     test('Should display the document title when a document is set and has a title', async () => {
       const title = breadcrumb.shadowRoot.querySelector('.current');
       expect(isElementVisible(title)).to.be.true;
-      expect(title.innerText).to.equal('my file');
+      expect(title.innerText).to.equal('my file (874fghd) ');
```
Gabez0r: Why do we need this extra space?
```suggestion
      expect(title.innerText).to.equal('my file (874fghd)');
```

Gabez0r (commented):
Well done. But now, the functional tests are failing, likely because of something introduced by the breadcrumb changes.

elements/nuxeo-browser/nuxeo-breadcrumb.js
```diff
@@ -151,6 +165,10 @@ import { microTask } from '@polymer/polymer/lib/utils/async.js';
           type: Object,
           observer: '_setBreadcrumbElements',
         },
+        icon: {
+          type: String,
+          value: 'nuxeo:delete',
+        },
```
Gabez0r: There's no requirement to expose this as a public property. Every property we expose is public API then we will have to maintain in the future. We can perfectly inline the icon name in the element.

elements/nuxeo-browser/nuxeo-breadcrumb.js
```diff
@@ -125,14 +149,28 @@ import { microTask } from '@polymer/polymer/lib/utils/async.js';
...
+                    >
+                      [[_title(document)]]
+                    </a>
+                  
+                        <span class="doc-uid">
+                        [[_documentUID(document)]]
+                      </span>
+                </div>
+                <template is="dom-if" if="[[_isTrashed(document)]]">
+                  <span class="trash-icon-parent">
```
Gabez0r, line 168: Why do we need to have a span here? Couldn't we just use the paper-icon-button directly?

elements/nuxeo-browser/nuxeo-breadcrumb.js
```diff
@@ -247,7 +285,19 @@ import { microTask } from '@polymer/polymer/lib/utils/async.js';
 
     _title(document) {
       if (document) {
-        return document.type === 'Root' ? this.i18n('browse.root') : document.title;
+        return document.type === 'Root' ? this.i18n('browse.root') : `${document.title}`;
```
Gabez0r: What are we changing here?

elements/nuxeo-browser/nuxeo-breadcrumb.js
```diff
@@ -247,7 +285,19 @@ import { microTask } from '@polymer/polymer/lib/utils/async.js';
...
+
+    _documentUID(document) {
+      if (document) {
+        return `(${document.uid.substring(document.uid.lastIndexOf('-') + 1, document.uid.length)})`.trim();
+      }
+    }
+
+    _isTrashed(document) {
+      if (document) {
+        return document.isTrashed;
```
Gabez0r: nitpicking: could be made smaller:
```suggestion
      return document && document.isTrashed;
```

test/nuxeo-breadcrumb.test.js
```diff
@@ -102,6 +102,76 @@ suite('nuxeo-breadcrumb', () => {
       expect(title.innerText).to.equal('my file');
     });
 
+    test('Should display the document uid when a document is set and has a title and has a UID', async () => {
+      const uid = breadcrumb.shadowRoot.querySelector('.doc-uid');
+      expect(isElementVisible(uid)).to.be.true;
+      expect(uid.innerText).to.equal('(874fghd)');
+    });
+
+    test('Should display trash icon beside the document title when a document is trashed', async () => {
```
Gabez0r, line 111: We should also test that the trash icon is not shown for non trashed documents.

Gabez0r (commented):
Looking good. For exercise 4, it seems that the trash icon doesn't have the correct size, which should 16x16.

elements/nuxeo-browser/nuxeo-breadcrumb.js
```diff
@@ -125,14 +149,28 @@ import { microTask } from '@polymer/polymer/lib/utils/async.js';
...
+                      class="current breadcrumb-item breadcrumb-item-current"
+                      aria-current="page"
+                      title="[[_title(document)]]"
+                    >
+                      [[_title(document)]]
+                    </a>
+                  
+                        <span class="doc-uid">
+                        [[_documentUID(document)]]
+                      </span>
```
Gabez0r: The indentation seems to be off here.

elements/nuxeo-browser/nuxeo-breadcrumb.js
swarnadipa-dev: the extra span has been added and been given a height, width and display style to align the icon on the same line as the title and uid

swarnadipa-dev: the icon size if made 16x16, it's not visible at all.

test/nuxeo-breadcrumb.test.js
```diff
@@ -102,6 +102,139 @@ suite('nuxeo-breadcrumb', () => {
...
+        isTrashed: true,
+      };
+      breadcrumb.set('document', trashDocument);
+      await flush();
+      const trashIcon = breadcrumb.shadowRoot.querySelector('.trash-icon-parent');
+
+      expect(isElementVisible(trashIcon)).to.be.true;
+    });
+
+    test('Should display trash icon beside the document title when a document is not trashed', async () => {
```
Gabez0r, line 175: I think we are missing a "not" here:
```suggestion
    test('Should not display trash icon beside the document title when a document is not trashed', async () => {
```

elements/nuxeo-browser/nuxeo-breadcrumb.js
Gabez0r: Please note that is being harder because we are using a `paper-icon-button`, which is designed to be interactive, whereas we just need an icon. We should use the [iron-icon](https://www.webcomponents.org/element/@polymer/iron-icon) instead, which already exposes [a couple of CSS variables to set width and height](https://github.com/PolymerElements/iron-icon/blob/master/iron-icon.js#L79-L80).

elements/nuxeo-browser/nuxeo-breadcrumb.js
```diff
@@ -125,14 +149,28 @@ import { microTask } from '@polymer/polymer/lib/utils/async.js';
...
+                      [[_title(document)]]
+                    </a>
+                  
+                    <span class="doc-uid">
+                      [[_documentUID(document)]]
+                    </span>
+                </div>
+                <template is="dom-if" if="[[_isTrashed(document)]]">
+                  <span class="trash-icon-parent">
+                    <paper-icon-button icon="nuxeo:delete" noink class="trash-icon"
```
Gabez0r: Please see https://github.com/nuxeo/nuxeo-web-ui/pull/1454/files#r907139132, ir order to get the correct size in the icon. You should also just be able to apply the usual width and height css properties direction on the iron-icon.

elements/nuxeo-document-actions/nuxeo-document-parent-inspector-button.js
```diff
@@ -0,0 +1,241 @@
...
+      notify: true,
+    },
+    icon: {
+      type: String,
+      value: 'icons:info',
+    },
+    layout: {
+      type: String,
+      value: 'parent-inspector',
+    },
```
Gabez0r: What's the goal of having `layout` as a property? Do we have any requirement to make it configurable?

elements/nuxeo-document-actions/nuxeo-document-parent-inspector-button.js
```diff
@@ -0,0 +1,241 @@
...
+    facets: {
```
Gabez0r: Since we have no requirement to make this public, we should make it private. We have not concrete wat to do this in vanilla js, but we can either turn this into a read only property or precede the name with an underscore:
```suggestion
    _facets: {
```

elements/nuxeo-document-actions/nuxeo-document-parent-inspector-button.js
```diff
@@ -0,0 +1,241 @@
...
+      type: Array,
+      value: [],
+    },
+    schemas: {
```
Gabez0r: Same here:
```suggestion
    _schemas: {
```

elements/nuxeo-document-actions/nuxeo-document-parent-inspector-button.js
```diff
@@ -0,0 +1,241 @@
...
+  },
+
+  _isAvailable() {
+    return true;
+  },
+
+  get dialog() {
+    this.facets = this.document.facets;
+    this.schemas = this.document.schemas;
+    return this.root.getElementById(`${this.layout}-dialog`);
```
Gabez0r: Why can't we just do?:
```suggestion
    return this.root.getElementById('parent-inspector-dialog-dialog');
```
Do we need to make it configurable by any reason?

Also, this should also work:
```suggestion
    return this.$['parent-inspector-dialog'];
```

elements/nuxeo-document-actions/nuxeo-document-parent-inspector-button.js
```diff
@@ -0,0 +1,241 @@
...
+                </div>
+              </div>
+            </template>
+          </div>
+          <div class="actions">
+            <paper-button
+              noink
+              class="secondary"
+              dialog-dismiss
+              on-tap="_closeDialog"
```
Gabez0r: We should no need this if we already have `dialog-dismiss`

elements/nuxeo-document-actions/nuxeo-document-parent-inspector-button.js
```diff
@@ -0,0 +1,241 @@
...
+              role="button"
+              tabindex="0"
+              animated=""
+              elevation="0"
+              aria-disabled="false"
```
Gabez0r: Do we need these?

elements/nuxeo-document-actions/nuxeo-document-parent-inspector-button.js
```diff
@@ -0,0 +1,241 @@
...
+  },
+
+  ready() {
+    this.$.nxcon.connect().then((user) => {
+      this.currentUser = user;
+    });
+  },
+
+  _isAvailable() {
+    return true;
```
Gabez0r: This is optional: If we were to follow the last acceptance criteria, we would check if the document was trashed here or not.

elements/nuxeo-document-actions/nuxeo-document-parent-inspector-button.js
```diff
@@ -0,0 +1,241 @@
...
+    this.dialog.open();
+  },
+
+  _closeDialog() {
+    this.dialog.close();
+  },
+
+  _computeLabel() {
+    const x = this.i18n('parentInspectorButton.tooltip');
+    return x;
```
Gabez0r: We could do this in a single line:
```suggestion
    return this.i18n('parentInspectorButton.tooltip');
```

elements/nuxeo-document-actions/nuxeo-document-parent-inspector-button.js
```diff
@@ -0,0 +1,241 @@
...
+import '@nuxeo/nuxeo-ui-elements/widgets/nuxeo-tooltip.js';
+import '../document/nuxeo-document-form-layout.js';
+import '@polymer/paper-button/paper-button.js';
+
+/**
+`nuxeo-document-parent-inspector-button`
+@group Nuxeo UI
+@element nuxeo-document-parent-inspector-button
+*/
+Polymer({
```
Gabez0r: Please note that this is the old syntax, using the Polymer factory function (used in polymer v1). From polymer v2 up, the class-based syntax is preferred (see [docs](https://polymer-library.polymer-project.org/3.0/docs/devguide/registering-elements) and [upgrade guide](https://polymer-library.polymer-project.org/2.0/docs/upgrade#upgrading-to-class-based-elements)), although we still support it for retro-compatibility.

We recommend updating this code to use class based syntax instead. Most element on nuxeo-elements already follow this and several of Web UI too. Here's an example: https://github.com/nuxeo/nuxeo-web-ui/blob/v3.0.13/elements/nuxeo-audit/nuxeo-audit-search.js

elements/nuxeo-document-actions/nuxeo-document-parent-inspector-button.js
```diff
@@ -0,0 +1,241 @@
...
+        <div class="parentscroll">
+          <div class="scrollable">
+            <div>
+              <h1>Parent Inspector</h1>
+              <div>
+                <p>This is the parent inspector. It displays information about the parent of this document.</p>
+                <table class="table">
+                  <tr>
+                    <td>Tile:</td>
+                    <td>[[document.title]]</td>
```
Gabez0r: We should be displaying information for the **parent** of the current document, but we are instead displaying if for the current document.

Gabez0r (commented):
Looking good. However, we are displaying information for the current document, whereas we should be displaying information for the **parent** of the current document.

There are several ways to reach this information. We can do a separate REST call, which is the hard way, or leverage the [document enrichers](https://doc.nuxeo.com/nxdoc/content-enrichers/) available in the platform. By default, Nuxeo Web UI already uses the [enricher](https://doc.nuxeo.com/nxdoc/content-enrichers/#first-accessible-ancestor) that fetches the parent document, so it is readily available (see [here](https://github.com/nuxeo/nuxeo-web-ui/blob/v3.0.13/elements/nuxeo-app.js#L751) and [here](https://github.com/nuxeo/nuxeo-web-ui/blob/v3.0.13/elements/nuxeo-app.js#L1517)).

elements/nuxeo-document-actions/nuxeo-document-parent-inspector-button.js
Gabez0r: > This is optional: If we were to follow the last acceptance criteria, we would check if the document was trashed here or not.

As a matter of fact, please disregard this comment, as we already do not show document actions for trashed documents by default.

Gabez0r: We should however check that the document is defined here.

elements/nuxeo-document-actions/nuxeo-document-parent-inspector-button.js
richardsd: No need to store the translation in `x` if we don't use it. ⬆️

elements/nuxeo-web-ui-bundle.html
```diff
@@ -231,6 +231,16 @@
 
 <!-- DOCUMENT ACTIONS -->
 
+<nuxeo-slot-content name="parentInspectorDocumentAction" slot="DOCUMENT_ACTIONS" order="10">
+  <template>
+    <nuxeo-filter user="[[user]]" expression="user.isAdministrator || this.isMember(user, 'powerusers')">
```
richardsd: As an alternative solution, please check https://github.com/nuxeo/nuxeo-elements/blob/maintenance-3.0.x/ui/nuxeo-filter.js#L167. `nuxeo-filter` has a `group` property that can be use to check user membership to groups.

i18n/messages-ar.json
```diff
@@ -983,6 +983,7 @@
   "label.dublincore.state": "State",
   "label.dublincore.subjects": "Subjects",
   "label.dublincore.title": "Title",
+  "label.dublincore.source": "Source",
```
richardsd, line 986: Did you manually add the `label.dublincore.source` label to all the `i18n/messages-*.json` files? Please notice that this is not required. 
When adding a new label, you should just add it to the `i18n/messages.json` file. The translation process will happen on Crowdin (please refer to the wiki page about translations for details about this process).

richardsd (commented):
Looking good 👍 🙂

Gabez0r (commented):
Please don't forget to update your element to class-based syntax. See more comments here: Good job. Please see comments here: https://github.com/nuxeo/nuxeo-web-ui/pull/1491#pullrequestreview-1027608150.

Gabez0r (commented):
Well done! Please see comments here: https://github.com/nuxeo/nuxeo-web-ui/pull/1491#pullrequestreview-1028279090

test/nuxeo-parent-inspector-button.test.js
```diff
@@ -0,0 +1,150 @@
...
+      };
+      actionButton.set('document', trashDocument);
+      await flush();
+      const actionBtn = button.$$('.action');
+      expect(isElementVisible(actionBtn)).to.be.false;
+    });
+
+    test('Should display title, UID, path, schemas and facets if user is administrator', async () => {
+      button = await buildButton();
+      button.document = document;
```
Gabez0r, line 122: I don't think this is does anything, as this is already done in when we instantiate actionButton.

test/nuxeo-parent-inspector-button.test.js
```diff
@@ -0,0 +1,150 @@
...
+          },
+        },
+      };
+      actionButton.set('document', trashDocument);
+      await flush();
+      const actionBtn = button.$$('.action');
+      expect(isElementVisible(actionBtn)).to.be.false;
+    });
+
+    test('Should display title, UID, path, schemas and facets if user is administrator', async () => {
```
Gabez0r, line 120: We also need to check what happens if the user is not administrator.

Gabez0r (commented):
Good job.

PR #1471, opened 2022-06-07
UI dev workshop diksha

Diksha-Verma-Dev:
Workshop exercise 1


ftest/features/create_doc.feature
```diff
@@ -8,13 +8,14 @@ Feature: Create Document
     And I have a Workspace document
     And I have permission ReadWrite for this document
     And I browse to the document
-
+  @watch
```
Gabez0r: We should not be committing this.

elements/document/collection/nuxeo-collection-create-layout.html
```diff
@@ -34,6 +34,15 @@
     >
     </nuxeo-input>
 
+    <nuxeo-input
+      role="widget"
+      label="[[i18n('source')]]"
```
Gabez0r, line 39: `source` is the key to the label to be translated. We are missing the respective key/value pair in the reference translation file.

elements/document/collection/nuxeo-collection-metadata-layout.html
```diff
@@ -30,6 +30,11 @@
       <div name="title">[[document.properties.dc:title]]</div>
     </div>
 
+    <div role="widget">
+      <label>[[i18n('source')]]</label>
+      <div name="source">[[document.properties.dc:source]]</div>
+    </div>
```
Gabez0r: What happens if `document.properties.dc:source` is empty? We will still see the field label with an empty value, right? But this won't fully meet the following AC:

> As a user, I want to be able to see dc:source on the metadata layout of a document of types Collection, Note, File, Folder and Workspace, **but only if that field has content**.

..., **but only if that field has content**

The same applies to the other metadata layouts.

Gabez0r (commented):
Good job with the functional tests. However, this only covers document creation. What about editing documents? Are there a y another feature file that needs to update?

i18n/messages-eu.json
```diff
@@ -1,32 +1,32 @@
 {
-  "activity.addedToCollection": "bildumara gehituta",
+  "activity.addedToCollection": "bildumara gehitua",
```
Gabez0r, line 2: Where are these changes coming from?

packages/nuxeo-web-ui-ftest/pages/ui/browser.js
```diff
@@ -180,6 +180,7 @@ export default class Browser extends BasePage {
 
   addToFavorites() {
     this.el.click('nuxeo-favorites-toggle-button');
+    driver.alertAccept();
```
Gabez0r, line 183: This won't work without the changes on the nuxeo-elements side.

Gabez0r (commented):
Looking good. However, I'm seeing on the commit log commits that are already on main branch and that are not needed to be there (from Ricardo, for example), as well as some duplicates commits (ebf461ffe822bdf0bfeca8d6490eb21eb5995113 duplicates 25452efe14b5f345bdad7f85cd25231d111f51e0, and 6b2dc55ec0b5fb7a544b35ec1b3fd2aeae1f1404 duplicates c85b7acce93c251e7f75346bb27da19bcd2e4374).

I'm also seeing some of your commits authored with your username and some with "global" which I'm not sure who it is?

PR #1478, opened 2022-06-08
UI dev workshop rakesh

rakeshkumar1019:


ftest/features/create_doc.feature
```diff
@@ -1,3 +1,4 @@
+@watch
```
Gabez0r: We should not push this unless necessary, as this is only used for development purposes.

Gabez0r (commented):
Well done. We have the funcional testes for creating documents, but we are still missing the funcional tests for editing documents.

ftest/features/create_doc.feature
rakeshkumar1019: thankyou for the comment.

PR #1479, opened 2022-06-08
WEBUI-814: delay bulk edit widget injection until layout is loaded

richardsd:


elements/bulk/nuxeo-edit-documents-button.js
```diff
@@ -486,53 +486,56 @@ class NuxeoEditDocumentsButton extends mixinBehaviors([I18nBehavior, FiltersBeha
    */
   _elementChanged() {
     const layout = this.$$('nuxeo-layout');
-    if (!layout || !layout.element || !layout.element.shadowRoot) {
```
richardsd, line 489: The problem lies in this condition. `_elementChanged` is triggered when the `element` is updated [here](https://github.com/nuxeo/nuxeo-elements/blob/maintenance-3.0.x/ui/nuxeo-layout.js#L167) and [here](https://github.com/nuxeo/nuxeo-elements/blob/maintenance-3.0.x/ui/nuxeo-layout.js#L187). However, we might get here before the actual DOM of the layout is loaded. Waiting for the layout to be loaded, seems the best approach to me.

semisse (approved):
Nice solution 👏

Gabez0r (approved):

PR #1497, opened 2022-06-20
WEBUI-822: update nuxeo js client

richardsd:
It seems that when implementing ELEMENTS-709, the `nuxeo-js-client` version that was locked on 10.10 release `NXP-26541` was not updated, and only on version `v3.13.0`the AbortSignal object is used on the fetch API.

We are going to update the `nuxeo-js-client`, but hide the `uncancelable` behind a feature flag.
Check https://github.com/nuxeo/nuxeo-elements/pull/548 for details on Elements side.


src/main/resources/OSGI-INF/web-ui-properties.xml
```diff
@@ -45,5 +45,8 @@
 
     <!-- Properties to be fetched when loading the user object in user management, default is empty -->
     <property name="org.nuxeo.web.ui.user.management.fetch.document" list="true"></property>
+
+    <!-- Control if requests for requests are cancelable or not -->
+    <property name="org.nuxeo.web.ui.resource.uncancelable">true</property>
```
richardsd: @Gabez0r should we be able to add something like the following contribution to override the property:

```
<?xml version="1.0"?>
<component name="org.nuxeo.web.ui.dev.uncancelable">
  <require>org.nuxeo.web.ui.properties.contrib</require>
    <property name="org.nuxeo.web.ui.resource.uncancelable">false</property>
  </extension>
</component>
```
or should we expose this to a property that can be contributed in `nuxeo.conf`? 🤔

Gabez0r: I believe allowing it to be set from nuxeo.conf would make this easier to use, so unless we have a good reason not to do it, I think it would be a good option.

richardsd: I think the same way, but I just wanted your opinion too 🙂

index.html
```diff
@@ -103,6 +103,11 @@
       }
     };
     */
+    /*
+    Nuxeo.UI.config.resource = {
+      uncancelable: true,
+    };
+    */
```
Gabez0r: By default, shouldn't we actually have this uncommented? Otherwise, when serving with `index.html` we will `uncancelable` set to `false`, which is different when what we will get wit the `index.jsp`, right?
```suggestion
    Nuxeo.UI.config.resource = {
      uncancelable: true,
    };
```

Gabez0r (commented):
Shouldn't we also do the same for `nuxeo-operation`?

index.html
richardsd: You are totally right, let me update it.

Gabez0r (dismissed):

semisse (approved):

Gabez0r (approved):

PR #1506, opened 2022-06-24
WEBUI-826: stabilize document tree unit tests

richardsd:


test/nuxeo-document-tree.test.js
```diff
@@ -166,7 +166,7 @@ suite('nuxeo-document-tree', () => {
     // set the breadcrumb for some documents
     levelTwoDocument.contextParameters.breadcrumb = {
       'entity-type': 'documents',
-      entries: [levelOneDocument],
+      entries: [levelOneDocument, JSON.parse(JSON.stringify(levelTwoDocument))],
```
Gabez0r, line 169: nit: ideally we would use `structuredClone`, but it is only supported from Safari 15.4 on, so I'm not sure it will run on saucelabs?
```suggestion
      entries: [levelOneDocument, structuredClone(levelTwoDocument)],
```

Gabez0r (approved):
🤞

semisse (approved):

test/nuxeo-document-tree.test.js
richardsd: Saucelabs uses Safari 15.1 at the moment, so it is not a good idea to use the `structuredClone`.

PR #1511, opened 2022-07-01
WEBUI-837: bump hf version used in workflows

richardsd:


Gabez0r (approved):

Gabez0r:
@richardsd, do we have documented somewhere that this needs to be bumped manually? 🤔

richardsd:
> @richardsd, do we have documented somewhere that this needs to be bumped manually? 🤔

The use is documented in the internal Web UI wiki page, but I just updated it to indicate it should be manually bumped.

PR #1513, opened 2022-07-11
WEBUI-786: Able to click 'Add' Button when no Collection selected

Diksha-Verma-Dev:
https://jira.nuxeo.com/browse/WEBUI-786

The Add button is disabled since no collection is selected

Screenshot 2022-07-11 at 5 00 50 PM

### DIFF
--- a/elements/nuxeo-document-bulk-actions/nuxeo-add-to-collection-documents-button.js
+++ b/elements/nuxeo-document-bulk-actions/nuxeo-add-to-collection-documents-button.js
@@ -231,7 +231,7 @@ class NuxeoAddToCollectionDocumentsButton extends mixinBehaviors(
   }
 
   _isValid() {
-    return this.collection !== '';
+    return this.collection;
   }
 
   _isNew() {


semisse:
Looking good, Diksha! 

Don't forget to sign our Contributor License Agreement:

> CLA assistant check
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/nuxeo/nuxeo-web-ui?pullRequest=1513) before we can accept your contribution.
> You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/nuxeo/nuxeo-web-ui?pullRequest=1513) it.

richardsd (approved):

elements/nuxeo-document-bulk-actions/nuxeo-add-to-collection-documents-button.js
```diff
@@ -231,7 +231,7 @@ class NuxeoAddToCollectionDocumentsButton extends mixinBehaviors(
   }
 
   _isValid() {
-    return this.collection !== '';
+    return this.collection;
```
Gabez0r, line 234: But why is it that this check works the first time, but not the second time around? This tells us that we have an inconsistency in the code: the starting value for `this.collection` is `''`, but when we reset the popup, we set it to `null`. I understand that this inconsistency was added with https://github.com/nuxeo/nuxeo-web-ui/commit/ed9a2b9c866da18016f4d55a80cc1874a2b4a756, but if we could fix it, that would also fix the issue.

Gabez0r (commented):
Looking good, Diksha. Also, please do not forget to backport this feature for LTS 2019. When doing so, please link boths PRs (A simple message with the link to this one will suffice).

elements/nuxeo-document-bulk-actions/nuxeo-add-to-collection-documents-button.js
Gabez0r: If we need to keep setting the collection to `null` when we reset the popup, then I'm ok with the solution as it is now.

Diksha-Verma-Dev: Hi Gabriel, I did some workaround in removing the data from collections but it caused some miss in other area like when I add something in collection but I cancel that operation then when I again click on add to collection then it was showing the add to collection button enabled and some other bugs so I feel that this code will not cause any side-effect or bug.

Gabez0r: @Diksha-Verma-Dev, this is not about the current code causing side effects. This is about code consistency, so that we don't have two different values for when we have no collection (`''` vs `null`), if we don't have to. Something along these lines:
```diff
diff --git a/elements/nuxeo-document-bulk-actions/nuxeo-add-to-collection-documents-button.js b/elements/nuxeo-document-bulk-actions/nuxeo-add-to-collection-documents-button.js
index 85fa2efc..b90dca2a 100644
--- a/elements/nuxeo-document-bulk-actions/nuxeo-add-to-collection-documents-button.js
+++ b/elements/nuxeo-document-bulk-actions/nuxeo-add-to-collection-documents-button.js
@@ -239,7 +239,7 @@ class NuxeoAddToCollectionDocumentsButton extends mixinBehaviors(
   }
 
   _resetPopup() {
-    this.set('collection', null);
+    this.set('collection', '');
     this.description = '';
   }
 }

```
Also, please note that even if we should fix this on `nuxeo-add-to-collection-documents-button.js`, which is used when performing selection, the reported bug is for the document action [nuxeo-add-to-collection-button.js](https://github.com/nuxeo/nuxeo-elements/blob/maintenance-3.0.x/ui/actions/nuxeo-add-to-collection-button.js)

Diksha-Verma-Dev: Hi Gabriel, here are my finding in the attached videos below kindly share your inputs for the same

https://user-images.githubusercontent.com/105918630/178761480-d10f1c17-0cd5-460d-8989-87715df37976.mov



https://user-images.githubusercontent.com/105918630/178761510-043d4deb-d001-4885-ac08-5fbcec536c2f.mov

https://user-images.githubusercontent.com/105918630/178761607-4f19c5b5-69e3-4642-bb2d-8fa811631ffa.mov

Gabez0r: Thank you so much, @Diksha-Verma-Dev . It seems that the small inconsistency we have there, is there for good reason! I'm ok with your solution. Please don't forget to fix this on [nuxeo-add-to-collection-button.js](https://github.com/nuxeo/nuxeo-elements/blob/maintenance-3.0.x/ui/actions/nuxeo-add-to-collection-button.js) too.

Gabez0r (approved):

PR #1517, opened 2022-07-14
WEBUI-397: share saved search popup issue resolved

SnehalNaik-DEV:


elements/search/nuxeo-search-form.js
```diff
@@ -403,14 +431,24 @@ Polymer({
...
 
-    <nuxeo-dialog id="shareDialog" with-backdrop reparent opened="{{permissionsVisible}}">
-      <h2>[[i18n('searchForm.shared.heading')]]</h2>
-      <nuxeo-document-permissions
-        doc-id="[[selectedSearch.id]]"
-        visible="[[permissionsVisible]]"
-      ></nuxeo-document-permissions>
-      <div class="buttons">
-        <paper-button dialog-dismiss class="secondary">[[i18n('command.close')]]</paper-button>
+    <nuxeo-dialog id="shareDialog" with-backdrop reparent opened="{{permissionsVisible}}" class="sharedialog">
```
richardsd: @SnehalNaik-DEV although this solution is kind of working, it seems a bit over-complicated. 
We could take this approach and use all this CSS ⬆️ to add scrolling to the panel, but I think we could use a similar solution to what we use in other places (see https://github.com/nuxeo/nuxeo-web-ui/blob/maintenance-3.0.x/addons/nuxeo-template-rendering/elements/nuxeo-delete-template-button.js#L49-L53). Using the `paper-dialog-scrollable` you can wrap up the part you want to be scrollable, without having to mess around with the CSS like this. What do you think?

A good practice is to always look for similar use cases in the application 😉 
paper-dialog-scrollable

Gabez0r (commented):
Well done. But now we have two commits, with two different solutions, but we only need one:
image
Please squash the commits.

rakeshkumar1019 (commented):
looks good to me.

Gabez0r (approved):
Good job!

rakeshkumar1019 (commented):
looks good

rakeshkumar1019 (approved):

PR #1530, opened 2022-07-21
WEBUI-842: wait for shadow root to be attached on bulk edit

Gabez0r:
I was only able to reproduce this locally using the itests contributions. It seems to be that, only in Safari, there are some conditions in which the bulk layout does not yet have a shadow root attached, but the properties observer is already being triggered, which causes a failure when it tries to do a querySelectorAll on the shadow root. The proposed solution waits for the shadow root to be defined for a given period of time (3 seconds, at the moment).


semisse (approved):

Gabez0r:
A possible alternative could consist of using a MutationObserver, but I don't think it detects when the shadow root is attached to element. See https://github.com/w3c/dom/issues/119.

rakeshkumar1019 (approved):

swarnadipa-dev (approved):

SnehalNaik-DEV (approved):
Looking good !

PR #1542, opened 2022-07-29
WEBUI-794: display a label instead of doc id for the Root document

GuruprasadAnk1996:
https://jira.nuxeo.com/browse/WEBUI-794


rakeshkumar1019 (approved):
looks good

PR #1555, opened 2022-07-29
WEBUI-794: display a label instead of doc id for the Root document

GuruprasadAnk1996:
https://jira.nuxeo.com/browse/WEBUI-794

### DIFF
--- a/elements/nuxeo-clipboard/nuxeo-clipboard.js
+++ b/elements/nuxeo-clipboard/nuxeo-clipboard.js
@@ -156,7 +156,11 @@ Polymer({
                 <div class="list-item-thumbnail">
                   <nuxeo-document-thumbnail document="[[document]]"></nuxeo-document-thumbnail>
                 </div>
-                <div class="list-item-title">[[document.title]]</div>
+                <div class="list-item-title">
+                  <label>
+                    [[_title(document)]]
+                  </label>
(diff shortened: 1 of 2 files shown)


elements/nuxeo-clipboard/nuxeo-clipboard.js
```diff
@@ -156,7 +156,11 @@ Polymer({
                 <div class="list-item-thumbnail">
                   <nuxeo-document-thumbnail document="[[document]]"></nuxeo-document-thumbnail>
                 </div>
-                <div class="list-item-title">[[document.title]]</div>
+                <div class="list-item-title">
+                  <label>
```
semisse, line 160: Why are we wrapping it with a `label`?

GuruprasadAnk1996: As a practice to wrap titles in label, should i remove it?

semisse: It’s not consistent with our code base and might introduce undesirable styles. Since the previous code didn’t have we should not introduce it without a good reason

GuruprasadAnk1996: Noted Sam.

PR #1568, opened 2022-08-19
Imp elements 1397 add unselect some feature ftest

Nisha-QA:


ftest/features/bulk_edit.feature
```diff
@@ -84,3 +84,29 @@ Feature: Bulk Edit
...
+    And I cannot see the display selection link
+    And I select all the documents
+    And I select all the documents
+    And I deselect the "First_File" document
+    And I click the bulk edit button with "default" layout
+    And I can bulk edit multiple properties in "default" layout:
+      | name         | value                | action    |
+      | subjects     |  Cinema              | replace   |
+    And I see a toast notification with the following message "Bulk Edit documents completed successfully on 2 document(s)"
+    And I click the toast notification dismiss button
```
Gabez0r: nit: I think we could use another When and Then here, we have a lot of Ands.

packages/nuxeo-web-ui-ftest/features/step_definitions/browser.js
```diff
@@ -71,6 +71,10 @@ Then('I can see the selection toolbar', function() {
   this.ui.browser.selectionToolbar.waitForVisible();
 });
 
+Then('I cannot see the display selection link', function() {
+  this.ui.browser.waitForNotVisible('.selectionLink').should.be.true;
+});
```
Gabez0r: Since the selection link is inside the selection toolbar, it would be more consistent (and efficient) to narrow down the search to the selection toolbar. Something like this:
```suggestion
Then('I cannot see the display selection link', function() {
  this.ui.browser.selectionToolbar.waitForNotVisible('.selectionLink').should.be.true;
});
```

Gabez0r (commented):
Looking good!

swarnadipa-dev (approved):
Looks good

Gabez0r (commented):
Looks good. There are a bunch of commit messages, all telling the same, so we might as well just squash them into a single commit.
image

PR #1585, opened 2022-09-02
WEBUI-873: make nuxeo-user-tag accessible

GuruprasadAnk1996:
https://jira.nuxeo.com/browse/WEBUI-873

The changes has been made to resolve the accessibility check fails in https://jira.nuxeo.com/browse/ELEMENTS-1364


Gabez0r:
@GuruprasadAnk1996 , if this depends on https://github.com/nuxeo/nuxeo-elements/pull/596, then you should be using a branch with the same name, otherwise, the a11y checks will fail.

PR #1589, opened 2022-09-06
WEBUI-104: add functional tests for document import with properties

Pranit-Sotre:
https://jira.nuxeo.com/browse/WEBUI-104

### DIFF
--- a/elements/document/nuxeo-document-import.js
+++ b/elements/document/nuxeo-document-import.js
@@ -561,6 +561,7 @@ Polymer({
                         attr-for-selected="key"
                         label="[[i18n('documentImportForm.type.label')]]"
                         placeholder="[[i18n('documentImportForm.type.placeholder')]]"
+                        name="assetType"
                         error-message="[[i18n('documentImportForm.type.error')]]"
                         required
                       >
@@ -668,6 +669,7 @@ Polymer({
             <paper-button
               noink
               class="text"
+              name="applyAll"
               on-tap="_applyToAll"
               disabled$="[[_disableApplyToAll(_initializingDoc,_creating,canCreate,customizing,docIdx)]]"
             >
@@ -678,6 +680,7 @@ Polymer({
           <paper-button
             noink
             class="primary"
+            name="createWithProperties"
             on-tap="_importWithProperties"
             disabled$="[[!_canImportWithMetadata(_creating,_initializingDoc,canCreate,hasLocalFilesUploaded,hasRemoteFiles,localFiles.*,remoteFiles.*)]]"
             aria-label$="[[i18n('command.create')]]"
--- a/ftest/features/import_doc.feature
+++ b/ftest/features/import_doc.feature
@@ -37,3 +37,63 @@ Feature: Import and Create Documents
     Then I can see that a document of the type Video and title sample.mp4 is created
     And I can see the inline nuxeo-video-viewer previewer
     And I can see the video conversions panel
+
+  Scenario: Import a Document with properties
+    When I click the Create Document button
+    Then I go to the import tab
+    And I can see the import tab content
+    And I upload the sample.png on the tab content page
+    When I click on Add Properties button
+    Then I select "Picture" asset type
+    And I add the following properties:
+      | name         | value             |
+      | description  | my description    |
+      | nature       | Application       |
+      | subjects     | Gastronomy,Comics |
+      | expired      | December 28, 2022 |
+    When I click the Create button to complete the import
+    And I see a toast notification with the following message "Created Picture sample.png."
+    Then I can see that a document of the type Picture and title sample.png is created
+    And I can see Picture metadata with the following properties:
+      | name         | value                            |
+      | title        | sample.png                       |
+      | description  | my description                   |
+      | nature       | Application                      |
+      | subjects     | Daily life/Gastronomy,Art/Comics |           
+    And I can see the inline nuxeo-image-viewer previewer
+    And I can see the picture formats panel
+
+    Scenario: Import 6 documents with properties
+    When I click the Create Document button
+    Then I go to the import tab
+    And I can see the import tab content
+    And I upload the following files on the tab content page:
+      | value      |
+      | sample.png |
+      | sample.mp3 |
+      | sample.mp4 |
+      | sample.pdf |
+      | sample.txt |
+      | sample.odt |
+    When I click on Add Properties button
+    Then I select "Picture" asset type
+    And I add the following properties:
+      | name         | value             |
+      | description  | my description    |
+      | nature       | Application       |
+      | subjects     | Gastronomy,Comics |
+      | expired      | December 28, 2022 |
+    And I click on Apply to all
+    When I click the Create button to complete the import
+    And I see a toast notification with the following message "Created 6 documents."
+    Then I can see 6 documents
+    When I navigate to "sample.png" child
+    Then I can see that a document of the type Picture and title sample.png is created
+    And I can see Picture metadata with the following properties:
+      | name         | value                            |
+      | title        | sample.png                       |
+      | description  | my description                   |
+      | nature       | Application                      |
+      | subjects     | Daily life/Gastronomy,Art/Comics |     
+    And I can see the inline nuxeo-image-viewer previewer
+    And I can see the picture formats panel
--- a/packages/nuxeo-web-ui-ftest/features/step_definitions/import_dialog.js
+++ b/packages/nuxeo-web-ui-ftest/features/step_definitions/import_dialog.js
@@ -1,0 +1,27 @@
+import { Then, When } from '@cucumber/cucumber';
+
+When('I click on Add Properties button', function() {
+  this.ui.createDialog.addProperties.waitForVisible();
+  this.ui.createDialog.addProperties.click();
+});
+
+Then('I select {string} asset type', function(val) {
+  this.ui.createDialog.selectAnAssetType.waitForVisible();
+  this.ui.createDialog.selectAnAssetType.click();
+  this.ui.createDialog.selectAssetType(val);
+});
+
+Then('I click the Create button to complete the import', function() {
+  const importButton = this.ui.createDialog.importCreateButtonProperties;
+  importButton.waitForVisible();
+  importButton.waitForEnabled();
+  importButton.click();
+});
+
+Then('I add the following properties:', function(table) {
+  this.ui.createDialog.documentImportLayout.fillMultipleValues(table);
+});
+
+Then('I click on Apply to all', function() {
+  this.ui.createDialog.applyAll.click();
+});
--- a/packages/nuxeo-web-ui-ftest/pages/ui/create_dialog.js
+++ b/packages/nuxeo-web-ui-ftest/pages/ui/create_dialog.js
@@ -1,10 +1,15 @@
 import path from 'path';
 import BasePage from '../base';
 import DocumentCreate from './browser/document_create';
+import DocumentLayout from './browser/document_layout';
 
 export default class CreateDialog extends BasePage {
   get documentCreate() {
     return new DocumentCreate('nuxeo-document-create');
+  }
+
+  get documentImportLayout() {
+    return new DocumentLayout('nuxeo-document-import');
   }
 
   importTab(name) {
@@ -58,6 +63,10 @@ export default class CreateDialog extends BasePage {
     return this.el.element('div[name="upload"] paper-button[id="create"]');
   }
 
+  get importCreateButtonProperties() {
+    return this.el.element('paper-button[name="createWithProperties"]');
+  }
+
   get importCSVButton() {
     return this.el.element('div[name="upload"] paper-button.primary');
   }
@@ -84,5 +93,25 @@ export default class CreateDialog extends BasePage {
 
   get createButton() {
     return this.el.element('paper-button[id="create"]');
+  }
+
+  get addProperties() {
+    return this.el.element('paper-button[id="edit"]');
+  }
+
+  get selectAnAssetType() {
+    return this.el.element('nuxeo-select[name="assetType"]');
+  }
+
+  get applyAll() {
+    return this.el.element('paper-button[name="applyAll"]');
+  }
+
+  selectAssetType(val) {
+    driver.waitForVisible('nuxeo-select[id="docTypeDropdown"] paper-item');
+    const selectasset = this.el
+      .elements('nuxeo-select[id="docTypeDropdown"] paper-item[role="option"]')
+      .find((e) => e.getText() === val);
+    selectasset.click();
   }
 }


ftest/features/import_doc.feature
```diff
@@ -37,3 +37,49 @@ Feature: Import and Create Documents
...
+    Then I select an asset type
+    And I add the following properties:
+      | name         | value             |
+      | description  | my description    |
+      | nature       | Application       |
+      | subjects     | Gastronomy,Comics |
+      | expired      | December 28, 2022 |
+    When I click the Create button to complete the import
+    Then I can see that a document of the type Picture and title sample.png is created
+    And I can see the inline nuxeo-image-viewer previewer
```
Nisha-QA, line 63: Add a step to validate the added properties

ftest/features/import_doc.feature
```diff
@@ -37,3 +37,49 @@ Feature: Import and Create Documents
...
+      | description  | my description    |
+      | nature       | Application       |
+      | subjects     | Gastronomy,Comics |
+      | expired      | December 28, 2022 |
+    And I click on Apply to all
+    When I click the Create button to complete the import
+    Then I can see 6 documents
+    When I navigate to "sample.png" child
+    Then I can see that a document of the type Picture and title sample.png is created
+    And I can see the inline nuxeo-image-viewer previewer
```
Nisha-QA, line 98: Here also, add a step to validate added properties

ftest/features/import_doc.feature
```diff
@@ -37,3 +37,49 @@ Feature: Import and Create Documents
     Then I can see that a document of the type Video and title sample.mp4 is created
     And I can see the inline nuxeo-video-viewer previewer
     And I can see the video conversions panel
+
+  Scenario: Import a Document with adding properties
```
Nisha-QA: Nit: you can rephrase this to something like : Import a document with properties

Nisha-QA (commented):
Since we have many new functions here you can create a new import_dialog.js file and move them

rakeshkumar1019 (changes requested):
make the changes suggested by @Nisha-QA Also, make sure the functional test check passes.

ftest/features/import_doc.feature
Pranit-Sotre: Done changes.

ftest/features/import_doc.feature
```diff
@@ -37,3 +37,59 @@ Feature: Import and Create Documents
...
     And I can see the inline nuxeo-video-viewer previewer
     And I can see the video conversions panel
+
+  Scenario: Import a Document with properties
+    When I click the Create Document button
+    Then I go to the import tab
+    And I can see the import tab content
+    And I upload the sample.png on the tab content page
+    When I click on Add Properties button
+    Then I select an asset type
```
Gabez0r: A reader won't be able to tell what asset type was selected. This step should actually allow different asset types to be selected (e.g. `I select "Picture" asset type`)

packages/nuxeo-web-ui-ftest/pages/ui.js
```diff
@@ -189,4 +207,24 @@ export default class UI extends BasePage {
...
+
+  fillMultipleValues(table) {
+    table.rows().forEach((row) => {
+      const fieldName = row[0];
+      const fieldEl = this.getField(fieldName);
+      fieldEl.waitForVisible();
+      fieldEl.scrollIntoView();
+      return fixtures.layouts.setValue(fieldEl, row[1]);
+    });
+  }
```
Gabez0r: Here we are duplicating a bunch of code which is already available in the [document_layout](https://github.com/nuxeo/nuxeo-web-ui/blob/maintenance-3.0.x/packages/nuxeo-web-ui-ftest/pages/ui/browser/document_layout.js#L4-L28) page helper. We should instead have an helper inside the createDialog so that we can retried the (import) document layout and thus reuse the page wrapper methods we already have.

packages/nuxeo-web-ui-ftest/pages/ui.js
```diff
@@ -168,6 +180,12 @@ export default class UI extends BasePage {
     dropdown.click(`#dropdown #contentWrapper div paper-menu div paper-icon-item[name="${selection}"]`);
   }
 
+  asset() {
+    const dropdown = driver.element('nuxeo-select[id="docTypeDropdown"]');
+    dropdown.waitForVisible('paper-item[role="option"]');
+    dropdown.click('paper-item:nth-child(4)');
+  }
+
```
Gabez0r: This returns an arbitrary item (the 4th item), which makes the method obscure an unpredictable. In case we deploy a new document type, this will might return different results. Also, "asset" as a name does not help much in the context of the ui app. Such an helper would be better placed inside the createDialog.

packages/nuxeo-web-ui-ftest/pages/ui.js
```diff
@@ -79,6 +79,18 @@ export default class UI extends BasePage {
...
+  get selectAnAssetType() {
+    return this.el.element('iron-input input[placeholder="Select an asset type."]');
+  }
+
+  get applyAll() {
+    return this.el.element('div[class="buttons horizontal justified layout"] paper-button.text:nth-child(3)');
+  }
+
```
Gabez0r: These should all be inside the createDialog (this.ui.createDialog), since we already have an abstraction for this.

packages/nuxeo-web-ui-ftest/features/step_definitions/import_dialog.js
```diff
@@ -0,0 +1,35 @@
...
+  this.ui.asset();
+});
+
+Then('I click the Create button to complete the import', function() {
+  const importButton = this.ui.createDialog.importCreateButtonProperties;
+  importButton.waitForVisible();
+  importButton.waitForEnabled();
+  importButton.click();
+  driver.waitForExist('iron-overlay-backdrop', driver.options.waitForTimeout, true);
+  driver.pause(3000); // XXX just give it some time to the server to do the conversions
```
Gabez0r: This will render the test flaky. If you need to wait for longer, you can always wait for the toast message.

packages/nuxeo-web-ui-ftest/features/step_definitions/import_dialog.js
```diff
@@ -0,0 +1,35 @@
...
+});
+
+Then('I can see the following properties:', function(table) {
```
Gabez0r: Why do we need to create a new step and don't reuse the other step we already use to check properties after creating documents? https://github.com/nuxeo/nuxeo-web-ui/blob/maintenance-3.0.x/ftest/features/create_doc.feature#L51

packages/nuxeo-web-ui-ftest/features/step_definitions/import_dialog.js
```diff
@@ -0,0 +1,35 @@
...
+  });
+});
+
+Then('I add the following properties:', function(table) {
+  this.ui.fillMultipleValues(table);
+});
+
+Then('I click on Apply to all', function() {
+  this.ui.applyAll.click();
+});
```
Gabez0r: This should live inside the createDialog (this.ui.createDialog).

Gabez0r (changes requested):
It's going in the right direction, good job!

packages/nuxeo-web-ui-ftest/pages/ui.js
Pranit-Sotre: We are not using the existing function, as xpath is different for create and import dialog box.

packages/nuxeo-web-ui-ftest/features/step_definitions/import_dialog.js
Pranit-Sotre: We need a new step as web element location is different when we import a document.

richardsd:
@Pranit-Sotre don't forget to rebase your branch. I see the `ftests` pr check failing on the install step (it could help).

packages/nuxeo-web-ui-ftest/features/step_definitions/import_dialog.js
Gabez0r: This step validates the metadata after navigating to the document that was just created, so I would not expect the element location to be different.

packages/nuxeo-web-ui-ftest/pages/ui.js
Gabez0r: I'm not following. You should have a helper that returns the a [DocumentLayout](https://github.com/nuxeo/nuxeo-web-ui/blob/maintenance-3.0.x/packages/nuxeo-web-ui-ftest/pages/ui/browser/document_layout.js#L4-L28) wrapping your import layout, which should allow you to reuse the helpers.

ftest/features/import_doc.feature
```diff
@@ -37,3 +37,59 @@ Feature: Import and Create Documents
...
+    Then I can see 6 documents
+    When I navigate to "sample.png" child
+    Then I can see that a document of the type Picture and title sample.png is created
+    And I can see the following properties:
+      | name         | value             |
+      | title        | sample.png        |
+      | description  | my description    |
+      | nature       | Application       |
+    And I can see the inline nuxeo-image-viewer previewer
+    And I can see the picture formats panel
```
Gabez0r: nit: Missing newline at the end of the file

packages/nuxeo-web-ui-ftest/pages/ui/create_dialog.js
```diff
@@ -58,6 +58,10 @@ export default class CreateDialog extends BasePage {
     return this.el.element('div[name="upload"] paper-button[id="create"]');
   }
 
+  get importCreateButtonProperties() {
+    return this.el.element('div[class="buttons horizontal justified layout"] paper-button.primary');
```
Gabez0r: We should probably add a `name`, otherwise this will break if someone adjusts a class. for example.

packages/nuxeo-web-ui-ftest/pages/ui/create_dialog.js
```diff
@@ -85,4 +89,24 @@ export default class CreateDialog extends BasePage {
   get createButton() {
     return this.el.element('paper-button[id="create"]');
   }
+
+  get addProperties() {
+    return this.el.element('paper-button[id="edit"]');
+  }
+
+  get selectAnAssetType() {
+    return this.el.element('iron-input input[placeholder="Select an asset type."]');
```
Gabez0r: Same as above, adding a name or id here is preferable, since the placeholder name might change.

packages/nuxeo-web-ui-ftest/pages/ui/create_dialog.js
```diff
@@ -85,4 +89,24 @@ export default class CreateDialog extends BasePage {
...
+  }
+
+  get applyAll() {
+    return this.el.element('div[class="buttons horizontal justified layout"] paper-button.text:nth-child(3)');
```
Gabez0r: Same as above, adding a name here is preferable.

packages/nuxeo-web-ui-ftest/pages/ui/create_dialog.js
```diff
@@ -85,4 +89,24 @@ export default class CreateDialog extends BasePage {
...
+  }
+
+  asset(val) {
```
Gabez0r: This method should be named after what it does. For example, "selectAsset".

elements/document/nuxeo-document-import.js
```diff
@@ -678,6 +680,7 @@ Polymer({
           <paper-button
             noink
             class="primary"
+            name="create"
```
Gabez0r: There is already another button that does creation (with id="create"), so perhaps we could make this name more explicit to avoid any confusions. For example "createWithProperties" or something that tells that this is going to be used in the advanced scenario where we are adding properties.

packages/nuxeo-web-ui-ftest/pages/ui/create_dialog.js
```diff
@@ -1,12 +1,17 @@
...
 
 export default class CreateDialog extends BasePage {
   get documentCreate() {
     return new DocumentCreate('nuxeo-document-create');
   }
 
+  get documentLayout() {
+    return new DocumentLayout('nuxeo-document-import');
+  }
+
```
Gabez0r: This is confusing because from this element, we can reach both "create" and "import" layouts. We have two options here:
1) Looking at the lines immediately above, I believe we could further extract this into a `DocumentImport` wrapper, which then could return the layout just have as you have here. This would also be consistent to what we are doing with DocumentCreate. This would allow us to move all the helpers we have below to this new wrapper.
2) Simpler option, but probably providing a more disperse API, to rename this method to "documentImportLayout".

Gabez0r (commented):
Looking good. We are almost there!

Nisha-QA (dismissed):
Looks good

rajatgoel-gl (dismissed):

packages/nuxeo-web-ui-ftest/pages/ui/create_dialog.js
```diff
@@ -85,4 +94,24 @@ export default class CreateDialog extends BasePage {
...
+  get selectAnAssetType() {
+    return this.el.element('nuxeo-select[name="assetType"]');
+  }
+
+  get applyAll() {
+    return this.el.element('paper-button[name="applyAll"]');
+  }
+
+  selectAsset(val) {
```
Gabez0r: I believe `selectAssetType` would be more appropriate even.

packages/nuxeo-web-ui-ftest/features/step_definitions/import_dialog.js
```diff
@@ -0,0 +1,28 @@
...
+  this.ui.createDialog.selectAnAssetType.click();
+  this.ui.createDialog.selectAsset(val);
+});
+
+Then('I click the Create button to complete the import', function() {
+  const importButton = this.ui.createDialog.importCreateButtonProperties;
+  importButton.waitForVisible();
+  importButton.waitForEnabled();
+  importButton.click();
+  driver.waitForExist('#snackbarPanel mwc-snackbar[open] .mdc-snackbar__label', driver.options.waitForTimeout, true);
```
Gabez0r: This just waits for the toast to be dismissed, but we are not even sure what message it was displaying, so this step might become flaky in some situations. We should wait to see the proper message on the toast that is displayed once the creation is complete.

rakeshkumar1019 (commented):
Squash the last commit, it looks good.

packages/nuxeo-web-ui-ftest/features/step_definitions/import_dialog.js
richardsd: I agree with @Gabez0r. We could probably do something like the following to assert that the message is displayed:

```js
  const notificationMessage = this.ui.getToastMessage(message);
  const trimmedMessage = message.trim().replace(/"/g, '');
  notificationMessage.should.be.equals(trimmedMessage);
```
Check https://github.com/nuxeo/nuxeo-web-ui/blob/maintenance-3.0.x/packages/nuxeo-web-ui-ftest/features/step_definitions/bulk_edit.js#L14-L18 for the context.

richardsd (changes requested):
Please review your code based on @Gabez0r comment 🙂

packages/nuxeo-web-ui-ftest/features/step_definitions/import_dialog.js
Pranit-Sotre: I have covered this point in import_doc.feature at line number 55 and 88.

Gabez0r: Ok then this this line waiting for the  toast to disappear is redundant.

FabioFSantos (changes requested):
Please review your code based on @Gabez0r last comment about the toast check redundancy, the rest looks good

packages/nuxeo-web-ui-ftest/features/step_definitions/import_dialog.js
Pranit-Sotre: Done

richardsd (dismissed):

Nisha-QA (dismissed):
Looks good

FabioFSantos (dismissed):

rakeshkumar1019 (approved):
looks good 👍

richardsd (approved):

Gabez0r (approved):

FabioFSantos (approved):

pddgoncalves (approved):

Nisha-QA (approved):
Looks good

PR #1605, opened 2022-09-27
WEBUI-920: fix URL to Kawaii background

tmartins:


rakeshkumar1019 (changes requested):
It appears that functional test checks were failing due to a random error. It usually appears. Please re-run the checks. It will solve the issue.

PR #1606, opened 2022-09-27
WEBUI-920: fix URL to Kawaii background

tmartins:


rakeshkumar1019 (changes requested):
Please rebase with `maintenance-3.0.x` all checks will now pass.

PR #1607, opened 2022-09-28
WEBUI-424: fix full text input requiring Enter key press in Assets an…

GuruprasadAnk1996:
https://jira.nuxeo.com/browse/WEBUI-424

### DIFF
--- a/elements/search/assets/nuxeo-assets-search-form.html
+++ b/elements/search/assets/nuxeo-assets-search-form.html
@@ -33,7 +33,7 @@ limitations under the License.
       type="search"
       placeholder="[[i18n('assetsSearch.fullText.placeholder')]]"
       autofocus
-      on-keydown="_checkForEnter"
+      on-change="_search"
     >
     </nuxeo-input>
 
@@ -148,12 +148,6 @@ limitations under the License.
           // dirty but needed, the above line for
           // ecm_fulltext removal being observed
           delete this.params.ecm_fulltext;
-        }
-      },
-
-      _checkForEnter(e) {
-        if (e.keyCode === 13) {
-          this._search();
         }
       },
     });
--- a/elements/search/trash/nuxeo-trash-search-form.html
+++ b/elements/search/trash/nuxeo-trash-search-form.html
@@ -32,7 +32,7 @@ limitations under the License.
       value="{{searchTerm}}"
       placeholder="[[i18n('defaultSearch.fullText.placeholder')]]"
       autofocus
-      on-keydown="_checkForEnter"
+      on-change="_search"
     >
     </nuxeo-input>
 
@@ -115,12 +115,6 @@ limitations under the License.
       _addPathSearch() {
         if (this._path) {
           this.set('params.ecm_path', [this._path]);
-        }
-      },
-
-      _checkForEnter(e) {
-        if (e.keyCode === 13) {
-          this._search();
         }
       },
     });


swarnadipa-dev (dismissed):
Looks good

rajatgoel-gl (dismissed):

Gabez0r (commented):
The solution solves the problem reported in the ticket, but it changes the behavior of the search:
1) **Before**: in auto mode, you had to press enter in full text search for the search to be executed.
2) **After**: in auto mode, search is performed automatically.

For me, the new behavior makes sense, but it is not consistent with what we have for other searches, including the default search. Please validate the change with @bchauvin first (or let me know if you already did).

bchauvin:
@Gabez0r @GuruprasadAnk1996 IIRC we requested people to use the enter key in the fulltext of auto search because otherwise typing would send too many requests to the server and create performance issues.

So, we have two options from here:
1. Remove the need to press enter in auto search but keep performance (maybe by using some kind of debounce or something?). If we do that, it needs to be consistent everywhere and not just in assets search.
=> it would certainly be appreciated by our customers, as long as performance is not impacted
=> scope might be bigger than what is requested in the support ticket, so maybe we could consider that as an improvement

2. Play it safe: keep the existing behavior that you need to press enter in auto search
=> solves the support ticket, so I would say it is ok and most likely the safest option to choose here

Gabez0r:
It is my understanding that we already have debouncers in place that prevent too many request to be made, so I wouldn't expect this to degrade performance, but I would recommend double checking.

elements/search/assets/nuxeo-assets-search-form.html
```diff
@@ -152,6 +152,7 @@
       },
 
       _checkForEnter(e) {
+        this.set('params.ecm_fulltext', this.formatFulltext(this.searchTerm));
```
Gabez0r: Do we need this here, since it is already done in `_search`?

elements/search/assets/nuxeo-assets-search-form.html
```diff
@@ -33,7 +33,7 @@
       type="search"
       placeholder="[[i18n('assetsSearch.fullText.placeholder')]]"
       autofocus
-      on-keydown="_checkForEnter"
+      on-change="_checkForEnter"
```
Gabez0r: This is misleading because we are calling a method that was designed to handle a key press (`_checkForEnter`), and using it to just do a search. That probably why you are modifying `_checkForEnter` with logic that already exists inside `_search`. If we want to add the new behavior of always triggering the search, instead of only when pressing enter, please consider calling something else here, like `_search` directly.

elements/search/trash/nuxeo-trash-search-form.html
```diff
@@ -119,6 +119,11 @@
       },
 
       _checkForEnter(e) {
+        this.set('params.ecm_fulltext', this.formatFulltext(this.searchTerm));
+        this.set(
+          'params.highlight',
+          'dc:title.fulltext,ecm:binarytext,dc:description.fulltext,ecm:tag,note:note.fulltext,file:content.name',
+        );
```
Gabez0r: See comments above.

poonamyadav252 (commented):
Kindly try to close your `open comments`.

elements/search/assets/nuxeo-assets-search-form.html
```diff
@@ -151,7 +151,10 @@
         }
       },
 
-      _checkForEnter(e) {
+      _showResults(e) {
```
rakeshkumar1019: In this function `_showResults(e)` I'm not sure what the purpose of writing if conditions is if we're calling the same `this. search()` inside as well.

elements/search/trash/nuxeo-trash-search-form.html
```diff
@@ -118,7 +118,10 @@
         }
       },
 
-      _checkForEnter(e) {
+      _showResults(e) {
+        // for autoSearch disabled
+        this._search();
```
rakeshkumar1019: Here also same.

rakeshkumar1019 (changes requested):

swarnadipa-dev (dismissed):
Looks good

rakeshkumar1019 (dismissed):
Looks good 👍.

rajatgoel-gl (dismissed):

elements/search/assets/nuxeo-assets-search-form.html
```diff
@@ -33,7 +33,7 @@
       type="search"
       placeholder="[[i18n('assetsSearch.fullText.placeholder')]]"
       autofocus
-      on-keydown="_checkForEnter"
```
Gabez0r, line 36: Is the `_checkForEnter` still used? If not, we should clean it up.

elements/search/trash/nuxeo-trash-search-form.html
```diff
@@ -32,7 +32,7 @@
       value="{{searchTerm}}"
       placeholder="[[i18n('defaultSearch.fullText.placeholder')]]"
       autofocus
-      on-keydown="_checkForEnter"
```
Gabez0r, line 35: Same as above.

Gabez0r (approved):

rakeshkumar1019 (approved):

poonamyadav252 (approved):
Looks fine to me.

Nishant0928 (approved):

rajatgoel-gl (approved):

PR #1619, opened 2022-10-13
restrict lit to version 2.3.1

rakeshkumar1019:


karma.conf.js
```diff
@@ -101,7 +103,9 @@ module.exports = (config) => {
     browserConsoleLogOptions: {
       level: 'error',
     },
-    logLevel: config.LOG_WARN,
+    logLevel: config.LOG_DEBUG,
```
richardsd, line 106: After everything is fixed, you probably want to reset the log level to `LOG_WARN`.

karma.conf.js
```diff
@@ -101,7 +103,9 @@ module.exports = (config) => {
...
+    captureTimeout: 180000,
```
richardsd, line 107: Do we really need such a high value for the `captureTimeout`, or is it just for debugging the problem?

richardsd (commented):
It looks good to me.

karma.conf.js
rakeshkumar1019: @richardsd thank you for the comment.this is for debugging the problem.

PR #1628, opened 2022-10-27
WEBUI-928: implementation of import create button disabled feature

Nishant0928:


rakeshkumar1019 (approved):
Great work  👍 looks good to me.

swarnadipa-dev (approved):
Looks good

SnehalNaik-DEV (approved):
Looks Good.

elements/document/nuxeo-document-import.js
```diff
@@ -879,6 +885,9 @@ Polymer({
   },
 
   _canImportWithMetadata() {
+    if (createDisabled) {
+      return this._getAllFiles().every((file) => 'checked' in file);
```
Gabez0r, line 889: Here we are completely skipping `_canImport`, which verifies if the uploaded has completed. This means that the button to create will most likely be enabled even if the upload is still in progress, for example, which will most likely causes problems.

Gabez0r (commented):
I'm not convinced that this strategy satisfies what's on the description of the ticket:
> The Import documents have a default `Create` and 'Add property ' features. We need to find out whether we can disable the 'Create' button for the particulate user or not.

If it needs to depend on the user, how will a configuration property help?

From my point of view, the ticket asks whether it is possible or not to disable it, and the answer is "no, only with custom code on the project side, at the moment"; anything beyond that, it should be an improvement with clear specifications.

Gabez0r (changes requested):
Please check my comments above.

elements/document/nuxeo-document-import.js
Gabez0r: Please double check this.

elements/document/nuxeo-document-import.js
```diff
@@ -41,8 +41,11 @@ import '@nuxeo/nuxeo-ui-elements/nuxeo-document-layout.js';
 import '../nuxeo-document-creation-stats/nuxeo-document-creation-stats.js';
 import { Polymer } from '@polymer/polymer/lib/legacy/polymer-fn.js';
 import { html } from '@polymer/polymer/lib/utils/html-tag.js';
+import { config } from '@nuxeo/nuxeo-elements';
 import { DocumentCreationBehavior } from '../nuxeo-document-creation/nuxeo-document-creation-behavior.js';
 
+const createDisabled = config.get('importCreateDisabled.buttonDisabeled', false);
```
nelsonsilva, line 47: nit: typo on `buttonDisabled`
also you should probably rename `importCreateDisabled` to something more generic like, at least something like `importCreate`, that way you can add subproperties to it in the future if needed

tmartins: this typo was mentioned during our call this morning

PR #1630, opened 2022-10-27
WEBUI-905: remove from favorites button should have a role and label

GuruprasadAnk1996:
https://jira.nuxeo.com/browse/WEBUI-905

PR #1637, opened 2022-11-01
WEBUI-904: added aria-expanded to quick search button to enable expan…

swarnadipa-dev:
…d/collapse state announced by screen reader

### DIFF
--- a/elements/nuxeo-suggester/nuxeo-suggester.js
+++ b/elements/nuxeo-suggester/nuxeo-suggester.js
@@ -301,6 +301,7 @@ Polymer({
       name="browser"
       on-tap="toggle"
       aria-label$="[[i18n('command.search')]]"
+      aria-expanded="[[toggled]]"
     ></paper-icon-button>
 
     <nuxeo-keys target="[[target]]" keys="up" on-pressed="_upPressed"></nuxeo-keys>


rakeshkumar1019 (commented):
Please see this comment if these changes also apply here.https://github.com/nuxeo/nuxeo-web-ui/pull/1630#discussion_r1007837156

rakeshkumar1019 (dismissed):

Nisha-QA (dismissed):
Looks good

SnehalNaik-DEV (dismissed):
Looks good

GuruprasadAnk1996 (dismissed):

rakeshkumar1019 (dismissed):

rajatgoel-gl (commented):
Going forward kindly do not touch the tag if we are not changing them. Approved for now.
image

semisse (changes requested):
Since `aria-expanded` is an attribute, we should be binding the attribute itself and not the property:
https://polymer-library.polymer-project.org/3.0/docs/devguide/data-binding#attribute-binding

```suggestion
      aria-expanded$="[[toggled]]"
```

Please, also keep in mind what @rajatgoel-gl mentioned. Only introduce the changes that you need (in this case, please leave the `on-tap="toggle"` on line 302).

swarnadipa-dev:
> Since `aria-expanded` is an attribute, we should be binding the attribute itself and not the property: https://polymer-library.polymer-project.org/3.0/docs/devguide/data-binding#attribute-binding
> 
> ```
>       aria-expanded$="[[toggled]]"
> ```
> 
> Please, also keep in mind what @rajatgoel-gl mentioned. Only introduce the changes that you need (in this case, please leave the `on-tap="toggle"` on line 302).

@semisse using aria-expanded$ instead of aria-expanded is not working correctly, as in Chrome screen reader is not announcing the expanded/collapsed state. I have taken reference from 2 other files (the only examples in the codebase) , nuxeo-document-tree.js and nuxeo-checkbox-aggregator.js, where aria-expanded has been used and not aria-expanded$. 

@rajatgoel-gl  Have implemented the change as requested

semisse (approved):

GuruprasadAnk1996 (approved):

rakeshkumar1019 (approved):

SnehalNaik-DEV (approved):
Looks good

rajatgoel-gl (approved):

PR #1654, opened 2022-11-17
WEBUI-948: select all not working for asset search

Nishant0928:
https://jira.nuxeo.com/browse/WEBUI-948


richardsd (changes requested):
Maybe this needs to be confirmed with the product team (cc @bchauvin), because you are not just fixing the select all in the asset search, you are changing the complete screen, by replacing the `nuxeo-justfied-grid` with the `nuxeo-data-grid`.

bchauvin:
@richardsd @Nishant0928  we cannot change the element used here, that would be considered a breaking change for our customers. The solution needs to make use of the existing element.

PR #1658, opened 2022-11-21
WEBUI-977: fixing accessibility check fails in ELEMENTS-1543

GuruprasadAnk1996:
https://jira.nuxeo.com/browse/WEBUI-977

https://jira.nuxeo.com/browse/ELEMENTS-1543

### DIFF
--- a/plugin/a11y/test/specs/home.js
+++ b/plugin/a11y/test/specs/home.js
@@ -7,7 +7,6 @@ import { reportA11y } from '../a11y-reporter.js';
 const EXPECTED_VIOLATIONS = {
   'aria-command-name': 1,
   'aria-allowed-attr': 4,
-  'aria-required-parent': 9,
   'aria-tooltip-name': 1,
   'duplicate-id': 28,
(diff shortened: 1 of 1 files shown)


richardsd (commented):
Ideally, we shouldn't increase the number of violations with code we develop (the goal is to decrease them). 
Maybe it is worth the effort to first analyze if the code you did for ELEMENTS-1543 can be changed to minimize the impact.

GuruprasadAnk1996:
> Ideally, we shouldn't increase the number of violations with code we develop (the goal is to decrease them). Maybe it is worth the effort to first analyze if the code you did for ELEMENTS-1543 can be changed to minimize the impact.

In the new change of ELEMENTS-1543, aria-required-parent violation is not happening hence clearing that

rakeshkumar1019 (approved):

swarnadipa-dev (approved):
Looks good

Nishant0928 (approved):

rajatgoel-gl (approved):

PR #1677, opened 2022-12-05
WEBUI-984: clear for select all grid and list view

rakeshkumar1019:
https://jira.nuxeo.com/browse/WEBUI-984


GuruprasadAnk1996 (commented):
As _excludedDocs is a Number type, you are setting it to empty array, may i know the reason?

PR #1700, opened 2023-01-11
Upgrade test node18

Nisha-QA:


Nisha-QA:
Upgrade to node 18 is on hold as of now

PR #1842, opened 2023-05-17
WEBUI-231: Fix Replace File dialog

rahuljain-dev:
https://jira.nuxeo.com/browse/WEBUI-231

### DIFF
--- a/elements/nuxeo-document-actions/nuxeo-replace-blob-button.js
+++ b/elements/nuxeo-document-actions/nuxeo-replace-blob-button.js
@@ -56,7 +56,7 @@ Polymer({
 
     <nuxeo-dialog id="dialog" with-backdrop>
       <h2>[[i18n('replaceBlobButton.dialog.heading')]]</h2>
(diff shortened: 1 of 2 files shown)


swarnadipa-dev (approved):

poonamyadav252 (commented):
As we have already a method `hasFiles` then why we are not using use the same file, and why there be required to create a new method `hasFilesUploaded` with the same property

rahuljain-dev:
> As we have already a method `hasFiles` then why we are not using use the same file, and why there be required to create a new method `hasFilesUploaded` with the same property

Has files method is used in nuxeo dropzone for progress bar and other functionality , this is the issue only previously same has been used for disabled the replace button, so whenever the dropzone has the file it enables the button, if we keep the same property in nuxeo replace blob button file then we need to do lot of changes in nuxeo dropzone file.

poonamyadav252 (approved):

rajatgoel-gl (approved):

PR #1860, opened 2023-06-09
WEBUI-1140: Fix search actions display after refreshing the page

rahuljain-dev:
https://jira.nuxeo.com/browse/WEBUI-1140

### DIFF
--- a/elements/search/nuxeo-saved-search-actions.js
+++ b/elements/search/nuxeo-saved-search-actions.js
@@ -105,7 +105,7 @@ Polymer({
     _isSearchFormVisible: Boolean,
   },
 
-  ready() {
+  created() {
     this._searchFormVisibilityChanged = (event) => {
       this._isSearchFormVisible = event.target.visible;
     };
--- a/elements/search/nuxeo-search-form.js
+++ b/elements/search/nuxeo-search-form.js
@@ -589,7 +589,10 @@ Polymer({
      * @ignore
      * If `true`, the current element is visible.
      * */
-    visible: Boolean,
+    visible: {
+      type: Boolean,
+      notify: true,
+    },
     /**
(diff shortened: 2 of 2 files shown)


elements/search/nuxeo-saved-search-actions.js
```diff
@@ -105,7 +105,7 @@ Polymer({
     _isSearchFormVisible: Boolean,
   },
 
-  ready() {
+  created() {
     this._searchFormVisibilityChanged = (event) => {
       this._isSearchFormVisible = event.target.visible;
     };
```
poonamyadav252, line 111: could you please elaborate that why we are changing the method `ready` to `created`

elements/search/nuxeo-saved-search-actions.js
```diff
@@ -105,7 +105,7 @@ Polymer({
     _isSearchFormVisible: Boolean,
   },
 
-  ready() {
+  created() {
```
poonamyadav252, line 108: please check the reference here:
https://github.com/nuxeo/nuxeo-web-ui/pull/1860#discussion_r1223996187

elements/search/nuxeo-saved-search-actions.js
rahuljain-dev: > could you please elaborate that why we are changing the method `ready` to `created`

we are using properties like isSearchFormVisible so created called at the root level and ready when our DOM loads so when we are setting isSearchFormVisible it will be undefined as this is not yet bind in ready call and that can be resolved using create instead of ready

rakeshkumar1019 (approved):

rakeshkumar1019 (approved):

poonamyadav252 (approved):

rajatgoel-gl (approved):

PR #1897, opened 2023-07-28
WEBUI-1224: (Backport) expose value-key on nuxeo-document-attachments.

poonamyadav252:


rakeshkumar1019 (commented):
Here, creating a new variable`fileName` can be omitted.
`return this.document.type === 'File' && this.xpath === 'files:files' ? 'file' : '';`

@rakeshkumar1019 yes, it can be, but we already merge the code in maintenance-3.1.x so we can't change it because we need to make a sync between both repo.

Nishant0928 (approved):

rakeshkumar1019 (approved):

rajatgoel-gl (approved):

PR #1925, opened 2023-08-09
WEBUI-1066: Fix browser tab name displaying "app.title.undefined" after refresh

rahuljain-dev:
https://jira.nuxeo.com/browse/WEBUI-1066

### DIFF
--- a/elements/nuxeo-app.js
+++ b/elements/nuxeo-app.js
@@ -823,6 +823,7 @@ Polymer({
   },
 
   _updateTitle() {
+    if (!this.page) return;
     const title = [];
     switch (this.page) {
       case 'browse':


elements/nuxeo-app.js
```diff
@@ -823,6 +823,7 @@ Polymer({
   },
 
   _updateTitle() {
+    if (!this.page) return;
```
swarnadipa-dev, line 826: please write the return statement in the next line for better readability.

swarnadipa-dev (changes requested):

elements/nuxeo-app.js
rahuljain-dev: @swarnadipa-dev If we do so prettier will move it to single line, either we can do like this:
if (!this.page) {
      return;
    }
    
    But I don't think it's a good option.

rakeshkumar1019 (approved):

swarnadipa-dev (approved):

rajatgoel-gl (approved):

PR #2342, opened 2024-10-25
Create messages-sq.json

amirbuzo:
Add Albanian language for nuxeo-web-ui!
I test the language on local instance and it seems correct!


amirbuzo:
Add Albanian Language Support to Nuxeo Web UI



Please include Albanian (sq) language support in the Nuxeo Web UI. We have tested it on our local instances, and it works correctly.
We kindly request that Albanian be included by default in future releases.

Thank you!

PR #2615, opened 2025-05-26
WEBUI-1719: veracode ter-fs vuln

rakeshkumar1019:


rahuljain-dev (approved):

packages/nuxeo-web-ui-ftest/package.json
```diff
@@ -14,7 +14,8 @@
     "node": ">=18.0.0"
   },
   "overrides": {
-    "ws": "8.17.1"
+    "ws": "8.17.1",
+    "tar-fs": "^3.0.9"
```
madhurkulshrestha-hyland, line 18: is this required in main package since it only affects ftests ?

madhurkulshrestha-hyland (changes requested):

packages/nuxeo-web-ui-ftest/package.json
madhurkulshrestha-hyland: As of now, some packages for ftest are in main package.json which will be moved to ftest package.json with Node22 upgrade efforts. This line can be removed from main package.json at that time.

madhurkulshrestha-hyland (approved):

Nishant0928 (approved):

PR #2632, opened 2025-06-17
WEBUI-1678: inflight vulnerability overrides glob

rakeshkumar1019:
https://hyland.atlassian.net/browse/WEBUI-1678

### DIFF
--- a/packages/nuxeo-designer-catalog/package.json
+++ b/packages/nuxeo-designer-catalog/package.json
@@ -22,7 +22,7 @@
     "check-engine": "^1.10.1",
     "concat-stream": "^1.6.2",
     "decamelize": "^1.2.0",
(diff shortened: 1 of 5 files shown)


packages/nuxeo-designer-catalog/package.json
```diff
@@ -13,6 +13,9 @@
     "postinstall": "check-engine --ignore",
     "clean-data": "find data/applications/nuxeo/**/ ! -name 'catalog-packages.json' -delete"
   },
+  "overrides": {
+    "glob": "^11.0.2"
+  },
```
madhurkulshrestha-hyland, line 18: is it required to be added here as well ? Should work with global one ?

madhurkulshrestha-hyland (changes requested):

PR #2718, opened 2025-09-08
WEBUI-1747 -  fix keyboard navigation DOM order

swarnadipa-dev:
https://hyland.atlassian.net/browse/WEBUI-1747

### DIFF
--- a/elements/nuxeo-app.js
+++ b/elements/nuxeo-app.js
@@ -754,7 +754,7 @@ Polymer({
     this.drawerWidth = this.sidebarWidth = getComputedStyle(this).getPropertyValue('--nuxeo-sidebar-width');
 
     this.$.drawerPanel.addEventListener('opened-changed', () => {
-      this.$.drawerPanel.parentElement.notifyResize();
+      window.dispatchEvent(new Event('resize'));
     });
 
     const { toast } = this.$;
@@ -1780,7 +1780,7 @@ Polymer({
     }
     const firstItem = drawer.querySelector('a, button, [tabindex]:not([tabindex="-1"])');
     if (firstItem) {
-      firstItem.focus();
+      //  firstItem.focus();
     }
   },
 
--- a/elements/nuxeo-app.js
+++ b/elements/nuxeo-app.js
@@ -809,10 +809,7 @@ Polymer({
       const lastItem = items[items.length - 1];
       const active = e.target;
 
-      if (e.key === 'ArrowUp' && active === firstItem) {
-        e.preventDefault();
-        logo.focus();
-      } else if (e.key === 'ArrowDown' && active === lastItem) {
+      if ((e.key === 'ArrowUp' && active === firstItem) || (e.key === 'ArrowDown' && active === lastItem)) {
         e.preventDefault();
         logo.focus();
       }
--- a/elements/nuxeo-app.js
+++ b/elements/nuxeo-app.js
@@ -109,36 +109,40 @@ Polymer({
         * for more details, see: https://github.com/PolymerElements/paper-dialog/issues/44#issuecomment-172013206
         * this will only work for iOS since it's the only supporting \`-webkit-overflow-scrolling\`
         */
+
       :host {
-        --paper-header-panel-container: {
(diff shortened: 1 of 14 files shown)


elements/nuxeo-app.js
```diff
@@ -723,17 +778,64 @@ Polymer({
...
+    window.removeEventListener('resize', this._updateIsNarrow.bind(this));
+    super.disconnectedCallback();
+  },
+
+  skipLinkEvent() {
+    const { skipLink, mainContent } = this.$;
+    let keyboardUsed = false;
+
+    const handleFirstTab = (e) => {
+      if (!keyboardUsed && e.key === 'Tab') {
```
madhurkulshrestha-hyland: Will the skip link be activated only on first tab ??? what about if I cycle back to it using tabs ?

madhurkulshrestha-hyland (changes requested):

elements/nuxeo-app.js
swarnadipa-dev: yes, this is just to detect the first time user presses tab key (as skip link should not be activated on mouse navigation). It will be activated again if user comes back to it using tab.

elements/nuxeo-app/nuxeo-menu-icon.js
```diff
@@ -68,11 +68,23 @@ Polymer({
...
+      paper-icon-button path {
+        tabindex: -1;
+      }
     </style>
 
     <a href$="[[_href(urlFor, route, link)]]">
-      <paper-icon-button noink id="button" name$="[[name]]" aria-labelledby="tooltip"></paper-icon-button>
-      <nuxeo-tooltip for="button" position="[[_tooltipPosition]]" offset="0" animation-delay="0" id="tooltip"
+      <paper-icon-button noink id="button" name$="[[name]]" aria-labelledby="tooltip" tabindex="-1"></paper-icon-button>
+      <nuxeo-tooltip
```
madhurkulshrestha-hyland, line 81: Does it required explicit setting of -1 for tabindex. Is it focussable by default ?

elements/nuxeo-app.js
```diff
@@ -355,149 +399,149 @@ Polymer({
...
-              <nuxeo-card>
-                <nuxeo-error id="error"></nuxeo-error>
-              </nuxeo-card>
-            </div>
-          </nuxeo-page>
-        </iron-pages>
+      </app-drawer>
+
+      <!-- Main content -->
+      <app-header-layout tabindex="-1">
```
madhurkulshrestha-hyland: <app-header-layout> does not get focussed by default. So -1 is not required here. Do check

elements/nuxeo-app.js
```diff
@@ -355,149 +399,149 @@ Polymer({
...
+              icon="menu"
+              on-tap="_openDrawer"
+              hidden$="[[!isNarrow]]"
+              aria-label$="[[i18n('command.menu')]]"
+              tabindex="-1"
+            ></paper-icon-button>
+          </app-toolbar>
+        </app-header>
+
+        <main id="mainContent" tabindex="-1">
```
madhurkulshrestha-hyland, line 504: <main> does not get focussed by default. So -1 is not required here. Do check

madhurkulshrestha-hyland (changes requested):

elements/nuxeo-app/nuxeo-menu-icon.js
swarnadipa-dev: required to prevent extra tabbing to go from menu icon to secondary menu

elements/nuxeo-app.js
swarnadipa-dev: Removed

elements/nuxeo-app.js
swarnadipa-dev: since we are programmatically putting focus on <main> which is by default not a focusable element, tabindex=-1 is required there to make it focusable

madhurkulshrestha-hyland (approved):

Nishant0928 (approved):

PR #2779, opened 2025-10-27
NXBT-3895: Automatically approve and merge Dependabot pull requests

ataillefer:


madhurkulshrestha-hyland:
Thank you for generating this PR @ataillefer . However, we are taking a conscious decision to NOT enable auto-merge of dependabot PRs. This is because there are instances where this may break the existing functionalities. A recent classic example is - https://github.com/nuxeo/nuxeo-web-ui/pull/2751 where I had to add an additional commit with the dependabot change

PR #2829, opened 2025-12-03
OPSEXP-3640 Use xvfb-record action from build tools

pmacius:


.github/workflows/ftest.yaml
```diff
@@ -136,7 +136,7 @@ jobs:
         env:
           RUN_ALL: false
           BAIL: 0
```
pmacius, line 141: Not sure for what they are used for

alxgomz: yeah just keep them these are related to the ftest execution IIRC

alxgomz (approved):
OK but change action ref after merged on build tools

alxgomz (approved):

madhurkulshrestha-hyland (approved):

PR #2960, opened 2026-03-02
WEBUI-1544: Fix nuxeo-dropzone blob list display in create/edit/import layouts [LTS-2023]

AnilKumarVanga:
https://hyland.atlassian.net/browse/WEBUI-1544

### DIFF
--- a/elements/nuxeo-dropzone/nuxeo-dropzone.js
+++ b/elements/nuxeo-dropzone/nuxeo-dropzone.js
@@ -401,14 +401,9 @@ Polymer({
       type: String,
       computed: 'formatPropertyXpath(xpath)',
     },
-
-    /**
-     * If false (default), each upload will re-upload all files (previous + new).
-     * If true, only new files are uploaded.
-     */
-    replaceMode: {
-      type: Boolean,
-      value: false,
+    _allUploadedFiles: {
+      type: Array,
+      value: () => [],
     },
   },
 
@@ -454,6 +449,13 @@ Polymer({
         this.value = [];
       }
       this.push('value', ...value);
+      const successfulFiles = this.files.filter((f) => !f.error);
+
+      // accumulate uploader-managed objects
+      this._allUploadedFiles = [...this._allUploadedFiles, ...successfulFiles];
+
+      // restore cumulative list into files
+      this.set('files', [...this._allUploadedFiles]);
     } else {
       this.set('value', value);
     }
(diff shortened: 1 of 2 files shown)


elements/nuxeo-dropzone/nuxeo-dropzone.js
```diff
@@ -401,14 +401,9 @@ Polymer({
...
     },
-
-    /**
-     * If false (default), each upload will re-upload all files (previous + new).
-     * If true, only new files are uploaded.
-     */
-    replaceMode: {
-      type: Boolean,
-      value: false,
+    _allUploadedFiles: {
```
madhurkulshrestha-hyland, line 404: there is already `uploadedFiles` property that is being used to track all uploaded files. This may introduce duplicacy

elements/nuxeo-dropzone/nuxeo-dropzone.js
```diff
@@ -589,14 +595,9 @@ Polymer({
...
-        if (this.replaceMode) {
-          this.uploadedFiles = this.uploadedFiles.concat(newFiles);
-          this.uploadFiles(newFiles);
-        } else {
-          const allFiles = this.uploadedFiles.concat(newFiles);
-          this.uploadedFiles = allFiles;
-          this.uploadFiles(allFiles);
-        }
+        const allFiles = this.uploadedFiles.concat(newFiles);
+        this.uploadedFiles = allFiles;
```
madhurkulshrestha-hyland: Both these lines are modifying the same variable `this.uploadedFiles` . Optimize it

madhurkulshrestha-hyland (changes requested):

elements/nuxeo-dropzone/nuxeo-dropzone.js
AnilKumarVanga: I initially tried reusing `uploadedFiles`, but it is managed by the uploader behavior and accumulates files across batches. When the next batch completes, it contains both the previous and new batch files again, which causes duplicates when rebuilding the `files` list for the UI.

The `_allUploadedFiles` array is used as a controlled accumulator that only stores the successfully processed files from each batch. This allows us to maintain a cumulative list for display while avoiding duplication caused by the uploader’s internal tracking.

madhurkulshrestha-hyland, line 404: Add Comments of why this attribute is introduced

madhurkulshrestha-hyland (changes requested):

swarnadipa-dev (approved):

madhurkulshrestha-hyland (approved):

PR #3041, opened 2026-04-13
WEBUI-1553: Fix results selection actions toolbar not appearing after refreshing search [LTS-2023]

naveen-konda:
Enhance safety checks for view methods and item retrieval in nuxeo-results.js

- Added defensive checks to ensure view methods are only called if the view exists and the method is available.
- Improved the items getter to handle cases where the view or its internal list may not be ready, preventing potential errors during navigation or refresh.
- Updated action context handling to avoid setting empty items during timing issues with the view's internal state.

### DIFF
--- a/elements/nuxeo-results/nuxeo-results.js
+++ b/elements/nuxeo-results/nuxeo-results.js
@@ -421,7 +421,7 @@ Polymer({
    *
    * Right after navigation or refresh, `view` may exist while `$` / `$.list` are not ready yet, or
    * reading `items` can throw inside iron-list. Those cases return `[]` so observers and toolbars
-   * do not break (WEBUI-1553).
+   * do not break.
    */
   get items() {
     if (!this.view) {
@@ -536,7 +536,7 @@ Polymer({
 
   reset() {
     if (this.view) {
-      // Guard: only call reset if view exists and method is available (WEBUI-1553)
+      // Guard: only call reset if view exists and method is available.
       if (typeof this.view.reset === 'function') {
         this.view.reset();
       }
@@ -650,7 +650,7 @@ Polymer({
   _updateActionContext() {
     /* Always publish the base action context so slots receive a model even during
      * the timing window where items are not ready yet; only resolve items once
-     * one of the supported data paths is available to avoid flicker (WEBUI-1553).
+     * one of the supported data paths is available to avoid flicker.
      */
     try {
       const hasItems = this.view && Array.isArray(this.view.items);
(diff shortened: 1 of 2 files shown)


elements/nuxeo-results/nuxeo-results.js
```diff
@@ -597,7 +536,10 @@ Polymer({
 
   reset() {
     if (this.view) {
-      this.view.reset();
+      // Guard: only call reset if view exists and method is available (WEBUI-1553)
```
madhurkulshrestha-hyland: No need to add ticket number for reference

madhurkulshrestha-hyland (changes requested):

elements/nuxeo-results/nuxeo-results.js
naveen-konda: Sure, I have updated the PR.

elements/nuxeo-results/nuxeo-results.js
```diff
@@ -623,16 +648,57 @@ Polymer({
...
+       * items field: use resolved items when ready, otherwise preserve previous items to prevent
+       * action toolbar from breaking during view initialization timing windows.
+       */
+      this.actionContext = {
+        baseUrl: this.$.nxcon.url,
+        displayMode: this.displayMode,
+        nxProvider: this.nxProvider,
+        selectedItems: this.selectedItems,
+        items: canResolveItems ? this.items : Array.isArray(previousItems) ? previousItems : [],
+        columns: this.columns,
```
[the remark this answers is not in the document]
naveen-konda: Thanks for the review! I've analyzed this concern and the current implementation is actually sufficient. Here's why:

**The temporary `[]` is handled by existing architecture:**

1. **Observer watches `view.*`** (line 404): When items become ready, the observer fires again automatically
2. **Explicit event listener** (line 581): We listen to `'items-changed'` event which fires when iron-list is initialized
3. **Natural recovery**: The temporary `[]` value is quickly replaced once items populate

**Timeline during refresh:**
```
t=0ms:   view.$.list created (not ready yet)
t=10ms:  _updateActionContext() fires → actionContext.items = [] temporarily
t=100ms: iron-list ready → 'items-changed' event fires
         → _updateActionContext() fires AGAIN → actionContext.items = [real items]
         → Toolbar appears naturally ✓
```

**Why preserving previous items would add complexity:**
- Need to distinguish "not ready yet" vs "legitimately empty search results"
- Risk showing stale items from different search when context changes
- Requires additional state management and edge case handling

I've added clarifying comments in the code (lines 653-658) documenting this behavior. The current approach prioritizes simplicity and correctness while still fixing WEBUI-1553.

If we observe actual issues during testing where the toolbar fails to appear, we can apply a targeted fix. For now, the simpler implementation is more maintainable.

madhurkulshrestha-hyland (approved):

swarnadipa-dev (approved):

PR #3131, opened 2026-05-12
fix(sonar): include addon test directories in sonar.tests and sonar.test.inclusions

swarnadipa-dev:
## Problem

Addon unit tests (`addons/*/test/**/*.test.js`) were invisible to SonarCloud because `sonar.tests` only pointed to the root `test/` directory.

This caused coverage for addon source files (e.g. `addons/nuxeo-drive/elements/nuxeo-drive-download-button.js`) to show as **N/A** instead of being measured. When coverage is N/A, SonarCloud skips the coverage gate condition entirely — allowing PRs that introduce uncovered code to pass.

## Fix

- Expand `sonar.tests` to include all addon test directories that contain test files:
  - `addons/nuxeo-csv/test`
  - `addons/nuxeo-drive/test`
  - `addons/nuxeo-liveconnect/test`
  - `addons/nuxeo-template-rendering/test`
  - `addons/nuxeo-wopi/test`
- Expand `sonar.test.inclusions` to match `addons/*/test/**/*.test.js`

## Impact

After this fix, SonarCloud will correctly measure coverage for addon source files and apply the Quality Gate coverage condition on new code in addon files.

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -6,7 +6,13 @@ sonar.organization=nuxeo
 sonar.sources=elements,addons,themes,i18n,index.js,legacy.js,.github
 
 # Test files (root + addon test directories)
-sonar.tests=test,addons/nuxeo-csv/test,addons/nuxeo-drive/test,addons/nuxeo-liveconnect/test,addons/nuxeo-template-rendering/test,addons/nuxeo-wopi/test
+sonar.tests=\
+  test,\
+  addons/nuxeo-csv/test,\
+  addons/nuxeo-drive/test,\
+  addons/nuxeo-liveconnect/test,\
+  addons/nuxeo-template-rendering/test,\
+  addons/nuxeo-wopi/test
 
 # Coverage report from Karma + Istanbul
 sonar.javascript.lcov.reportPaths=coverage/lcov.info
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -5,8 +5,8 @@ sonar.organization=nuxeo
 # JS/HTML front-end sources
 sonar.sources=elements,addons,themes,i18n,index.js,legacy.js,.github
 
-# Test files
-sonar.tests=test
+# Test files (root + addon test directories)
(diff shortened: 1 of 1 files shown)


sonar-project.properties
```diff
@@ -5,8 +5,14 @@ sonar.organization=nuxeo
 # JS/HTML front-end sources
 sonar.sources=elements,addons,themes,i18n,index.js,legacy.js,.github
 
-# Test files
-sonar.tests=test
+# Test files (root + addon test directories)
```
madhurkulshrestha-hyland, line 8: cant we have addons/**/test

madhurkulshrestha-hyland (changes requested):

swarnadipa-choudhury:
Thanks @Copilot — good catch. This has been addressed in commit [3f73f5c](https://github.com/nuxeo/nuxeo-web-ui/commit/3f73f5c9b).

`addons/**/test/**` has been added to `sonar.exclusions`, which prevents addon test directories from being processed as source code even though `addons/` is listed in `sonar.sources`. The directories listed in `sonar.tests` are the authoritative test roots, and the exclusion ensures there is no double-indexing.

sonar-project.properties
```diff
@@ -5,8 +5,8 @@ sonar.organization=nuxeo
...
+sonar.tests=test,addons/nuxeo-csv/test,addons/nuxeo-drive/test,addons/nuxeo-liveconnect/test,addons/nuxeo-template-rendering/test,addons/nuxeo-wopi/test
```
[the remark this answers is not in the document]
swarnadipa-dev: addons/**/test/** has been added to sonar.exclusions in commit to prevent addon test dirs from being processed as source code.

sonar-project.properties
swarnadipa-dev: sonar.tests does not support glob or wildcard patterns. The official SonarQube Cloud docs state:
"The entries in the list are simple paths. Wildcard patterns are not allowed."

Source: [Setting initial scope — SonarQube Cloud docs]

A value like addons/**/test would be treated as a literal path, find nothing on disk, and silently register no tests. The explicit directory list is the only supported approach for this property.

madhurkulshrestha-hyland (approved):

PR #3133, opened 2026-05-13
WEBUI-2035: Quick filter needs to be clicked twice to be unchecked [LTS-2023]

AnilKumarVanga:
https://hyland.atlassian.net/browse/WEBUI-2035


test/nuxeo-results.test.js
```diff
@@ -1019,6 +1019,62 @@ suite('nuxeo-results', () => {
...
+      results.quickFilters = ['Validated'];
+
+      results._onQuickFiltersChanged({ detail: { value: ['Most Recent'] } });
+      await flush();
+
+      expect(results.quickFilters).to.deep.equal(['Most Recent']);
+      expect(results.nxProvider.quickFilters).to.deep.equal(['Most Recent']);
+      expect(results.view.quickFilters).to.deep.equal(['Most Recent']);
+
+      fetchSpy.restore();
```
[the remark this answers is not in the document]
AnilKumarVanga: thanks for point out , fixed it

elements/nuxeo-results/nuxeo-results.js
```diff
@@ -511,6 +512,39 @@ Polymer({
...
+    let eventFilters = this.quickFilters;
+    if (Array.isArray(e?.detail?.value)) {
+      eventFilters = e.detail.value;
+    } else if (Array.isArray(e?.target?.quickFilters)) {
+      eventFilters = e.target.quickFilters;
+    }
+    const filters = this._cloneQuickFilters(eventFilters);
+    this.quickFilters = this._cloneQuickFilters(filters);
+    this._pendingQuickFilters = this._cloneQuickFilters(filters);
+    this._quickFiltersDirty = true;
```
madhurkulshrestha-hyland: **Redundant cloning**: `_cloneQuickFilters` already returns a new array via `.slice()`. Cloning an already-cloned array is unnecessary overhead.

Here for example:
```js
const filters = this._cloneQuickFilters(eventFilters); // clone #1
this.quickFilters = this._cloneQuickFilters(filters);   // clone #2 (of clone #1)
this._pendingQuickFilters = this._cloneQuickFilters(filters); // clone #3 (of clone #1)
```

Since quick filters are arrays of strings (no nested references), a single clone suffices:
```js
const filters = this._cloneQuickFilters(eventFilters);
this.quickFilters = filters;
this._pendingQuickFilters = filters.slice(); // only need one extra copy if both must be independent
```

Same pattern repeats in `_quickFiltersChanged`, `_enforcePendingQuickFilters`, and the view-selection restore block.

elements/nuxeo-results/nuxeo-results.js
```diff
@@ -468,6 +468,7 @@ Polymer({
```
madhurkulshrestha-hyland, line 468: **Re-entrancy risk**: This observer sets `this.quickFilters` inside itself, which can re-trigger the same `'_enforcePendingQuickFilters(quickFilters.*)'` observer. The `_quickFiltersEqual` guard prevents infinite loops, but Polymer property effects can batch or defer in unexpected ways.

Consider adding a reentrancy guard:
```js
_enforcePendingQuickFilters() {
  if (this._enforcingPending) return;
  if (!this._quickFiltersDirty || !this._pendingQuickFilters) return;
  if (this._quickFiltersEqual(this.quickFilters, this._pendingQuickFilters)) return;

  this._enforcingPending = true;
  try {
    // ... apply pending filters
  } finally {
    this._enforcingPending = false;
  }
}
```

elements/nuxeo-results/nuxeo-results.js
```diff
@@ -511,6 +512,39 @@ Polymer({
...
+    const filters = this._cloneQuickFilters(eventFilters);
+    this.quickFilters = this._cloneQuickFilters(filters);
+    this._pendingQuickFilters = this._cloneQuickFilters(filters);
+    this._quickFiltersDirty = true;
+
+    // Keep provider/view in sync before fetching to avoid stale quick-filter state after navigation.
+    if (this.nxProvider) {
+      this.set('nxProvider.quickFilters', this._cloneQuickFilters(filters));
+    }
+    if (this.view?.quickFilters !== undefined) {
```
madhurkulshrestha-hyland, line 531: **`_quickFiltersDirty` may never be cleared**: This flag is only cleared inside `_quickFiltersChanged` when `incomingFilters` matches `_pendingQuickFilters`. If the provider never echoes back the exact pending value (e.g., server returns filters in a different order, or a navigation/view-switch/reset occurs without triggering the matching path), the flag stays set **permanently**.

This would cause `_quickFiltersChanged` to reject all future legitimate provider updates, and `_enforcePendingQuickFilters` to keep overwriting any new state.

Suggestion: Clear the dirty flag on navigation, on explicit user reset, or add a timeout/TTL safety net:
```js
// In _onQuickFiltersChanged, after setting dirty:
this._quickFilterDirtyTimeout = this.async(() => {
  this._quickFiltersDirty = false;
  this._pendingQuickFilters = null;
}, 5000);
```

elements/nuxeo-results/nuxeo-results.js
```diff
@@ -511,6 +512,39 @@ Polymer({
     return [];
```
madhurkulshrestha-hyland, line 512: **Confusing naming**: Both `_onQuickFiltersChanged` (this new handler) and the existing `_quickFiltersChanged` (view's `quick-filters-changed` listener, line 942) deal with quick filter changes but serve different purposes:
- `_onQuickFiltersChanged` → handles the **user action** from `nuxeo-quick-filters` component
- `_quickFiltersChanged` → handles the **view echoing back** filter state

These are easy to confuse during maintenance. Consider more descriptive names like `_handleUserQuickFilterToggle` and `_handleViewQuickFiltersSync`.

elements/nuxeo-app.js
```diff
@@ -1263,6 +1264,20 @@ Polymer({
...
+    this.searchForm = this.$$(`[search-name='${this.searchName}']`) || this.searchForm;
+
+    if (!this.searchForm || !Array.isArray(e?.detail?.quickFilters)) {
+      return;
+    }
+
+    const quickFilters = e.detail.quickFilters.slice();
+    this.searchForm.set('_quickFilters', quickFilters);
+    if (this.searchForm?.$.provider) {
+      this.searchForm.$.provider.quickFilters = quickFilters.slice();
```
madhurkulshrestha-hyland: **Multiple overlapping sync paths**: This `_syncSearchQuickFilters` in `nuxeo-app` sets `searchForm._quickFilters` and `searchForm.$.provider.quickFilters`. However, `nuxeo-search-form._syncQuickFiltersFromResults` (added in this same PR) also sets the same properties by listening to `quick-filters-changed` from the results element.

For a single user toggle, both paths fire — resulting in the search form's provider being set twice with the same value. While not harmful, it suggests the sync architecture could be simplified to a single path. Consider whether this app-level handler is needed given the direct listener in `nuxeo-search-form`.

elements/nuxeo-results/nuxeo-results.js
```diff
@@ -940,8 +987,69 @@ Polymer({
...
+  _enforcePendingQuickFilters() {
+    if (!this._quickFiltersDirty || !this._pendingQuickFilters) {
+      return;
+    }
+
+    if (this._quickFiltersEqual(this.quickFilters, this._pendingQuickFilters)) {
+      return;
+    }
+
+    const pendingFilters = this._cloneQuickFilters(this._pendingQuickFilters);
```
madhurkulshrestha-hyland: **`JSON.stringify` for equality check**: This works for arrays of strings but has subtle edge cases:
- If quick filter entries are ever objects (not just strings), property ordering differences would cause false negatives
- `JSON.stringify(undefined)` returns `undefined` (not a string), so the `Array.isArray(a) ? a : []` fallback is critical

This is acceptable for the current string-array format, but consider adding a brief comment documenting the assumption that quick filters are always flat string arrays, so future changes don't silently break this comparison.

madhurkulshrestha-hyland (changes requested):
The fix correctly addresses the root cause (stale provider state overwriting user intent), and test coverage is comprehensive. However, the implementation is over-engineered with redundant cloning, multiple parallel sync paths, and a dirty flag that may never expire. See inline comments for specific concerns and suggestions.

elements/nuxeo-results/nuxeo-results.js
AnilKumarVanga: thanks for review , i will resolve it

elements/nuxeo-results/nuxeo-results.js
AnilKumarVanga: Resolved

elements/nuxeo-results/nuxeo-results.js
```diff
@@ -511,6 +512,41 @@ Polymer({
...
+    this._quickFiltersDirty = true;
+
+    // Safety net: clear dirty flag after 5 s in case the view never echoes back the pending value
+    if (this._quickFilterDirtyTimeout) {
+      this.cancelAsync(this._quickFilterDirtyTimeout);
+    }
+    this._quickFilterDirtyTimeout = this.async(() => {
+      this._quickFiltersDirty = false;
+      this._pendingQuickFilters = null;
+    }, 5000);
```
madhurkulshrestha-hyland: Need to avoid random timeout of 5 seconds

madhurkulshrestha-hyland (changes requested):

elements/nuxeo-results/nuxeo-results.js
AnilKumarVanga: Removed random tiemout

madhurkulshrestha-hyland (approved):

vaibhavagarwal4-lab (approved):

PR #3186, opened 2026-05-25
WEBUI-2040: Ability to adjust the width of the Browse pane, Search pane, and Information pane [LTS-2023]

naveen-konda:
# ELEMENTS-1844 / WEBUI-2040 — Feature Summary

> Resizable Browse / Search drawer and Information pane in Nuxeo Web UI.

**Shipping ticket:** [WEBUI-2040](https://github.com/nuxeo/nuxeo-web-ui/pull/3186) — *Ability to adjust the width of the Browse pane, Search pane, and Information pane* ([PR #3186](https://github.com/nuxeo/nuxeo-web-ui/pull/3186)).

**Note:** Browse and Search are not separate layouts — both use the **same left drawer** in `nuxeo-app` (different drawer tabs). Only the **Information pane** on document pages is a second resizable panel.


**Dual-repo shipping:** UI work spans **nuxeo-web-ui** (WEBUI-2040) and **nuxeo-elements** (ELEMENTS-1844 — shared `nuxeo-resize-handle` widget, i18n keys, tooltip clone styling). Both must land (elements RC published before or with web-ui) for CI to resolve imports.

### At a glance (PM / QA)


| Question               | Answer                                                                                          |
| ---------------------- | ----------------------------------------------------------------------------------------------- |
| What can users resize? | **Left:** Browse/Search drawer (all routes). **Right:** Information pane (document pages only). |
| How?                   | 6 px edge — drag, keyboard, double-click / Enter to reset.                                      |
| Remembered?            | Yes — per browser (`localStorage`).                                                             |
| Breaks mobile?         | No — handles **hidden** ≤720 px (drawer) / ≤1024 px stacked layout (info pane).                 |
| Squeezes the document? | No — main column keeps **≥ half** of doc-page width (240–640 px band).                          |
| Two panes fighting?    | Info pane can ask app to **shrink drawer** (`nuxeo-shrink-drawer`).                             |
| Ship order             | **nuxeo-elements RC first** (or same release train), then web-ui.                               |
| RTL locales            | Handles stay on the **inner** edge toward main content; **arrow keys mirror** (§ 7a).           |


---

## 1. What this feature does

Users can now **drag the edge** of two side panels to make them wider or narrower:

- **Left panel** — the Browse / Search drawer that opens when you click an icon on the side-nav.
- **Right panel** — the Information pane on document detail pages.

The chosen width is **remembered across page refreshes** (per browser).

That's it. No new buttons, no new menus — just a draggable edge.

---

## 2. How it looks

**Wide desktop (document page)** — the info pane lives *inside* the main column, not beside the whole app:

```
┌ nuxeo-app ─────────────────────────────────────────────────────────┐
│ sidebar │ Browse/Search drawer │  #mainContent (any route)         │
│  52 px  │  ⇨ drag here         │  ┌ nuxeo-document-page ────────┐ │
│         │                      │  │ document view │ Info ⇦ drag │ │
│         │                      │  │    (.main)    │  (.side)    │ │
│         │                      │  └──────────────────────────────┘ │
└────────────────────────────────────────────────────────────────────┘
```

On **search / tasks / admin** routes, only the **drawer** handle applies (no info pane).

Each handle is a `**nuxeo-resize-handle`** strip (6 px). It is **invisible until hover/focus** (`--nuxeo-resize-handle-color` per theme). Double-click or **Enter / Space** (when focused) resets that pane to its default width.

---

## 3. What was added (high level)


| Area            | What changed                                                                                                                                                                                                                                                                                         |
| --------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Shared widget   | `**nuxeo-resize-handle`** in **nuxeo-elements** — 6 px strip, pointer/keyboard/RTL, tooltip, ARIA; hosts listen for `resize-*` events.                                                                                                                                                               |
| Left drawer     | Resize logic in `**NuxeoAppDrawerResizeBehavior`** (`elements/behaviors/nuxeo-app-drawer-resize-behavior.js`); shell wiring in `nuxeo-app.js`.                                                                                                                                                       |
| Right info pane | Clamp math, persistence, push-back in `elements/document/nuxeo-document-page.js`; same widget in the template.                                                                                                                                                                                       |
| Memory          | Widths stored in `localStorage` (`nuxeo.drawerWidth`, `nuxeo.documentPage.sidePaneWidth`).                                                                                                                                                                                                           |
| Coordination    | If the info pane needs more width at its cap, it fires `nuxeo-shrink-drawer` so the drawer shrinks and the main column stays safe.                                                                                                                                                                   |
| Layout notify   | App-level `**_notifyLayoutChanged()**` → `notifyResize()` + guarded `window.resize` for **all** main-content routes; drawer **drag** uses iron-resize only (no global resize per mousemove). Info pane fires `**nuxeo-layout-updated`** on settle so zoom reflows even when the drawer did not move. |
| Safety          | The document content is **always guaranteed a usable minimum width** — neither side panel can squeeze the main view to nothing.                                                                                                                                                                      |
| Themes          | `--nuxeo-resize-handle-color` in default, light, dark, and kawaii themes (hover/focus on the widget). Tooltip label styles live in **nuxeo-elements** (`nuxeo-tooltip` clone), not `themes/base.js`.                                                                                                 |
| Translations    | English keys `**app.drawer.resize`**, `**documentPage.resize.side**` in **nuxeo-elements** `ui/i18n/messages.json` (Crowdin sync as usual).                                                                                                                                                          |
| Tests           | `test/nuxeo-app-drawer-resize-behavior.test.js`, `test/nuxeo-document-page.test.js`, `test/nuxeo-app.test.js` (RTL integration); **nuxeo-elements** `ui/test/nuxeo-resize-handle-extras.test.js`, `nuxeo-tooltip.test.js`.                                                                           |


No existing screens, layouts, or workflows were changed when the user has not resized anything.

---

## 3a. Charts — layout, resize, and who gets updated

### Where each handle lives (accurate nesting)

```mermaid
flowchart TB
  subgraph app ["nuxeo-app"]
    SB["Sidebar\n52 px"]
    DR["Drawer\n+ nuxeo-resize-handle\nedge=end ⇨"]
    subgraph main ["#mainContent — all routes"]
      ROUTE["search · tasks · admin · …"]
      subgraph doconly ["document route only"]
        DV[".main — document view"]
        INFO[".side — info pane\n+ nuxeo-resize-handle\nedge=start ⇦"]
      end
    end
  end
  SB --- DR --- main
  DV --- INFO
```




| Handle    | Visible when                                     | Affects                              |
| --------- | ------------------------------------------------ | ------------------------------------ |
| Drawer    | Drawer open **and** viewport **> 720 px**        | Width of **every** main-content page |
| Info pane | Info pane open **and** wide layout **≥ 1025 px** | Document page only                   |


### Three ideas to keep straight

```mermaid
flowchart TB
  subgraph perPane ["Per pane"]
    PREF["User preference\nlocalStorage"]
    PAINT["What you see now\ninline width / CSS var"]
    LEGAL["Legal range right now\nmin … max for this screen"]
  end
  PREF --> LEGAL
  LEGAL --> PAINT
```



On **zoom or window resize**, preference in storage stays; min/max change; code **re-clamps** painted width back toward preference (same pattern for drawer and info pane).

### When widths are saved


| User action                     | Saved to `localStorage`   | Layout reflow to rest of app                                        |
| ------------------------------- | ------------------------- | ------------------------------------------------------------------- |
| Drawer drag **while moving**    | No                        | `iron-resize` only (tables reflow, no global `window.resize` storm) |
| Drawer drag **release**         | Yes                       | Full settle (`notifyResize` + `window.resize`)                      |
| Drawer arrow / Home / End       | Yes (each step)           | Full settle                                                         |
| Info pane drag **while moving** | No                        | Bubbling `resize` on document-page only                             |
| Info pane drag **release**      | Yes                       | `nuxeo-layout-updated` → app full settle                            |
| Info pane arrow / Home / End    | Yes (each step)           | `nuxeo-layout-updated` → app full settle                            |
| Browser zoom / window resize    | No (preference unchanged) | Re-clamp both panes + full settle                                   |


### How main content learns the width changed

```mermaid
flowchart LR
  subgraph during ["While dragging"]
    D1["Drawer drag"]
    D2["Info drag"]
    D1 --> IR1["iron-resize only"]
    D2 --> R2["resize event\non document-page"]
  end
  subgraph settle ["On settle / zoom"]
    S1["Drawer: release,\nkeys, reclamp"]
    S2["Info: release,\nkeys, reclamp"]
    S2 --> E["nuxeo-layout-updated"]
    S1 --> N["_notifyLayoutChanged"]
    E --> N
    N --> IR2["iron-resize"]
    N --> WR["window.resize\n(guarded)"]
  end
  IR1 --> W["Tables · viewers\nin main column"]
  IR2 --> W
  WR --> W
```




| Channel                          | Who cares                                       | When                                         |
| -------------------------------- | ----------------------------------------------- | -------------------------------------------- |
| `notifyResize()` → `iron-resize` | Data tables, picture viewer, etc.               | Drawer drag move **and** every settle        |
| `window.resize`                  | App narrow flag, info-pane reclamp, legacy code | **Settle only** (not each drawer drag frame) |
| `nuxeo-layout-updated`           | `nuxeo-app` listener                            | Info pane **settle** / zoom reclamp          |


**Not page-by-page:** the shell does not special-case routes; widgets must listen via `IronResizableBehavior` or `window.resize`.

Contract: [implementation-strategy §3.4a](./ELEMENTS-1844-implementation-strategy.md#34a-nuxeo-layout-updated-event-contract).

### When both panes compete for space (push-back)

```mermaid
sequenceDiagram
  participant U as User
  participant I as Info pane
  participant A as nuxeo-app
  participant D as Drawer

  U->>I: Drag info pane past its max
  I->>A: nuxeo-shrink-drawer { amount }
  A->>D: Shrink drawer (min floor applies)
  Note over I,D: Main document column stays ≥ reserved width
```



If the drawer is already at its minimum, the info pane stops growing — no error, no layout break.

### Zoom in / out (drawer vs info pane)

```mermaid
sequenceDiagram
  participant User
  participant Browser
  participant App as nuxeo-app
  participant Doc as document-page
  participant LS as localStorage

  User->>Browser: Zoom in or out
  Browser->>App: window resize
  App->>App: Re-clamp drawer from LS preference
  App->>App: notifyResize + window.resize
  Browser->>Doc: window resize
  Doc->>Doc: Next frame: re-clamp info pane from LS preference
  Doc->>App: nuxeo-layout-updated
  App->>App: notifyResize + window.resize
  Note over LS: Saved widths unchanged unless user releases drag
```



### Responsive breakpoints (when handles appear)

```mermaid
flowchart LR
  W["Viewport width"]
  W --> A["≥ 1025 px\nBoth handles\n(side-by-side doc)"]
  W --> B["721–1024 px\nDrawer handle only\ninfo stacks below doc"]
  W --> C["≤ 720 px\nNo handles\nmobile drawer overlay"]
```




| Effective width | Drawer handle        | Info handle        | Layout                      |
| --------------- | -------------------- | ------------------ | --------------------------- |
| ≥ 1025 px       | Yes (if drawer open) | Yes (if info open) | Doc + info side by side     |
| 721–1024 px     | Yes (if drawer open) | No                 | Info below document         |
| ≤ 720 px        | No                   | No                 | Drawer overlay; stacked doc |


Browser **zoom** changes effective CSS pixels the same way as resizing the window.

### Release train (dual repo)

```mermaid
flowchart LR
  E["nuxeo-elements PR\nnuxeo-resize-handle\n+ i18n + tooltip"]
  RC["Published @nuxeo/nuxeo-ui-elements RC"]
  W["nuxeo-web-ui PR\nbehavior + hosts"]
  CI["CI: npm ci + latest RC"]
  E --> RC --> W --> CI
```



### What ships in the PRs

**nuxeo-web-ui** ([WEBUI-2040](https://github.com/nuxeo/nuxeo-web-ui/pull/3186)):

- `elements/nuxeo-app.js` — shell, layout notify, shrink-drawer listener, `nuxeo-layout-updated` consumer
- `elements/behaviors/nuxeo-app-drawer-resize-behavior.js` — drawer width math, persistence, resize handlers
- `elements/document/nuxeo-document-page.js` — info-pane math, persistence, push-back, `nuxeo-layout-updated` producer
- `themes/*/theme.html` — `--nuxeo-resize-handle-color`
- `test/nuxeo-app-drawer-resize-behavior.test.js`, `test/nuxeo-document-page.test.js`, `test/nuxeo-app.test.js`
- `docs/ELEMENTS-1844-*.md`

**nuxeo-elements** (coordinated release — `nuxeo-resize-handle`, tooltip clone fix, i18n keys):

- `ui/widgets/nuxeo-resize-handle.js`, `ui/widgets/nuxeo-tooltip.js`
- `ui/nuxeo-ui-elements.js`, `ui/i18n/messages.json`
- `ui/test/nuxeo-resize-handle-extras.test.js`, `ui/test/nuxeo-tooltip.test.js`

CI installs the latest `@nuxeo/nuxeo-ui-elements` RC from the registry — the elements PR must publish an RC that includes `nuxeo-resize-handle` before web-ui CI passes on the widget refactor.

---

## 4. How the resize works in practice

### What happens while you drag

1. The user focuses or grabs a `**nuxeo-resize-handle`** edge (mouse, touch, or Tab).
2. While dragging, width updates live; the drawer skips global `window.resize` each frame (performance).
3. On **pointer release**, width is written to `localStorage` and the app runs a **full layout settle**.
4. Each **keyboard arrow / Home / End** step saves immediately and settles layout.
5. On the next visit, panes restore from `localStorage` (clamped to the current screen).
6. On **zoom or window resize**, stored preference is kept; painted width is **re-clamped** to the new legal range (a temporary shrink while zoomed in does not overwrite what the user saved).

### What stops the user from making things unusable

Each pane has a **minimum width** (so labels and icons never get cut off) and a **maximum width** (so the rest of the page is never squashed).

- **Browse / Search drawer minimum: 350 px** — natural open width (298 px content + sidebar). Below this, navigation labels start to wrap.
- **Information pane minimum: 280 px** — safe floor below which Info / Properties / Collections cards start to clip.
- The maximum on either side is whatever leaves enough room for the main document content. (See § 5.)

If the user tries to drag past these limits, the pane simply stops moving — there is no jump, no error, no jarring snap.

### Resetting back to the default

- Double-click the edge, **or**
- Focus the edge with the keyboard and press `Enter` or `Space`.

---

## 5. How the main document content stays readable

This is the **most important behaviour** to understand.

The document view (the middle column) is the focus of the user's work. The two side panels exist to support it. So the implementation reserves space for the main column **before** the side panels are allowed to grow.

The reservation rule is: **the main column is guaranteed at least half of the document-page container's width**, capped at a comfortable 640 px on very large screens and floored at 240 px on very small ones. That gives readers enough room for data tables and document content even when both panes are maxed out.

In plain numbers — both panes dragged to their maximum at that viewport (drawer max is **dynamic**, grows when you zoom out):


| Viewport    | Drawer max | Info-pane max | **Main column ≥** |
| ----------- | ---------- | ------------- | ----------------- |
| 1920 × 1080 | 960 px     | 610 px        | **610 px**        |
| 1600 × 900  | 800 px     | 450 px        | **450 px**        |
| 1440 × 900  | 700 px     | 370 px        | **370 px**        |
| 1366 × 768  | 683 px     | 341 px        | **342 px**        |
| 1280 × 720  | 640 px     | 320 px        | **320 px**        |


If the user drags the info pane bigger and the drawer is already wide, **the drawer automatically shrinks to make room** — the main column never drops below its reserved minimum.

The reservation is proportional, so it adapts to smaller screens and to browser zoom. Below the 1024 px breakpoint the document page switches to vertical stacking (main on top, info below) and resize edges are hidden — matching the stock responsive behaviour.

---

## 6. Different screen sizes and browser zoom

The feature follows the **same responsive breakpoints** the rest of Nuxeo Web UI already uses:


| Viewport               | Behaviour                                                                                                                                                        |
| ---------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Wide (≥ 1025 px)       | Resize edges are visible and active on both panes.                                                                                                               |
| Tablet (721 – 1024 px) | Document view switches to vertical stacking (main on top, info below). Resize edges on the info pane are hidden. The drawer is still side-by-side and resizable. |
| Mobile (≤ 720 px)      | Drawer turns into a mobile overlay opened from the hamburger menu. Info pane stacks below the main view. No resize edges.                                        |


Browser zoom is handled the same way as window resizing:

- At 100 % zoom, the user gets the full feature.
- At 175 %, the document view stays visible and the panels can still be resized.
- At 250 % or more, the layout drops into mobile overlay mode automatically (matching the existing app behaviour).
- Coming back down from a high zoom level, the previously-open drawer is **restored to its saved width** automatically — including the case where the user opened the drawer while in narrow / overlay mode and then zoomed back out.
- After any zoom transition, **layout-dependent components re-flow** — picture viewers, adaptive data tables, etc. — via the app shell: `drawerPanel.notifyResize()` (iron-resize) **and** a guarded `window.resize`. This applies to **whatever page is open** in main content, not only the document view.
- Info pane width is re-applied from `localStorage` after zoom (so it matches the drawer: refresh is not required to restore a widened info pane).

---

## 7. Accessibility & keyboard support

The draggable edges are **full keyboard widgets**, not just mouse targets. They show up in the keyboard tab order and announce themselves to screen readers.


| Input              | What it does                                              |
| ------------------ | --------------------------------------------------------- |
| Tab                | Focuses the resize edge.                                  |
| `←` / `→` arrow    | Resizes by 16 px.                                         |
| `Shift` + arrow    | Resizes by 64 px (faster).                                |
| `Home`             | Jumps to minimum width.                                   |
| `End`              | Jumps to maximum width.                                   |
| `Enter` or `Space` | Resets to default width (documented in the `aria-label`). |
| Double-click       | Resets to default width.                                  |


The mouse cursor changes to a horizontal-resize arrow when hovering. Keyboard `**:focus-visible`** uses an outline in `--nuxeo-resize-handle-color` (same token as the hover fill).

**Labels:**

- Both handles use `**label-key`** on `nuxeo-resize-handle` → translated `aria-label` plus `**nuxeo-tooltip**` on hover (styles scoped on the tooltip clone, not global theme CSS).
- Keys: `app.drawer.resize` (drawer), `documentPage.resize.side` (info pane).

Example SR text: *"Resize the side navigation panel. Use Left/Right arrow keys to adjust width, Home or End for minimum or maximum, Enter or Space to reset."*

**RTL (Arabic, Hebrew):** see [§ 7a](#7a-ltr-vs-rtl-arabic-hebrew) — mirrored arrows, handle placement, and QA.

---

## 7a. LTR vs RTL (Arabic, Hebrew)

The UI follows the app / page `**dir*`* attribute. Resize behaviour is mirrored so users always widen a pane by dragging **away from the main content** and narrow it by dragging **toward** the main content. Keyboard **Left/Right** are remapped the same way (labels still say “Left/Right arrow keys” in English).

### Handle placement (inner edge toward main)

The widget uses `edge` + explicit `dir` (not CSS `:host-context` through `app-drawer` shadow DOM).


| Pane                   | `edge`  | `dir` from host                                       | Handle sits on (visual)                                            |
| ---------------------- | ------- | ----------------------------------------------------- | ------------------------------------------------------------------ |
| Browse / Search drawer | `end`   | `nuxeo-app` `_resizeHandleDir(_isRTL)`                | **LTR:** right side of drawer · **RTL:** left side of drawer       |
| Information pane       | `start` | `nuxeo-document-page` `_resizeHandleDir` (page `dir`) | **LTR:** left side of info pane · **RTL:** right side of info pane |


```text
LTR — document page (top view, simplified)

  [sidebar][  drawer  |⇨|  main + doc view  |⇦|  info ]

RTL — same structure mirrored

  [sidebar][|⇦  drawer  ][  main + doc view  |⇨|  info ]
            ↑ handle on drawer’s inner edge toward main
```

Tooltip hover position flips with direction (`_drawerResizeTooltipPosition` / `_sideResizeTooltipPosition`).

### Which arrow **widens** the pane?

Logic lives in `**nuxeo-resize-handle`** (`resizeDeltaForKey`). “Widen” means **increase pane width** in pixels.


| Pane      | Edge    | Direction | Arrow that **widens** | Arrow that **narrows** |
| --------- | ------- | --------- | --------------------- | ---------------------- |
| Drawer    | `end`   | **LTR**   | `→` Right             | `←` Left               |
| Drawer    | `end`   | **RTL**   | `←` Left              | `→` Right              |
| Info pane | `start` | **LTR**   | `←` Left              | `→` Right              |
| Info pane | `start` | **RTL**   | `→` Right             | `←` Left               |


`Home` / `End` still mean **minimum** / **maximum** width in both directions. `Shift` + arrow keeps the same widen/narrow mapping with a 64 px step.

### Pointer drag (mouse / touch)

Drag uses `resizeDeltaFromPointer`: the sign of movement is flipped in RTL so the **same physical gesture** (pull the inner edge away from the document) **widens** the pane in both LTR and RTL. QA should verify drag direction feels natural in an RTL locale, not only arrow keys.

```mermaid
flowchart LR
  subgraph widget ["nuxeo-resize-handle"]
    D["dir=ltr | rtl"]
    E["edge=start | end"]
    K["resizeDeltaForKey\nresizeDeltaFromPointer"]
  end
  subgraph hosts ["Hosts apply delta"]
    APP["nuxeo-app\n_drawerResizeStep…"]
    DOC["nuxeo-document-page\n_onSideResizeStep…"]
  end
  D --> K
  E --> K
  K --> APP
  K --> DOC
```



### What QA should check in RTL


| Check               | Pass criteria                                                                     |
| ------------------- | --------------------------------------------------------------------------------- |
| Handle visible      | Both handles on the **inner** edge (toward main), not missing inside `app-drawer` |
| Drag widen / narrow | Dragging away from the document **widens**; toward document **narrows**           |
| Arrow keys          | Widening/narrowing matches the table above                                        |
| Push-back           | Growing info pane past max still fires `nuxeo-shrink-drawer`; drawer shrinks      |
| Persistence         | `localStorage` keys unchanged; width restores after refresh                       |


Automated coverage: `test/nuxeo-app.test.js` (drawer handle), `test/nuxeo-document-page.test.js` and `test/nuxeo-app-drawer-resize-behavior.test.js` (RTL suites), **nuxeo-elements** `ui/test/nuxeo-resize-handle.test.js` (delta helpers).

---

## 8. Persistence

The chosen widths are saved in the browser's `localStorage`:


| Pane                   | Key                                |
| ---------------------- | ---------------------------------- |
| Browse / Search drawer | `nuxeo.drawerWidth`                |
| Information pane       | `nuxeo.documentPage.sidePaneWidth` |


- Per-browser, per-machine (not synced across devices — same as theme / language today).
- Restored on every page load (drawer on `ready`; info pane on `attached` via `animationFrame`).
- On zoom or window resize, reclamp reads **saved preference** again (not only the last in-memory value) — same idea as the drawer.
- If the saved value is too large for the current screen, it is clamped automatically.
- Reset clears the key and restores default flex / natural widths.
- If `localStorage` is unavailable (private mode, quota, etc.), the panes use defaults and errors are ignored.

---

## 9. Things product / QA should be aware of

1. **No new server calls.** The feature is purely client-side. No new API, no new permissions.
2. **No effect on existing layouts** when the user has not resized anything — defaults are unchanged.
3. **Already-open documents survive zooming.** If you zoom in past the mobile breakpoint, the drawer collapses; zooming back out automatically restores it at the user's saved width when the inline width still looks "open".
4. **Main column is guaranteed half the document-page container** (cap 640 px, floor 240 px). On a 1920 × 1080 monitor that's at least **610 px** even when both panes are dragged to max.
5. **Layout reflow is global** — `nuxeo-app` runs `notifyResize` on every notify path; synthetic `window.resize` only on **settle** (not each drawer drag frame). Info pane uses `nuxeo-layout-updated` to trigger the same settle path. See § 3a.
6. **Future pages** — if a screen looks wrong after resize/zoom, check whether its components use those listeners; the shell does not special-case each route.
7. **Themes:** resize-edge hover/focus uses `**--nuxeo-resize-handle-color`**, not the primary brand color.
8. **Tablet gap (721–1024 px):** drawer is resizable; info-pane handle is hidden by design (stacked layout).

### Minimum QA pass (before merge)


| #   | Steps                                                  | Expected                                                                                              |
| --- | ------------------------------------------------------ | ----------------------------------------------------------------------------------------------------- |
| 1   | 100 % zoom — drag drawer and info pane to max, refresh | Widths restored; main column still usable                                                             |
| 2   | Max info + max drawer                                  | Drawer shrinks or panes hit limits; no crushed main view                                              |
| 3   | Double-click each handle                               | Default width; `localStorage` key cleared                                                             |
| 4   | Keyboard: Tab to handle, arrows, Home, End, Enter      | Same as drag limits; reset on Enter                                                                   |
| 5   | Zoom 100 % → 150 % → 100 % with drawer open            | Drawer reclamps; tables/viewers reflow without hover                                                  |
| 6   | Zoom past 250 % then back                              | Overlay drawer; restore saved width on zoom out                                                       |
| 7   | RTL locale or `dir=rtl`                                | Per [§ 7a](#7a-ltr-vs-rtl-arabic-hebrew): handles on inner edge; arrows + drag widen/narrow correctly |
| 8   | Route: open **Search** (not document)                  | Drawer handle works; no info handle                                                                   |


---

## 10. Files to look at


| File                                                     | Repo         | Purpose                                                                                  |
| -------------------------------------------------------- | ------------ | ---------------------------------------------------------------------------------------- |
| `ui/widgets/nuxeo-resize-handle.js`                      | **elements** | Shared resize strip: input, RTL, events, ARIA helpers.                                   |
| `ui/widgets/nuxeo-tooltip.js`                            | **elements** | Tooltip clone styling for resize labels (no leak into `themes/base.js`).                 |
| `elements/behaviors/nuxeo-app-drawer-resize-behavior.js` | web-ui       | Drawer math, persistence, resize handlers, push-back consumer.                           |
| `elements/nuxeo-app.js`                                  | web-ui       | Shell, `**_notifyLayoutChanged` / `_runLayoutNotify`**, `nuxeo-layout-updated` listener. |
| `elements/document/nuxeo-document-page.js`               | web-ui       | Info-pane math, push-back producer, `**nuxeo-layout-updated**` on settle/zoom.           |
| `ui/i18n/messages.json`                                  | **elements** | `app.drawer.resize`, `documentPage.resize.side`.                                         |
| `themes/*/theme.html`                                    | web-ui       | `--nuxeo-resize-handle-color`.                                                           |
| `test/nuxeo-app-drawer-resize-behavior.test.js`          | web-ui       | Drawer unit tests (math, layout notify, RTL, shrink-drawer).                             |
| `test/nuxeo-document-page.test.js`                       | web-ui       | Info-pane unit tests (incl. RTL, `nuxeo-layout-updated`).                                |
| `test/nuxeo-app.test.js`                                 | web-ui       | App shell RTL integration on `#drawerResizeHandle`.                                      |


---

## TL;DR

Users resize the **Browse/Search drawer** (all routes) and **Information pane** (document pages) via `**nuxeo-resize-handle`**. Widths persist in `**nuxeo.drawerWidth**` and `**nuxeo.documentPage.sidePaneWidth**`. Main content keeps **≥ half** of the doc-page width (240–640 px band). **RTL:** inner-edge handles + mirrored arrows/drag (§ 7a). **While dragging:** light reflow (`iron-resize` / local `resize`). **On settle:** app runs `notifyResize` + guarded `window.resize`; info pane signals `**nuxeo-layout-updated`**. Info pane can `**nuxeo-shrink-drawer**`. Ship **elements RC then web-ui**. **WEBUI-2040** / **ELEMENTS-1844**. Diagrams: § 3a · QA: § 9.

[WEBUI-2040]: https://hyland.atlassian.net/browse/WEBUI-2040?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WEBUI-2040]: https://hyland.atlassian.net/browse/WEBUI-2040?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


elements/nuxeo-app.js
```diff
@@ -498,6 +566,20 @@ Polymer({
...
+                aria-orientation="vertical"
+                tabindex="0"
+                aria-label$="[[i18n('app.drawer.resize')]]"
+                title$="[[i18n('app.drawer.resize')]]"
+                on-mousedown="_onDrawerResizeStart"
+                on-touchstart="_onDrawerResizeStart"
+                on-keydown="_onDrawerResizeKey"
+                on-dblclick="_resetDrawerWidth"
+                hidden$="[[_drawerResizeHidden]]"
+              ></div>
```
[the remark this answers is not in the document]
naveen-konda: Thanks for flagging this — agreed that a focusable `role="separator"` should expose its range.

**Fix (upcoming commit on this branch):**
- Bound `aria-valuemin`, `aria-valuemax`, and `aria-valuenow` on the drawer `.resize-handle` via `aria-valuemin$` / `aria-valuemax$` / `aria-valuenow$`.
- Added `_updateDrawerResizeAria()` to keep them in sync with `_minDrawerWidth()`, `_maxDrawerWidth()`, and `_computeOpenDrawerWidth()` whenever the width changes (open drawer, drag, keyboard, shrink-drawer, reset, narrow layout, reclamp on window resize).
- When the handle is hidden (`!drawerOpened` or `isNarrow`), values are set to `0` so assistive tech does not announce stale bounds.

**Tests:** `test/nuxeo-app.test.js` — `_updateDrawerResizeAria sets min max and now for screen readers`.

elements/document/nuxeo-document-page.js
```diff
@@ -192,6 +279,17 @@ Polymer({
...
+          class="side-resize-handle"
+          role="separator"
+          aria-orientation="vertical"
+          tabindex="0"
+          aria-label$="[[i18n('documentPage.resize.side')]]"
+          on-mousedown="_onSideResizeStart"
+          on-touchstart="_onSideResizeStart"
+          on-keydown="_onSideResizeKey"
+          on-dblclick="_resetSideWidth"
+        ></div>
```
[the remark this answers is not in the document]
naveen-konda: Good catch — the info-pane handle should announce position and bounds the same way as the drawer splitter.

**Fix (upcoming commit on this branch):**
- Bound `aria-valuemin`, `aria-valuemax`, and `aria-valuenow` on `.side-resize-handle`.
- Added `_updateSideResizeAria()` (min/max from `_minSideWidth()` / `_maxSideWidth()`, now from `sideWidth` or measured `.side` width) and call it from `_sideWidthChanged`, drag/keyboard handlers, open/close, attach/reclamp paths.
- Hidden/narrow: same `0` reset pattern as the drawer handle.

**Related cleanup (review summary):** `_sideWidthChanged` no longer manually sets/removes `side-width`; `reflectToAttribute: true` on `sideWidth` is sufficient — we only update `--nuxeo-side-pane-width` and ARIA there now.

**Tests:** `test/nuxeo-document-page.test.js` — `_updateSideResizeAria sets min max and now for screen readers`, plus `sideWidth null clears side-width via reflectToAttribute only`.

test/nuxeo-diff.test.js
```diff
@@ -53,13 +53,6 @@ suite('nuxeo-diff', () => {
     test('should default _hasVersions to false', () => {
       expect(element._hasVersions).to.be.false;
     });
-
-    test('should have headers with fetch-document and translate-directoryEntry', () => {
-      expect(element.headers).to.deep.equal({
-        'fetch-document': 'properties',
-        'translate-directoryEntry': 'label',
-      });
-    });
```
madhurkulshrestha-hyland, line 62: Was this related to the WEBUI-2040 changes ?

elements/document/nuxeo-document-page.js
madhurkulshrestha-hyland, line 1: **B2 — `nuxeo-shrink-drawer` must be `bubbles: true, composed: true`**

The push-back contract requires the event fired from `nuxeo-document-page` to reach `nuxeo-app` across ShadowDOM. The current test (`_onSideResizeStart dispatches nuxeo-shrink-drawer when growing past max`) only listens on `element` itself, which would pass even if `composed: false`.

Please:
1. Confirm the dispatch options are `{ bubbles: true, composed: true, detail: { amount } }`.
2. Add an assertion that the event reaches a parent host — e.g. mount the fixture inside a wrapper element and attach the listener there.

Without this, a future internal template refactor inside `nuxeo-document-page` could silently break drawer push-back with no failing test.

madhurkulshrestha-hyland, line 1: **B3 — `localStorage.setItem` failure path is unhandled in tests**

`_loadStoredSideWidth` is covered for the "throws" case (`_loadStoredSideWidth returns null when localStorage throws`), but `_persistSideWidth` is not. In Safari Private mode and quota-full scenarios `setItem` throws — if the call inside the drag/keyboard path isn't wrapped, the throw happens mid-mousemove and leaves the element stuck with the `side-resizing` attribute and window listeners attached.

Please add a test where `setItem` throws and assert:
1. The drag completes cleanly.
2. The `side-resizing` attribute is removed.
3. The temporary `mousemove`/`mouseup` window listeners are detached.
4. No uncaught error escapes.

Same review applies to the drawer side in `elements/nuxeo-app.js`.

elements/nuxeo-app.js
madhurkulshrestha-hyland, line 1: **S1 — Extract drawer resize into a behavior/mixin**

`nuxeo-app.js` already carries shell routing, narrow-mode, drawer state, search, notifications, etc. Adding ~445 lines of resize logic in-place makes the file hard to own and hard to review on subsequent changes.

Suggest extracting into `behaviors/nuxeo-drawer-resize-behavior.js` (matching the existing `FiltersBehavior` / `RoutingBehavior` / `I18nBehavior` pattern in this codebase). `nuxeo-app` then just applies the behavior. No functional change, much easier review next time and unit tests can target the behavior in isolation.

madhurkulshrestha-hyland, line 1: **S2 — Promote the resize handle to a reusable element**

The 6 px strip, hover / `:focus-visible` outline, ARIA pattern, keyboard map (←/→, Shift+←/→, Home/End, Enter/Space) and RTL mirroring are duplicated between `elements/nuxeo-app.js` (this file) and `elements/document/nuxeo-document-page.js`.

Suggest a small `<nuxeo-resize-handle>` element in `nuxeo-elements/ui/`. Both hosts then own only their clamp math, persistence and push-back coordination. The two new i18n keys (`app.drawer.resize`, `documentPage.resize.side`) would move to `ui/i18n/messages.json` so they ship with the widget and stay in sync across consumers.

Worth doing now while there are only two call sites — once a third one lands the duplication will be much harder to unwind.

themes/base.js
```diff
@@ -614,6 +614,15 @@ const template = html`
           --nuxeo-document-blob-actions-menu-max-width: 80px;
         }
       }
+
+      /* nuxeo-tooltip clones slot content into paper-tooltip on document.body */
+      .resize-handle-tooltip-label {
```
madhurkulshrestha-hyland: **S3 — Component-specific style leaking into the global theme**

`.resize-handle-tooltip-label` is here only because `nuxeo-tooltip` clones slot content into `document.body`. That's a component-specific rule sitting in a shared global `<custom-style>` — exactly the kind of leak `themes/base.js` is supposed to avoid, and a future maintainer has no signal that deleting it will break the resize handle.

Please either:
1. Move the rule into `nuxeo-tooltip` itself (where the body-clone behaviour can be documented), or
2. At minimum scope the selector with an explicit role — e.g. `[data-nx-tooltip-role="resize-handle"] .resize-handle-tooltip-label` — and add the matching `data-nx-tooltip-role` attribute on the handle's tooltip. That way the cross-boundary intent is visible from both ends.

elements/nuxeo-app.js
madhurkulshrestha-hyland, line 1: **S4 — Drop the synthesised `window.resize` dispatch**

The reflow path in `_notifyLayoutChanged` / `_runLayoutNotify` fires both `drawerPanel.notifyResize()` (iron-resize) **and** a synthetic `window.resize`. Synthesising a window resize event triggers every listener on the page — Polymer media queries, third-party widgets, analytics, lazy-loaders, etc. — even when the only thing that changed is one pane's width.

`iron-resize` plus the new `nuxeo-layout-updated` event should already cover the legitimate consumers. Please audit who actually needs `window.resize` and migrate them to `iron-resize` / `nuxeo-layout-updated`. Either remove the synth or keep it behind an explicit allow-list. Otherwise we're paying a global cost every time the user touches the handle (or holds an arrow key down).

test/nuxeo-document-page.test.js
madhurkulshrestha-hyland, line 1: **S5 — RTL coverage missing**

The PR description claims the handles and arrow-key directions mirror in RTL (Arabic, Hebrew), but every test in the `side pane resize` suite sets `element.setAttribute('dir', 'ltr')`. There's no coverage for the RTL paths.

Please add at least:
1. `ArrowLeft` / `ArrowRight` inversion in `dir="rtl"` for the side pane (and the equivalent for the drawer in `nuxeo-app.test.js`).
2. Drag direction inversion in RTL (`_onSideResizeStart` with `clientX` moving in the opposite direction should still grow/shrink correctly).
3. Push-back direction in RTL — `nuxeo-shrink-drawer` should still fire when the side pane is grown past its max regardless of `dir`.

Without these, the RTL claim is unverified and easy to regress.

madhurkulshrestha-hyland, line 1: **S6 — `window.matchMedia` stub shape is incomplete**

The fixtures stub `matchMedia` with `{ matches }` only:

```js
const matchMediaStub = sinon.stub(window, 'matchMedia').returns({ matches: false });
```

Polymer's `<iron-media-query>` and several Nuxeo helpers call `addEventListener('change', …)` / `addListener(…)` / read `media`. If the implementation (or a transitive dep) registers a breakpoint listener, this stub will throw on the first call.

Please return a more complete stub:

```js
sinon.stub(window, 'matchMedia').returns({
  matches: false,
  media: '',
  onchange: null,
  addListener: () => {},
  removeListener: () => {},
  addEventListener: () => {},
  removeEventListener: () => {},
  dispatchEvent: () => false,
});
```

A small `createMatchMediaStub(matches)` helper would also avoid copy-paste across the suites.

test/nuxeo-app.test.js
madhurkulshrestha-hyland, line 1: **N2 — Test file size / organisation**

This file grows by ~1400 lines in a single PR, all for the drawer resize concern. The existing suites in here cover routing, narrow-mode, notifications, etc., and the resize block is now larger than all of them combined.

Suggest splitting the resize concern into a sibling file, e.g. `test/nuxeo-app.resize.test.js`. Karma's glob picks it up automatically, and both files become easier to scan and to triage on failure. Pairs naturally with **S1** (extract a behavior) — the new file can then test the behavior directly.

elements/document/nuxeo-document-page.js
madhurkulshrestha-hyland, line 1: **N7 — Document the `nuxeo-layout-updated` event contract**

`nuxeo-layout-updated` is a new public-ish event in the app shell layer: it bubbles from `nuxeo-document-page` and is observed by `nuxeo-app`, and it's the recommended replacement for `window.resize` listeners (see **S4**). Future consumers (custom layouts, third-party extensions) will need to know:

1. Who dispatches it (sources: drawer reclamp, side-pane reclamp, zoom reclamp, …).
2. The payload shape (`detail.*` — what's guaranteed, what's best-effort).
3. Where to listen (the originating element? `nuxeo-app`? `document`?).
4. Whether it's `composed`/`bubbles` (so it can be observed from outside the shadow root).
5. Cadence / throttling guarantees, if any.

Please add a short section to `docs/ELEMENTS-1844-implementation-strategy.md` (and ideally a JSDoc `@event` block on the dispatching element). Without this the event becomes "stable by accident" and is easy to break in a follow-up.

madhurkulshrestha-hyland (changes requested):

test/nuxeo-diff.test.js
naveen-konda: No — this removal is **not** part of WEBUI-2040.

It was an unrelated cleanup in an earlier commit on this branch (`address copilot review comments`): the `headers` assertion duplicated coverage already exercised elsewhere in the diff suite and was dropped to reduce noise. The resize feature does not touch `nuxeo-diff` production code.

elements/document/nuxeo-document-page.js
naveen-konda: Addressed in latest commit **b031d1a4228cd1a711ecdf174a11a56a8dc96e9a** `elements/document/nuxeo-document-page.js` and `test/nuxeo-document-page.test.js`.

1. **Dispatch** — `_applySideWidth` fires `nuxeo-shrink-drawer` with `{ bubbles: true, composed: true, detail: { amount } }` (see `_applySideWidth`).
2. **Tests** — Added assertions that the event reaches a **parent host** outside the element shadow tree:
   - `nuxeo-shrink-drawer from drag reaches a shadow host when composed` (listener on wrapper, `composed`/`bubbles` checked)
   - Step/drag variants also assert `composed: true` on the captured event
   - RTL suite repeats push-back coverage under `dir="rtl"`

This would fail if a refactor set `composed: false` or stopped bubbling.

naveen-konda: Addressed on both sides. Changes available in latest commit **b031d1a4228cd1a711ecdf174a11a56a8dc96e9a**

**Production:** `_persistSideWidth` / `_persistDrawerWidth` (and load/remove paths) are wrapped in `try/catch` so `setItem`/`removeItem` failures (private mode, quota) do not propagate. Drag/keyboard handlers always run handle teardown (`resize-drag-end`, `side-resizing` / `drawer-resizing` removal, global listener detach) in the widget/behavior **before** persistence is attempted on settle.

**Tests:**
- `test/nuxeo-document-page.test.js` — `_persistSideWidth does not throw when setItem throws` (+ `removeItem`)
- `test/nuxeo-app-drawer-resize-behavior.test.js` — `_persistDrawerWidth does not throw when setItem throws`

elements/nuxeo-app.js
naveen-konda: Addressed in latest commit **b031d1a4228cd1a711ecdf174a11a56a8dc96e9a** — drawer resize logic is now in `elements/behaviors/nuxeo-app-drawer-resize-behavior.js` (`NuxeoAppDrawerResizeBehavior`), applied from `nuxeo-app` alongside the existing behaviors.

`nuxeo-app.js` keeps shell wiring (template handle, `_runLayoutNotify` / `_notifyLayoutChanged`, `nuxeo-layout-updated` listener, narrow-mode integration). Clamp math, persistence, keyboard/drag handlers, and `nuxeo-shrink-drawer` handling live in the behavior.

**Tests:** `test/nuxeo-app-drawer-resize-behavior.test.js` targets the behavior in isolation (min/max, persistence, RTL, layout notify, reclamp, shrink-drawer).

naveen-konda: Addressed in latest commit **b031d1a4228cd1a711ecdf174a11a56a8dc96e9a****nuxeo-elements** 

- New `ui/widgets/nuxeo-resize-handle.js` — 6px strip, hover/focus-visible, ARIA (`role="separator"`, value min/max/now), keyboard map, RTL-aware deltas, tooltip via moving anchor + `nuxeo-tooltip`.
- Exported helpers `resizeDeltaForKey` / `resizeDeltaFromPointer` + unit tests in `ui/test/nuxeo-resize-handle.test.js`.
- i18n keys also in `ui/i18n/messages.json`; **web-ui** keeps `i18n/messages.json` entries for runtime (`app.drawer.resize`, `documentPage.resize.side`).

Hosts (`nuxeo-app`, `nuxeo-document-page`) now only own clamp math, persistence, push-back, and layout coordination.

themes/base.js
naveen-konda: Addressed — rule removed from `themes/base.js`.

Tooltip slot content is cloned to `document.body`, so styles now live in **`nuxeo-elements/ui/widgets/nuxeo-tooltip.js`** (`ensureClonedContentStyles()`), scoped as:

`[data-nx-tooltip-role="resize-handle"].resize-handle-tooltip-label { … }`

`<nuxeo-resize-handle>` sets `data-nx-tooltip-role="resize-handle"` on its `<nuxeo-tooltip>`, which is stamped onto the cloned node. Intent is visible from both the widget and the tooltip module.

elements/nuxeo-app.js
naveen-konda: **Partially addressed**  **b031d1a4228cd1a711ecdf174a11a56a8dc96e9a** — we audited and split the reflow path instead of removing synthetic `window.resize` entirely.

### What we changed (addresses the “global storm on every mousemove” concern)
- **Drawer drag move** → `_scheduleDrawerDragLayoutNotify()` → `_runLayoutNotify({ includeWindowResize: false })` — **`drawerPanel.notifyResize()` / iron-resize only**, rAF-coalesced (≤1/frame). No synthetic `window.resize` while the pointer moves.
- **Drawer settle** (drag end, keyboard step, reset, reclamp, narrow↔wide) → `_notifyLayoutChanged()` → `_runLayoutNotify({ includeWindowResize: true })` (default).
- **Info pane drag move** → bubbling `resize` on `nuxeo-document-page` only; **no** `nuxeo-layout-updated` per frame.
- **Info pane settle / zoom reclamp** → `nuxeo-layout-updated` → app `_notifyLayoutChanged()` (same settle path as drawer).

### Why we still need synthetic `window.resize` on settle
Removing it completely broke **zoom in/out** : picture viewers, data tables, and `nuxeo-document-page`’s own `_scheduleViewportReclamp()` (listens on **real** `window` `resize` from browser zoom) stopped reflowing reliably when only iron-resize ran. Many in-tree and third-party listeners still subscribe to `window.resize` / narrow breakpoints; migrating all of them is out of scope for WEBUI-2040.

So settle keeps **both** channels, with guards:
1. `notifyResize()` on `#drawerPanel` (iron-resize chain for main content).
2. One guarded `globalThis.dispatchEvent(new Event('resize'))` with `_suppressLayoutResizeHandler` so app narrow-mode does not loop.

### Tests / docs
- `test/nuxeo-app-drawer-resize-behavior.test.js` — `_scheduleDrawerDragLayoutNotify runs iron-resize without synthetic window.resize`

test/nuxeo-document-page.test.js
naveen-konda: Addressed in latest commit **b031d1a4228cd1a711ecdf174a11a56a8dc96e9a**`test/nuxeo-document-page.test.js` and `test/nuxeo-app-drawer-resize-behavior.test.js` / `test/nuxeo-app.test.js`.

**Info pane (`RTL` suite):**
- `resizeDeltaForKey` mirrors ArrowLeft/ArrowRight for `edge="start"` + `dir="rtl"`
- ArrowRight/ArrowLeft on handle change width in the expected direction
- Drag grow/shrink with opposite `clientX` motion
- `nuxeo-shrink-drawer` when growing past max (host listener)

**Drawer:**
- `test/nuxeo-app-drawer-resize-behavior.test.js` — RTL delta + handle keyboard tests
- `test/nuxeo-app.test.js` — `drawer resize handle (RTL)` suite

Widget RTL math is also covered in `nuxeo-elements/ui/test/nuxeo-resize-handle.test.js` via exported `resizeDeltaForKey` / `resizeDeltaFromPointer`.

naveen-konda: Addressed in latest commit **b031d1a4228cd1a711ecdf174a11a56a8dc96e9a**— stubs now return a full MediaQueryList-shaped object.

In `test/nuxeo-document-page.test.js` we added a local `createMatchMediaStub(matches)` (used by the wide/narrow `attached` tests) that stubs `window.matchMedia` with `matches`, `media`, `addListener` / `removeListener`, `addEventListener` / `removeEventListener`, and `dispatchEvent` so `<iron-media-query>` / breakpoint helpers do not throw.

Kept it inline in that file (only consumer today) rather than a separate helper module.

test/nuxeo-app.test.js
naveen-konda: Addressed (**b031d1a4228cd1a711ecdf174a11a56a8dc96e9a**) via **S1** split — drawer resize tests live in `test/nuxeo-app-drawer-resize-behavior.test.js` (Karma picks up `*.test.js` automatically).

That file covers min/max, persistence, keyboard, drag, RTL, `_scheduleDrawerDragLayoutNotify` vs `_notifyLayoutChanged`, narrow→wide reclamp, and `nuxeo-shrink-drawer`. `test/nuxeo-app.test.js` keeps a smaller RTL handle smoke suite; the large resize block is no longer mixed with routing/narrow-mode suites.

Naming is `nuxeo-app-drawer-resize-behavior.test.js` rather than `nuxeo-app.resize.test.js` to match the extracted behavior module.

elements/document/nuxeo-document-page.js
naveen-konda: Addressed in latest commit **b031d1a4228cd1a711ecdf174a11a56a8dc96e9a**

**Docs:** `docs/ELEMENTS-1844-implementation-strategy.md` — **§3.4a `nuxeo-layout-updated` event contract`** (dispatcher, sources, empty `detail`, bubbles/composed, where to listen, cadence/coalescing, relationship to S4).

**JSDoc:** `elements/document/nuxeo-document-page.js` — `@event nuxeo-layout-updated` on the element plus `@fires` on `_scheduleViewportReclamp`, `_notifySideLayoutUpdated`, and `_resetSideWidth`.

**Consumer:** `nuxeo-app` `ready()` listener documented with pointer to §3.4a.

Note: drawer reclamp does **not** dispatch this event (app calls `_notifyLayoutChanged()` directly); that is spelled out in §3.4a to avoid confusion.

madhurkulshrestha-hyland (approved):

swarnadipa-dev (approved):

PR #3277, opened 2026-07-03
WEBUI-1857: redirect anonymous user to login on a private-document 403 [lts-2025]

narasimhahyland:
## Problem
With anonymous authentication enabled, an anonymous user who opens a permalink to a document they are not allowed to read gets a dead-end "403 – Permission Denied" page. The only way forward is to manually log out of the anonymous session, log back in with a real account, and reopen the permalink. Jira: [WEBUI-1857](https://hyland.atlassian.net/browse/WEBUI-1857)

## Root cause
`nuxeo-app`'s document `load()` sends every failure (including a 403 for an anonymous user) to `showError`, rendering the permission-error page. There was no handling to recognise that an *anonymous* 403 is recoverable by authenticating with a real account.

Redirecting an anonymous user straight to `login.jsp` does not work either: the anonymous session already carries `USERIDENT_KEY`, so Nuxeo treats them as logged in and bounces back to the requested page.

## Changes
* **`elements/behaviors/nuxeo-anonymous-behavior.js`** (new): `NuxeoAnonymousBehavior` detects an anonymous 403 and routes the user through the Nuxeo **logout** endpoint, which invalidates the anonymous session and — because the user was anonymous — forwards to the login form (`forceAnonymousLogin=true`) while preserving `requestedUrl`. The Web UI route lives in the URL fragment (not sent to the server), so it is re-appended to the logout URL and stored in the `nuxeo.start.url.fragment` cookie (the same cookie `nuxeo-connection` uses for its form-auth 401 redirect); Nuxeo restores it after login so the user lands back on the requested document.
* **`elements/nuxeo-app.js`**: compose the new behavior and, in the `load()` catch, redirect on an anonymous 403 instead of showing the dead-end error page. Authenticated users still get the normal error page.
* **`test/nuxeo-anonymous-behavior.test.js`** (new): unit tests for the anonymity/403 detection, the logout redirect (URL, requestedUrl, fragment, one-shot guard) and the `load()` integration (anonymous 403 redirects; authenticated 403 still shows the error page).

## Test plan
- [x] `npm run lint` passes
- [x] `npm test` passes (2256 passed, incl. new behavior tests)
- [x] Manual/E2E against a Nuxeo instance with anonymous auth: anonymous opens a restricted permalink → redirected to the login page (`requestedUrl` preserved) → after logging in with a real account, lands back on the exact document. Before/after screenshots and videos captured.

## Notes
Self-contained client logic is kept in a Polymer behavior rather than inlined into the already-large `nuxeo-app.js`. Backport to `maintenance-3.1.x` is opened as a separate PR.

[WEBUI-1857]: https://hyland.atlassian.net/browse/WEBUI-1857?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/elements/behaviors/nuxeo-anonymous-behavior.js
+++ b/elements/behaviors/nuxeo-anonymous-behavior.js
@@ -1,0 +1,93 @@
+/**
+@license
+©2023 Hyland Software, Inc. and its affiliates. All rights reserved.
+All Hyland product names are registered or unregistered trademarks of Hyland Software, Inc. or its affiliates.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/**
+ * Anonymous authentication flow (WEBUI-1857).
+ *
+ * When anonymous authentication is enabled and an anonymous user opens a permalink to a document they
+ * are not allowed to read, the document load fails with a 403. Without this behavior the app shows a
+ * dead-end "Permission Denied" error page, and the only way forward is to manually log out of the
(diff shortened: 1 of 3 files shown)


narasimhahyland:
Recreating as a same-repo PR (CI checks out the branch by name from the upstream repo; fork branches aren't found). Superseded shortly.

PR #3283, opened 2026-07-06
WEBUI-2111: Fix shell-quote critical CVE-2026-9277 [maintenance-3.1.x]

narasimhahyland:
## Problem
Dependabot flagged a **critical** command-injection advisory (CVSS 8.1 / 9.2) in the transitive dev dependency `shell-quote@1.8.3`. `quote()` did not escape line terminators (`\n`, `\r`, U+2028, U+2029) in object `.op` values, so a newline could be treated by POSIX shells as a command separator, enabling shell command injection. Jira: [WEBUI-2111](https://hyland.atlassian.net/browse/WEBUI-2111) · [Dependabot #319](https://github.com/nuxeo/nuxeo-web-ui/security/dependabot/319) · GHSA-w7jw-789q-3m8p · CVE-2026-9277

## Root cause
`shell-quote` versions `>= 1.1.0, <= 1.8.3` are vulnerable; the fix landed in `1.8.4`. It is a transitive dev dependency pulled in via `launch-editor` (`^1.8.3`) and `npm-run-all` (`^1.6.1`).

## Changes
* **`package-lock.json`**: bump `shell-quote` `1.8.3` → `1.9.0` (patched, `>= 1.8.4`). Both consumer ranges (`^1.8.3`, `^1.6.1`) already accept `1.9.0`, so `package.json` is unchanged and no other tree entries move — surgical lockfile-only diff.

## Test plan
- [ ] `npm run lint` passes (CI) — no source changed
- [ ] `npm test` passes (CI) — no source changed
- [x] Lockfile is valid JSON and internally consistent; integrity hash matches the npm registry for `shell-quote@1.9.0`

## Notes
Lockfile-only, dev-scope transitive dependency; no runtime/app behavior change. Backported as a matched pair — companion PR targets `lts-2025`.

[WEBUI-2111]: https://hyland.atlassian.net/browse/WEBUI-2111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package-lock.json
```diff
@@ -23647,9 +23647,9 @@
       }
     },
     "node_modules/shell-quote": {
-      "version": "1.8.3",
-      "resolved": "https://registry.npmjs.org/shell-quote/-/shell-quote-1.8.3.tgz",
-      "integrity": "sha512-ObmnIF4hXNg1BqhnHmgbDETF8dLPCggZWBjkQfhZpbszZnYur5DUljTcCHii5LC3J5E0yeO/1LIMyH+UvHQgyw==",
+      "version": "1.9.0",
```
madhurkulshrestha-hyland: Optional (non-blocking) hardening idea: this lockfile-only bump is correct and sufficient — `1.9.0` is `> 1.8.4`, both consumer ranges (`launch-editor@^1.8.3`, `npm-run-all@^1.6.1`) accept it, and `shell-quote` has no deps so no other tree entries move. Since the lockfile pins `1.9.0`, `npm ci` won't regress it.

If we want defense-in-depth against a future `npm update`/manual re-resolution silently pulling an older-but-in-range version, we could add an `overrides` entry in `package.json` (consistent with the existing overrides like `ws`, `cross-spawn`, `qs`):

```jsonc
"overrides": {
  "shell-quote": ">=1.8.4"
}
```

This is hardening, not a requirement — with Dependabot watching this dependency.

PR #3285, opened 2026-07-06
WEBUI-2113: Fix form-data high CVE-2026-12143 [maintenance-3.1.x]

narasimhahyland:
## Problem
Dependabot flagged a **high**-severity CRLF-injection advisory (CVSS 7.5 / 8.7) in the transitive runtime dependency `form-data@4.0.5` in the `nuxeo-web-ui-ftest` sub-package. `form-data` concatenated `field` name and `filename` directly into the `Content-Disposition` header without escaping CR (`\r`), LF (`\n`), or `"`, so untrusted field names/filenames could inject headers or smuggle extra multipart parts. Jira: [WEBUI-2113](https://hyland.atlassian.net/browse/WEBUI-2113) · [Dependabot #334](https://github.com/nuxeo/nuxeo-web-ui/security/dependabot/334) · GHSA-hmw2-7cc7-3qxx · CVE-2026-12143

## Root cause
`form-data` `>= 4.0.0, < 4.0.6` is vulnerable (CWE-93); the fix (escaping CR/LF/`"` per the WHATWG multipart algorithm) landed in `4.0.6`. It is a transitive dependency pulled in via `nuxeo` (`^4.0.0`).

## Changes
* **`packages/nuxeo-web-ui-ftest/package-lock.json`**: bump `form-data` `4.0.5` → `4.0.6` (patched). The consumer range (`^4.0.0`) already accepts `4.0.6`, so `package.json` is unchanged and no other tree entries move — surgical lockfile-only diff. Integrity hash verified against the npm registry.

## Test plan
- [ ] CI (`lint`, `unit-test`, `ftests`, `build`) passes — no source changed
- [x] Sub-package lockfile is valid JSON; `form-data@4.0.6` integrity matches the npm registry

## Notes
Lockfile-only change scoped to the functional-test sub-package (`packages/nuxeo-web-ui-ftest`), which uses `npm ci` in CI. Backported as a matched pair — companion PR targets `lts-2025`.

[WEBUI-2113]: https://hyland.atlassian.net/browse/WEBUI-2113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/nuxeo-web-ui-ftest/package-lock.json
+++ b/packages/nuxeo-web-ui-ftest/package-lock.json
@@ -6011,9 +6011,9 @@
       }
     },
     "node_modules/form-data": {
-      "version": "4.0.5",
-      "resolved": "https://registry.npmjs.org/form-data/-/form-data-4.0.5.tgz",
-      "integrity": "sha512-8RipRLol37bNs2bhoV67fiTEvdTrbMUYcFTiy3+wuuOnUog2QBHCZWXDRijWQfAkhBj2Uf5UnVaiWwA5vdd82w==",
+      "version": "4.0.6",
+      "resolved": "https://registry.npmjs.org/form-data/-/form-data-4.0.6.tgz",
+      "integrity": "sha512-vKatAh4SlVfgbv+YtmhiRjhEMJsYpsG1Y2rMQtR+SVSbytsSD1YGzDIcrAJmdFec88u/+VoGmxnl+80gL1tRCQ==",
       "license": "MIT",
       "dependencies": {
         "asynckit": "^0.4.0",


packages/nuxeo-web-ui-ftest/package-lock.json
```diff
@@ -6011,9 +6011,9 @@
       }
     },
     "node_modules/form-data": {
-      "version": "4.0.5",
-      "resolved": "https://registry.npmjs.org/form-data/-/form-data-4.0.5.tgz",
-      "integrity": "sha512-8RipRLol37bNs2bhoV67fiTEvdTrbMUYcFTiy3+wuuOnUog2QBHCZWXDRijWQfAkhBj2Uf5UnVaiWwA5vdd82w==",
+      "version": "4.0.6",
```
madhurkulshrestha-hyland, line 6014: Optional (non-blocking) hardening idea: this lockfile-only bump is correct and sufficient — `4.0.6` is the patched release, the consumer range (`nuxeo` → `form-data@^4.0.0`) accepts it, and no other tree entries move. Since the lockfile pins `4.0.6`, `npm ci` won't regress it.

If we want defense-in-depth against a future `npm update`/manual re-resolution silently pulling an older-but-in-range version, we could add an `overrides` entry in this sub-package's `packages/nuxeo-web-ui-ftest/package.json` (consistent with the existing overrides there — `ws`, `tar-fs`, `serialize-javascript`, `tmp`, `uuid`):

```jsonc
"overrides": {
  "form-data": ">=4.0.6"
}
```

Note the override belongs in the ftest sub-package's `package.json` (this lockfile's scope), not the root app one. This is hardening, not a requirement — with Dependabot watching this dependency

madhurkulshrestha-hyland: Follow-up / clarification: there *is* an existing `form-data` override in the repo, but it's in the **root** `package.json` — `"overrides": { "form-data": "^2.5.4", ... }` — not in this ftest sub-package. Two reasons it doesn't change the picture here:

1. **Different major line, not the vulnerable range.** CVE-2026-12143 affects `form-data >= 4.0.0, < 4.0.6`. The root pin is `^2.5.4` (the v2 line, `>=2.5.4 <3.0.0`), which is outside the advisory range — so unlike the `ws` case (#3287, where the override pinned the *exact vulnerable* version), this override does **not** force a vulnerable build.
2. **Different scope.** The root override only governs the root app's install; it has no effect on this `packages/nuxeo-web-ui-ftest` tree, which is installed separately and resolves `form-data` via `nuxeo@^4.0.0`. The ftest sub-package has **no** `form-data` override, so the lockfile bump to `4.0.6` is consistent and won't be reverted.

So this PR remains correct as-is, and the optional hardening pin (if added) still belongs in `packages/nuxeo-web-ui-ftest/package.json`, not root. No change needed to the root `^2.5.4` entry for this CVE.

madhurkulshrestha-hyland (approved):

PR #3289, opened 2026-07-06
ELEMENTS-1616: request grid/list thumbnails with crossorigin [lts-2025]

narasimhahyland:
## Problem
Under S3 direct download in Chrome, document thumbnails fail to load with a CORS error when switching between grid and list views. Jira: [ELEMENTS-1616](https://hyland.atlassian.net/browse/ELEMENTS-1616)

## Root cause
`nuxeo-document-grid-thumbnail` and `nuxeo-document-list-item` render the thumbnail URL with a plain `` (no `crossorigin`). Chrome caches that cross-origin response **without** CORS headers, so the `` in `nuxeo-document-thumbnail` (nuxeo-elements) reuses the poisoned cache entry and is blocked by the CORS check → broken thumbnail.

## Changes
* **`elements/nuxeo-data-grid/nuxeo-document-grid-thumbnail.js`**: render the thumbnail with `crossorigin="anonymous"`.
* **`elements/nuxeo-data-list/nuxeo-document-list-item.js`**: render the thumbnail with `crossorigin="anonymous"`.
* **tests**: assert both elements render ``.

Makes every element that renders the same thumbnail URL consistent, so the first request always caches with CORS headers. Paired with the `nuxeo-elements` PR (nuxeo/nuxeo-elements#1498 / #1499). Safe for same-origin (default) deployments.

## Test plan
- [x] `npm run lint` passes
- [x] `npm test` passes (2245 unit tests, incl. new crossorigin assertions)
- [ ] Manual: S3 direct download, grid → list view thumbnails load in Chrome

## Notes
Requires bucket CORS for cross-origin (S3) setups (already required for direct download). Paired PRs in `nuxeo-elements`. Backport pair: `lts-2025` + `maintenance-3.1.x` (Jira fixVersions also list `3.0.x` — flagged for a possible `maintenance-3.0.x` backport).

[ELEMENTS-1616]: https://hyland.atlassian.net/browse/ELEMENTS-1616?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


madhurkulshrestha-hyland (dismissed):

AnilKumarVanga (dismissed):

test/nuxeo-document-list-item.test.js
```diff
@@ -44,6 +44,14 @@ suite('nuxeo-document-list-item', () => {
...
       expect(element.selectedItems).to.deep.equal([]);
     });
+
+    // ELEMENTS-1616: the list-item thumbnail must request the same URL with CORS
+    // so it stays consistent with the grid/preview <img> and avoids Chrome's
+    // cross-origin image cache poisoning under S3 direct download.
+    test('requests its thumbnail with CORS enabled', () => {
+      const thumbnail = element.shadowRoot.querySelector('.thumbnailContainer img');
+      expect(thumbnail.crossOrigin).to.equal('anonymous');
+    });
```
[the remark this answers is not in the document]
narasimhahyland: Addressed in 34ebcf178: assert the thumbnail `` exists before reading its attribute, and check the rendered `crossorigin` value directly for consistency with the grid-thumbnail test (no TypeError if the selector changes). Same change on the maintenance-3.1.x sibling #3290 (72d14b4b5).

madhurkulshrestha-hyland (dismissed):

elements/nuxeo-data-list/nuxeo-document-list-item.js
```diff
@@ -226,7 +226,7 @@ Polymer({
     <div class="listBox grid-box" selection-mode$="[[selectionMode]]">
       <div class="horizontal layout">
         <div class="vignette thumbnailContainer" on-tap="handleClick" on-keydown="_handleKeydown">
-          <img src="[[_thumbnail(doc)]]" alt$="[[doc.title]]" />
+          <img crossorigin="anonymous" src="[[_thumbnail(doc)]]" alt$="[[doc.title]]" />
```
vivekchaturved013: Three things to fix or acknowledge:

It's incomplete. Two other components render the byte-identical thumbnail URL through iron-image, which has no crossorigin support: nuxeo-justified-grid-item in nuxeo-elements (used by the assets search and picturebook views) and nuxeo-document-viewer in web-ui. Whichever request lands first owns the cache entry, so the original bug still reproduces going from picturebook or the viewer into grid/list.

The merge order is backwards. On today's lts-2025, nothing requests thumbnails with CORS — nuxeo-document-thumbnail only gains the attribute in the still-open [nuxeo-elements#1498](https://github.com/nuxeo/nuxeo-elements/pull/1498). So merging this first delivers no benefit and breaks thumbnails for any S3 direct-download deployment without bucket CORS. That state already exists on maintenance-3.1.x, where web-ui #3290 is merged and elements #1499 isn't.

No graceful failure. Unlike nuxeo-document-thumbnail, the grid and list  have no on-error handler, so a CORS failure now shows a broken-image icon instead of the transparent-pixel fallback.

Same-origin deployments are genuinely safe (anonymous still sends cookies same-origin), and there are no i18n, a11y, or scope concerns. I also flagged one open question: whether lts-2023 needs this backport, since no PR exists there

narasimhahyland: Thanks @vivekchaturved013 — good catches. Status per point:

**3) No graceful failure — fixed.** Added an `on-error` transparent-pixel fallback to both the grid and list `` (mirroring `nuxeo-document-thumbnail`) in this PR, and a maintenance-3.1.x follow-up in #3416 (since #3290 is already merged). A CORS/misconfig failure now degrades to a transparent pixel instead of a broken-image icon.

**2) Merge order — agreed.** On lts-2025 this PR shouldn't merge before elements nuxeo/nuxeo-elements#1498; sequencing them together avoids the "CORS request against a non-CORS server" window. Worth flagging that this window is already open on maintenance-3.1.x — #3290 is merged but elements nuxeo/nuxeo-elements#1499 isn't. #1499's conflicts are now resolved and it's ready, so merging it closes the gap; the #3416 fallback also protects that base in the meantime.

**1) Incompleteness — agreed, with one correction.** `iron-image` actually *does* support `crossorigin` (it's a declared property bound to its internal ``). But your underlying concern still holds: both `nuxeo-justified-grid-item` and `nuxeo-document-viewer` set `sizing`, so the pixels the user sees come from a CSS `background-image` that can't carry a CORS mode — adding `crossorigin` there just yields a second (CORS) fetch alongside the non-CORS background fetch, i.e. a partial mitigation, not a real fix. There are also a few more same-URL sites (video storyboard, the image-viewer preview, and the `nuxeo-template-rendering` layouts). Rather than ship a partial fix here, I've scoped those as a dedicated follow-up under WEBUI-2124.

**Backport:** the pair currently targets lts-2025 + maintenance-3.1.x; ELEMENTS-1616's Jira fixVersions also list 3.0.x, so we'll confirm whether a maintenance-3.0.x backport is also needed.

Same-origin safety note acknowledged — thanks for confirming.

narasimhahyland:

- **Handler naming** — renamed the thumbnail `on-error` handler from `_error` to `_onError` in both the grid and list components (template + method), to match the repo convention (`nuxeo-app`, `nuxeo-search-form`, `nuxeo-distribution-analytics`).
- **Test wiring** — the unit tests now dispatch a real `error` event on the `` (`img.dispatchEvent(new Event("error"))`) instead of calling the handler directly, so the declarative `on-error="_onError"` binding is exercised and a removed/renamed listener would be caught.

The transparent-pixel data URI stays centralized in the shared `elements/nuxeo-thumbnail-fallback.js` helper. Same changes applied to the lts-2025 PR (#3289) to keep the two branches in sync.

elements/nuxeo-thumbnail-fallback.js
madhurkulshrestha-hyland, line 1: I don't think a separate file is warranted here. A few reasons:

- **Folder convention:** `elements/` is for Polymer custom elements (`nuxeo-*` components). This module defines no element — it's just a constant plus a one-line helper — so a `nuxeo-`prefixed file here is misleading. We already have `elements/common-utils.js` for exactly this kind of shared, non-element utility. Please move `BLANK_THUMBNAIL_SRC` / `applyThumbnailFallback` there and drop this file.
- **Motivation:** The extraction was driven by Sonar duplication between the grid/list twins, which folding the constant into `common-utils.js` resolves just as well — without adding a new top-level module.
- **Redundancy with upstream:** The identical base64 pixel already lives inline in `nuxeo-document-thumbnail._error()` in nuxeo-elements, so we now have three copies across two repos. If this is meant to be shared, it would be better exported from `nuxeo-ui-elements` and reused rather than re-declared in web-ui.

Net: keep the `crossorigin` + fallback behavior, but relocate the helper into `common-utils.js` and remove `nuxeo-thumbnail-fallback.js`.

narasimhahyland: Good call — done. `elements/` should hold Polymer custom elements only, and this was just a constant + one-liner, so I've moved `BLANK_THUMBNAIL_SRC` and `applyThumbnailFallback` into the existing shared `elements/common-utils.js` (with the ELEMENTS-1616 explanatory comment), repointed the grid and list thumbnail imports to `../common-utils.js`, and deleted `elements/nuxeo-thumbnail-fallback.js`. The crossorigin + on-error transparent-pixel fallback behaviour is unchanged.

Same change has been applied to the maintenance-3.1.x PR (#3416) to keep the pair in sync.

On the wider de-duplication you flagged (the same base64 pixel is also inlined in nuxeo-elements' `nuxeo-document-thumbnail._error()` — 3 copies across 2 repos): agreed it'd be cleaner to export the pixel from nuxeo-ui-elements and have `nuxeo-document-thumbnail` reuse it. That's a cross-repo change though, so I'll track it as a separate follow-up rather than fold it into this PR.

madhurkulshrestha-hyland (approved):

naveen-konda (approved):

PR #3307, opened 2026-07-13
WEBUI-2119: Upgrade @babel/core to 7.29.7 (CVE-2026-49356)

naveen-konda:
## Summary
Resolves Dependabot alerts #336, #350, #362 — GHSA-4x5r-pxfx-6jf8 / CVE-2026-49356 (low), arbitrary file read via `sourceMappingURL`. Bumps `@babel/core` (and its `@babel/*` family) **7.29.0 → 7.29.7** across all three lockfiles (root, `packages/nuxeo-web-ui-ftest`, `plugin/a11y`). Lockfile-only; only `@babel/*` packages moved.

## Impact
🟢 Low — `@babel/core` is a **build/test-time dev dependency** (transpilation, istanbul coverage, workbox service-worker build) and is **not shipped** in the browser bundle. Patch bump; no API/transpilation-semantics change.

## Sanity
- `npm run build` emits a valid bundle + service worker
- `npm start` boots; smoke login/browse/search/open document
- CI: unit (+coverage) and ftest suites

Jira: https://hyland.atlassian.net/browse/WEBUI-2119

🤖 Generated with [Claude Code](https://claude.com/claude-code)
---
### ℹ️ Note on additional `plugin/a11y/package-lock.json` changes (not part of the CVE fix)

Reviewers / @copilot: besides the `@babel/*` bumps, this PR also updates a few lines in `plugin/a11y/package-lock.json` — the embedded **mirror** of the locally-linked `@nuxeo/nuxeo-web-ui-ftest` package (its `version`, the `multiple-cucumber-html-reporter` range, and removal of `node-fetch`). **These are not dependency decisions made in this PR.**

**Why they appear here:** `plugin/a11y` links `nuxeo-web-ui-ftest` via `file:`, so a11y's lockfile keeps a cached copy of ftest's manifest. That copy went stale because a11y's lockfile was never regenerated after:
- `node-fetch` was removed in **WEBUI-2094 (#3259)** — *"fix node-fetch v2 'Premature close' on Node ≥ 19"*
- `multiple-cucumber-html-reporter` was bumped to `^3.10.0` in ftest
- routine repo release version bumps (`2025.16 → 2025.18-SNAPSHOT`)

Running `npm update @babel/core --package-lock-only` inside `plugin/a11y` regenerated that lockfile and reconciled the stale mirror as a side effect. The change makes a11y's mirror **correct** (matches ftest's real manifest).

[...]

PR #3318, opened 2026-07-15
Dependabot CVE Fixes [LTS-2025] (WEBUI-2185)

naveen-konda:
https://hyland.atlassian.net/browse/WEBUI-2185

## Summary

Rolls up the Dependabot / CVE remediation for the LTS-2025 line into `lts-2025`.

**No application source code changes.** Every change on this branch is a `package.json` or lockfile entry across the four manifests: root, `packages/nuxeo-web-ui-ftest`, `plugin/a11y` and `packages/nuxeo-designer-catalog`.

This branch is a one-for-one port of the LTS-2023 work in #3319; the resolved dependency versions were compared commit by commit and are identical.

`npm audit` reports **zero vulnerabilities** on root, ftest and a11y.

## Tickets in this PR

| Ticket | Fix | Landed as |
| --- | --- | --- |
| WEBUI-2116 | http-proxy-middleware 2.0.10 (CVE-2026-55602) | #3306 |
| WEBUI-2119 | @babel/core 7.29.7 (CVE-2026-49356) | #3307 |
| WEBUI-2120 | esbuild 0.28.1 (GHSA-g7r4-m6w7-qqqr) | #3316 |
| WEBUI-2118 | webpack-dev-server 5.2.5 (CVE-2026-9595) | #3329 |
| WEBUI-2114 | tmp 0.2.7 (CVE-2026-44705) | #3343 |
| WEBUI-2117 | launch-editor 2.14.1 (CVE-2026-53632) | #3345 |
| WEBUI-2115 | ws 8.21.0 (CVE-2026-48779) | #3357 |
| WEBUI-2113 | form-data (CVE-2026-12143) | #3358 |
| WEBUI-2111 | shell-quote 1.10.0 (critical, CVE-2026-9277) | #3360 |
| WEBUI-2181 | postcss 8.5.25 (GHSA-r28c-9q8g-f849) | #3410 |
| WEBUI-2185 | remaining advisories across nine packages, incl. critical websocket-driver | #3413 + d87849a1e |

## Note on WEBUI-2185

The bumps in WEBUI-2115 (`ws`) and WEBUI-2113 (`form-data`) were initially defeated by hard-pinned entries in the root `package.json` `overrides` block, which take precedence over what the consuming packages would otherwise resolve ΓÇö so the manifest said one thing and the installed tree resolved another. WEBUI-2185 corrected the overrides themselves and cleared the remaining advisories, including the critical `websocket-driver`.

Key resolved versions on this branch:

| Package | Version |
| --- | --- |
| websocket-driver | 0.7.5 |
| ws | 8.21.0 |
| form-data | 2.5.6 |
| shell-quote | 1.10.0 |
| undici | 6.28.0 / 7.29.0 |
| postcss | 8.5.25 |
| tmp | 0.2.7 |
| launch-editor | 2.14.1 |
| ip-address | 10.4.0 |

## ip-address (added during review)

Three high advisories against `ip-address` <= 10.3.0 ΓÇö SSRF and trust-boundary bypasses, GHSA-mwp4-54f8-5fhr, GHSA-4xrf-jv44-h6hh and GHSA-22jq-vg5j-6vgg ΓÇö were published on 2026-08-03, after this branch was already open.

Nothing depends on it directly; it is reached only through `socks`, under the Puppeteer and webdriver tooling, so it never enters the shipped bundle. `socks` already declares `ip-address@^10.1.1`, which permits the patched 10.3.1, so **no manifest change was needed** ΓÇö only a refresh of the resolved version in the three lockfiles that carry it. `ip-address` is the only package whose version moves.

It was folded in here rather than tracked separately because this branch merges into `lts-2025`, so fixing it here resolves it for the base too.

## Dependency-resolution order (added during review)

Review feedback asked whether the flagged transitive dependencies could be cleared by upgrading their direct parents first, falling back to a lockfile re-resolve, and only then to overrides. All nine packages were re-checked against that order.

**Six needed no manifest change at all** — `websocket-driver`, `undici`, `js-yaml`, `fast-uri`, `body-parser` and `ip-address` — because the patched version was already inside the range the existing parent declared. Where a newer parent does exist (`@wdio/cli`, `mocha`, `eslint`), upgrading it would have produced the same child version, since those ranges were never the constraint; those upgrades are unrelated maintenance and don't belong in a security PR.

**Two were on overrides where a re-resolve was sufficient**, and both overrides have now been removed in `cf1bc5a4a`:

| Package | Was | Now resolves from | Result |
| --- | --- | --- | --- |
| `ws` | override `8.21.0`, exact | `webpack-dev-server ^8.18.0`, `webpack-bundle-analyzer ^8.19.0` | 8.21.0 top level, 7.5.13 under `@web/dev-server-core` |
| `form-data` | override `^2.5.6` | `nuxeo ^4.0.0` | 4.0.6 |

The same redundant `ws` pin was removed from `packages/nuxeo-web-ui-ftest/package.json`. That lockfile is byte-identical after the re-resolve, which confirms the pin was doing nothing.

Every resolved copy remains at or above its advisory threshold — CVE-2026-48779 is patched per line at 8.21.0 and 7.5.11, CVE-2026-12143 at 4.0.6. Both overrides were also working against their consumers: `form-data` had been holding the `nuxeo` client two majors below what it declares, and the exact `ws` pin had been forcing `@web/dev-server-core` across a major it never declared support for. The exact pin is also what silently defeated the original WEBUI-2115 bump.

`form-data` does not reach the shipped bundle in either version. The `nuxeo` client maps `./lib/deps/form-data.js` to a browser variant built on the native `FormData`, and the Node implementation was confirmed absent from `dist`.

**One override is still required.** `brace-expansion` stays, because every route to it runs through `minimatch`, which hard-pins it; the only alternative is a major upgrade of `glob` from 7.2.3 to 13.x.

## Verification

Verified locally against the installed tree, independently from the LTS-2023 branch:

* `npm test` — 2366 unit tests pass
* `npm run build` — production build completes
* Shipped bundle confirmed to contain the browser `FormData` shim and no Node `form-data` code
* `npm audit` — zero vulnerabilities across all four manifests
* Resolved versions compared commit by commit against LTS-2023 and are identical

All six CI workflows were green on the previous head and are re-running against `cf1bc5a4a`.


narasimhahyland:
## Review: dependency-security rollup

Reviewed the manifest changes, all four lockfiles, and the backport parity against #3319. The dependency work is sound: CI is green on all 14 checks, the lockfiles and manifests are mutually consistent, and the two branches really are one-for-one — after stripping hunk headers the two PR diffs differ only in the per-branch `-SNAPSHOT` strings and the workspace-snapshot resync, so neither release line is left behind.

Three things to address before merge, one of them blocking.

---

### 1. `ws` pinned to `8.21.0` leaves CVE-2026-62389 unfixed — blocking

`overrides.ws` moves from one exact pin to another (`8.20.1` → `8.21.0`), in both the root and `packages/nuxeo-web-ui-ftest` manifests. `8.21.0` is affected by:

**[GHSA-73jw-fp74-p77x](https://github.com/advisories/GHSA-73jw-fp74-p77x) / CVE-2026-62389** (high) — *"ws before **8.21.1** contains a memory exhaustion vulnerability in `lib/receiver.js` where the fragment guard only triggers when fragment count reaches `maxFragments`, allowing attackers to exhaust memory by sending incomplete fragmented WebSocket messages."*

**Why `npm audit` doesn't show this.** The advisory is `type: unreviewed` and its `vulnerabilities` array is **empty** — it carries no machine-readable version range, so neither `npm audit` nor the npm bulk-advisory API can match it. Reproducible:

```bash
# returns only GHSA-96hv-2xvq-fx4p, ">=8.0.0 <8.21.0" — 8.21.0 looks clean
curl -s -H 'content-type: application/json' -X POST \
  https://registry.npmjs.org/-/npm/v1/security/advisories/bulk \
  -d '{"ws":["8.21.0"]}'

# but the advisory exists, is high, and says "before 8.21.1"
curl -s https://api.github.com/advisories/GHSA-73jw-fp74-p77x \
  | python3 -c "import json,sys;a=json.load(sys.stdin);print(a['severity'],a['type'],a['vulnerabilities']);print(a['summary'])"
```

So the PR's "zero vulnerabilities" claim is accurate as measured, but the tool can't see this one.

**The strongest evidence is inside this PR.** `plugin/a11y/package-lock.json` has no `ws` override and resolved to **`8.21.1`**, while root and ftest — the two manifests that *are* overridden — sit at `8.21.0`. The patched version was available when these lockfiles were generated; the exact pin is the only thing holding root and ftest back. That is the same failure mode the "Note on WEBUI-2185" section describes fixing for `8.20.1`, reproduced one version later.

**Suggested fix:** `"ws": "^8.21.1"` in both `package.json` and `packages/nuxeo-web-ui-ftest/package.json`, then refresh the lockfiles. A caret range also avoids a third round trip — `8.21.2` shipped 2026-08-03. This would additionally align all four manifests on one version.

**Scope, to be fair:** the root `ws` is dev-only (webpack-dev-server, Puppeteer, WebdriverIO, `@web/dev-server-core`), so this is not a production exposure — but it does mean the stated goal isn't met on either release line.

---

### 2. The `form-data` override is unscoped and downgrades `nuxeo` two majors

`overrides."form-data": "^2.5.6"` is a bare top-level key, so it applies to the whole root tree. Verified against this branch's `package-lock.json`:

- exactly one `form-data` in the root tree, at `2.5.6`, **not** `dev`
- its **sole** requirer is `node_modules/nuxeo@4.0.6`, which declares `"form-data": "^4.0.0"`
- there is no nested `node_modules/nuxeo/node_modules/form-data`
- `nuxeo` is in root `dependencies` (`^4.0.3`), not `devDependencies`

So the Nuxeo client SDK is silently forced two majors below its own declared range. `packages/nuxeo-web-ui-ftest`, which has no `form-data` override, resolves the identical dependency to `4.0.6` — meaning the ftest suite exercises a different `form-data` line than the root tree and cannot detect a 2.x incompatibility there.

Both `2.5.6` and `4.0.6` are advisory-clean ([GHSA-hmw2-7cc7-3qxx](https://github.com/advisories/GHSA-hmw2-7cc7-3qxx) is patched at 2.5.6 / 3.0.5 / 4.0.6), so this is a compatibility risk rather than a vulnerability, and it predates this PR. But the PR edits that exact line, so it's a natural moment to fix it.

**Suggested fix:** since `nuxeo@4.0.6`'s own `^4.0.0` already resolves to the patched `4.0.6` — as the ftest lockfile demonstrates — the override looks obsolete and could simply be dropped. If a floor must be recorded, scope it: `"nuxeo": { "form-data": "^4.0.6" }`.

---

### 3. `webpack-dev-server` resolves to 5.2.6, not the 5.2.5 in the description

The ticket table records *"WEBUI-2118 | webpack-dev-server **5.2.5** (CVE-2026-9595)"*, and the "Key resolved versions" table omits the package entirely, but the lockfile moves `5.2.4` → **`5.2.6`**.

The resolved version is the correct one — 5.2.6 is required, not optional:

| Advisory | Affected | Severity |
| --- | --- | --- |
| [GHSA-mx8g-39q3-5c79](https://github.com/advisories/GHSA-mx8g-39q3-5c79) | `< 5.2.5` | moderate |
| [GHSA-f5vj-f2hx-8m93](https://github.com/advisories/GHSA-f5vj-f2hx-8m93) | `<= 5.2.5` | moderate |
| [GHSA-m28w-2pqf-7qgj](https://github.com/advisories/GHSA-m28w-2pqf-7qgj) | `<= 5.2.5` | moderate |

Only the description is stale. Worth noting in the body that 5.2.6 adds a `Sec-Fetch-Site` / Origin check on the dev server's `/webpack-dev-server/invalidate` and `/open-editor` endpoints, since this repo runs a dev server behind a `/nuxeo` proxy and that is a behaviour change, not an inert patch.

---

### Worth considering (non-blocking)

- **`plugin/a11y` has no `ws` override**, which is why it floated to `8.21.1`. Once the pin above is fixed, decide deliberately whether a11y should carry a matching override or keep resolving independently. More generally, a11y and ftest lack most of the security overrides root has, so a future lockfile regeneration in those directories has nothing holding the floor.
- **Two override floors sit below the versions they're meant to guarantee.** `"tmp": "^0.2.4"` permits `0.2.4` and `0.2.5`, both inside [GHSA-ph9p-34f9-6g65](https://github.com/advisories/GHSA-ph9p-34f9-6g65) (`< 0.2.6`); `ip-address` has no floor at all and relies entirely on `socks`' `^10.1.1`, which permits the vulnerable `10.1.1`–`10.3.0`. The lockfiles pin patched versions so there's no exposure today — the gap is that a regeneration wouldn't be caught by manifest review.
- **`brace-expansion@5.0.9` drops Node 18** (engines narrow from `18 || 20 || >=22` to `20 || >=22`), while all three manifests still declare `node >= 18` and at least one workflow installs on Node 18. Without `engine-strict` this is an `EBADENGINE` warning rather than a failure — supported-platform drift, not a breakage.
- **`nanoid@3.3.16`** still carries the zero-size infinite loop; the upstream changelog shows `3.3.16` fixed the *negative*-size loop and `3.3.17` fixed the *zero*-size one. There is no advisory with a structured range covering `3.3.16`, and reaching it needs attacker-controlled `size=0` through build-time PostCSS, so this is low priority — just something to pick up on the next refresh.

### Documentation nits

- The `ip-address` section says "three **high** advisories ... against `<= 10.3.0`". It is one high and two moderate, with three different ranges: `<= 10.3.0` ([GHSA-mwp4-54f8-5fhr](https://github.com/advisories/GHSA-mwp4-54f8-5fhr), high), `>= 10.1.1 <= 10.2.1` ([GHSA-4xrf-jv44-h6hh](https://github.com/advisories/GHSA-4xrf-jv44-h6hh), moderate), `>= 10.1.1 <= 10.2.0` ([GHSA-22jq-vg5j-6vgg](https://github.com/advisories/GHSA-22jq-vg5j-6vgg), moderate).
- "`ip-address` is the only package whose version moves" isn't literally true — that commit also rewrites the `../../packages/nuxeo-web-ui-ftest` workspace snapshot version in `plugin/a11y/package-lock.json`.
- The audit claim covers root, ftest and a11y, but `packages/nuxeo-designer-catalog/package-lock.json` also changed (`brace-expansion` 1.1.14 → 1.1.18) and CI installs it too.
- The "Key resolved versions" table lists a single `form-data 2.5.6`; ftest resolves `4.0.6`. Same for `ws 8.21.0` vs a11y's `8.21.1`.
- The workspace-link resync (`multiple-cucumber-html-reporter` ^1.18.0 → ^3.10.0, `node-fetch` dropped, `minimist` added) reads as undisclosed non-security drift, but it is actually a **correction** — the base lockfile's snapshot of that workspace package was stale and now matches its `package.json` exactly. Worth one line in the body so a future auditor doesn't have to re-derive that.

---

*Reviewed with a panel of six models; findings above are the ones I independently verified against the GitHub advisory API, the npm registry, and this branch's lockfiles. The same comment applies to #3319 — the `ws` and `form-data` items need to land on both branches.*

madhurkulshrestha-hyland (approved):

vivekchaturved013 (approved):

PR #3321, opened 2026-07-16
WEBUI-2138: include custom schema columns in Web UI CSV export

narasimhahyland:
### Problem
CSV exports generated from Web UI results (search and folder content) omit properties that belong to custom document schemas. Only fields from `dublincore`, `common`, `file` and `uid` are exported. The same metadata is available on the documents and is included when exporting through JSF UI, or through the backend CSV bulk action when the custom schema is provided explicitly.

Ticket: https://hyland.atlassian.net/browse/WEBUI-2138

### Root cause
`nuxeo-csv-export-button` builds the `csvExport` bulk-action `schemas` parameter from the page provider's display schemas. For the default search that value is the contribution's hardcoded `schemas="dublincore,common,file,uid"` (kept narrow for display performance). Server-side, `DocumentModelCSVWriter` emits columns only for the exact schema names supplied — it silently skips unknown names, has no "all"/wildcard fallback, and rejects `*` with HTTP 400. So every custom-schema property is dropped from the export.

### Changes
- `elements/nuxeo-csv-export/nuxeo-csv-export-button.js`: resolve the export schemas from the document **types actually present in the provider results** (via `config/types`), unioned with the provider's display schemas, and use that when no explicit `schemas` is set. Each exported document's own schemas — including custom ones — are now included, without hardcoding schema names. Explicit `schemas` still takes precedence; falls back to the provider's display schemas when result types or their config are unavailable.
- `elements/fetch-types.js`: small cached `config/types` fetch helper (mirrors `fetch-schemas.js`).
- Unit tests for the resolution logic and the new helper.

### Test plan
- `npm run lint` — passes.
- `npm test` — passes (2255 tests), including new `nuxeo-csv-export-button` and `fetch-types` cases.
- Manual A/B against a 2025 server, inspecting the `Bulk.RunAction` request (the ticket's repro step):
  - Before: `{"schemas":["dublincore","common","uid","file"]}` — custom/extra schema columns missing.
  - After: `{"schemas":["dublincore","common","uid","file","note","files","facetedTag","relatedtext"]}` — the document type's own schemas (a stand-in for the reported custom schema) are now exported, with no unrelated schema noise.

### Notes
- Export runs over the whole provider result set; the schema list is derived from the currently loaded result page's types. For heterogeneous, multi-page results, types that appear only on not-yet-loaded pages are not added. This is strictly better than the previous hardcoded behavior and covers the reported cases.
- Backported to both `lts-2025` and `maintenance-3.1.x`.

### DIFF
--- a/elements/fetch-types.js
+++ b/elements/fetch-types.js
@@ -1,9 +1,14 @@
-let typesCache;
+let typesPromise;
 
-export const _fetchTypes = async (resource) => {
-  if (!typesCache) {
+export const _fetchTypes = (resource) => {
+  if (!typesPromise) {
     resource.path = 'config/types';
-    typesCache = await resource.get();
+    // cache the in-flight promise so concurrent callers share a single config/types request;
+    // clear it on failure so a later call can retry
+    typesPromise = resource.get().catch((e) => {
+      typesPromise = undefined;
+      throw e;
+    });
   }
-  return typesCache;
+  return typesPromise;
 };
--- a/elements/fetch-types.js
+++ b/elements/fetch-types.js
@@ -1,0 +1,9 @@
+let typesCache;
+
+export const _fetchTypes = async (resource) => {
+  if (!typesCache) {
+    resource.path = 'config/types';
+    typesCache = await resource.get();
+  }
+  return typesCache;
+};
--- a/elements/nuxeo-csv-export/nuxeo-csv-export-button.js
+++ b/elements/nuxeo-csv-export/nuxeo-csv-export-button.js
@@ -110,6 +110,8 @@ Polymer({
   ready() {
     this.$.btn.addEventListener('poll-start', this._onPollStart.bind(this));
     this.$.btn.addEventListener('response', this._onResponse.bind(this));
+    // resolve once the local DOM (`this.$.types`) exists, in case `provider` was data-bound before `ready`
+    this._resolveSchemas();
   },
 
   _providerChanged(provider, oldProvider) {
@@ -132,6 +134,9 @@ Polymer({
    * `provider`'s schemas when the result types or their configuration are not available.
    */
   async _resolveSchemas() {
+    // track the latest resolution so a slower in-flight call cannot overwrite a newer one with a stale type set
+    const token = (this._resolveToken || 0) + 1;
+    this._resolveToken = token;
     const provider = this.provider;
     if (!provider) {
       this._resolvedSchemas = undefined;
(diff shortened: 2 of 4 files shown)


narasimhahyland:
- **Initial resolution before `ready()`**: `_resolveSchemas()` is now also called from `ready()`, so the first resolution runs once `this.$.types` exists even when `provider` is data-bound before `ready` (no longer stuck on display schemas until the next `current-page-changed`).
- **Stale async resolution**: `_resolveSchemas()` now tracks a monotonic resolution token; only the latest in-flight call updates `_resolvedSchemas`, so a slower earlier call can't overwrite a newer one on rapid page changes.
- **Flaky cache-busting import**: the `fetch-types` test now uses a monotonic nonce instead of `Date.now()` to guarantee a unique module specifier per load.

Lint + unit tests (2255) pass locally.

narasimhahyland:
Addressed the latest Copilot review comments:
- **Listener leak** ([nuxeo-csv-export-button.js](https://github.com/nuxeo/nuxeo-web-ui/blob/master/elements/nuxeo-csv-export/nuxeo-csv-export-button.js)): added a `detached()` callback that removes the `current-page-changed` listener from the current provider, so no listener leaks on page/`dom-if` switches.
- **Test comment**: fixed the stale "Karma" wording in `test/fetch-types.test.js` (this repo uses `@web/test-runner`).

Lint + unit tests pass locally (2266 tests); new `detached()` code is fully covered. Backported to both lts-2025 (#3320) and maintenance-3.1.x (#3321).

madhurkulshrestha-hyland:
Resolve copilot comment

narasimhahyland:
Updated the `schemas` property JSDoc to document the actual precedence: when unset, the type-derived `_resolvedSchemas` union is used, falling back to the provider's own schemas when the resolved set is unavailable. (Also applied to #3320.)

AnilKumarVanga (dismissed):

narasimhahyland:
Addressed the latest Copilot comments on both bases:
- **fetch-types**: cache the in-flight `config/types` promise so concurrent `_resolveSchemas()` calls share a single request (cleared on failure so a later call can retry).
- **error logging**: pass the caught error as a separate `console.warn` argument instead of interpolating it, preserving the stack (no more `[object Object]`).

Added unit tests for the concurrent-dedupe and retry-after-failure paths. Lint + unit tests pass (2269); new code fully covered.

madhurkulshrestha-hyland (approved):

AnilKumarVanga (approved):

PR #3374, opened 2026-07-29
WEBUI-2009: Show full result names in the search drawer [maintenance-3.1.x]

vivekchaturved013:
## Problem

In the search drawer, each result name is rendered on a single line and clipped with an
ellipsis, with no tooltip. When several documents share a long common prefix — the
reporter's case is `D-DF-ASTURL61CZRIE-LSPS-PLANNING-2025-REV-01`, `-REV-02`, `-REV-03` —
every row renders as the same truncated string, so the list is unusable: the only way to
tell the results apart is to open each one.

Reported on [WEBUI-2009](https://hyland.atlassian.net/browse/WEBUI-2009).

## Root cause

`elements/search/nuxeo-search-form.js` renders each result name as
`<span class="list-item-title ellipsis">`. The element's local `.ellipsis` rule is
`text-overflow: ellipsis; overflow: hidden; white-space: nowrap`, so the name is truncated
and nothing exposes the untruncated value.

Measured on a live instance before the change: of 36 rendered rows, 2 were clipped and 0
carried any tooltip.

## Change

Every result row gets a `nuxeo-tooltip` carrying `[[item.title]]`, opening to the right of
the row:

```html
<span class="list-item-title ellipsis">[[item.title]]</span>
<nuxeo-tooltip position="right" offset="8" aria-hidden="true"
  ><span class="nuxeo-search-result-tooltip-name">[[item.title]]</span></nuxeo-tooltip
>
```

Names deliberately stay on a single truncated line, and every row gets the tooltip rather
than only the clipped ones. Four details are load-bearing:

- **The tooltip sits outside the name span.** Nesting it would append a duplicate of the
  title to the span's own `textContent`.
- **It is anchored on the parent row, not by `for`.** This template is stamped once per
  result into the same shadow root, so an `id` would not be unique and every row's tooltip
  would resolve to the first match.
- **It opens to the right.** Below the row it would cover the next results while the user
  is scanning the list.
- **It is `aria-hidden`.** The name is already exposed as the row's text; announcing it a
  second time is noise.

### Why a document-level stylesheet

`nuxeo-tooltip` does not render its content in place: on `show()` it creates a
`paper-tooltip`, appends it to `document.body`, and clones the slot content into it. The
rendered node therefore lives outside this element's shadow root, so a rule in the Polymer
template never matches it — the widget's own JSDoc says as much, and it solves the same
problem for itself with `ensureClonedContentStyles()` and its `resize-handle` role.

Without a cap the bubble is sized purely by its content, so a name with no break
opportunity runs off-screen. Measured on another consumer of the same widget (the grid
thumbnail): a 241-character name renders **2067px wide with ~787px off-screen** at a
1280px viewport.

`nuxeo-search-result-tooltip-styles.js` therefore injects one document-level rule, scoped
by a class name that travels with the clone:

```css
.nuxeo-search-result-tooltip-name {
  display: block;
  max-width: min(400px, 40vw);
  white-space: normal;
  overflow-wrap: break-word;
  line-height: 1.4;
  text-align: start;
}
```

The cap is absolute *and* viewport-relative so it still holds when the user zooms in, which
shrinks the CSS viewport while an absolute cap would stay put. Injection is idempotent.

`nuxeo-search-form` backs the `defaultSearch`, `expiredSearch`, `assetsSearch` and `trash`
drawers, so all four benefit.

## Test plan

Local gate, the same checks CI gates on:

- `npx eslint` and `npx prettier --check` on all four changed files — clean.
- `npm test` — **2318 passed, 0 failed**, including the coverage gate.
- SonarCloud — quality gate passed, **0 new issues**, 100% coverage on new code.

Thirteen new unit tests: nine in `test/nuxeo-search-form.test.js` pin the binding text, the
placement outside the span, the parent anchoring, `aria-hidden`, the single-line truncation
with no `-webkit-line-clamp`, the right-hand placement, the wrapper element the cap matches,
that a name with no break opportunity stays inside the viewport, and that every row
`iron-list` stamps gets its own tooltip. Four in
`test/nuxeo-search-result-tooltip-styles.test.js` cover the injection, its idempotence, the
width cap and the wrapping declarations. The stylesheet tests were mutation-checked by
removing the injection and confirming they fail.

Verified against a live instance (Nuxeo 2025 plus the webpack dev server), search drawer,
results-list view:

| Check | Before | After |
|---|---|---|
| Rows carrying a tooltip | 0 of 36 | 36 of 36 |
| Tooltip text matching the row title exactly | — | 36 of 36 |
| `white-space` / `text-overflow` on the name | `nowrap` / `ellipsis` | unchanged |
| Lines shown per name | 1 | 1 |
| `-webkit-line-clamp` anywhere in the element | none | none |
| Tooltip placement | — | right of the row |
| Bubble containment, 241-char unbreakable name | runs off-screen | inside the viewport |
| Panel header `white-space` | `nowrap` | `nowrap` (unchanged) |

Containment was probed at 1280px and 640px viewport widths against three seeded names — one
broken by spaces, one by hyphens, and one 241-character token with no break opportunity at
all — including rows scrolled to the very top and bottom of the list, where `paper-tooltip`
has to flip to stay in bounds.

## Notes

- A tooltip on every row was a deliberate product decision, rather than measuring geometry
  to tooltip only the clipped rows. It keeps the template declarative: whether a name is
  clipped depends on the drawer's current width and `iron-list` recycles rows between
  documents, so a conditional tooltip needs hover-time measurement to stay correct.
- `paper-tooltip` fades in via a compositor-driven opacity animation that Chrome's
  screenshot rasteriser does not reflect. It is hit-testable and reports `opacity: 0.9` yet
  comes out invisible in in-page captures, so the attached evidence is a desktop grab from a
  headful browser.
- The same single-line truncation exists in the Favorites, Clipboard, Collections and
  Recently viewed drawers. Those were deliberately left out to keep this PR scoped to what
  was reported.
- The missing width cap is a `nuxeo-tooltip` gap rather than a search one — any consumer
  passing long content hits it. Worth a follow-up in nuxeo-elements; the measurements above
  are the starting evidence.

[WEBUI-2009]: https://hyland.atlassian.net/browse/WEBUI-2009?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/elements/search/nuxeo-search-form.js
+++ b/elements/search/nuxeo-search-form.js
@@ -237,19 +237,6 @@ Polymer({
         @apply --layout-center;
       }
 
-      /* Show the name on up to two lines before clipping it, so results sharing a long
-         prefix stay distinguishable. Kept self-contained rather than leaning on the
-         .ellipsis class these spans also carry, which the panel header still needs for
-         single-line truncation. */
-      .list-item-title {
-        display: -webkit-box;
-        -webkit-line-clamp: 2;
-        -webkit-box-orient: vertical;
-        overflow: hidden;
-        overflow-wrap: break-word;
-        white-space: normal;
-      }
-
       .list-item-property {
         opacity: 0.5;
         margin-right: 0.2em;
@@ -377,7 +364,10 @@ Polymer({
                     <div class="vertical layout center">
                       <nuxeo-document-thumbnail document="[[item]]"></nuxeo-document-thumbnail>
                     </div>
-                    <span class="list-item-title ellipsis" on-mouseenter="_showTitleIfTruncated">[[item.title]]</span>
+                    <span class="list-item-title ellipsis">[[item.title]]</span>
+                    <!-- Kept outside the span so the name text stays exactly the title, and
+                         anchored on the parent row since an id would not be unique per stamp. -->
+                    <nuxeo-tooltip aria-hidden="true">[[item.title]]</nuxeo-tooltip>
                   </div>
                 </div>
               </div>
@@ -784,19 +774,6 @@ Polymer({
       classes += ' selected';
     }
     return classes;
-  },
-
-  // Whether a name is clipped depends on rendered geometry, not on the document, so it
-  // cannot be expressed as a binding: the drawer is resizable and iron-list recycles rows
-  // into different documents. Measuring on hover keeps the tooltip accurate at the current
-  // width without re-measuring every row on resize or scroll.
-  _showTitleIfTruncated(e) {
-    const el = e.currentTarget;
-    if (el.scrollHeight > el.clientHeight + 1) {
-      el.setAttribute('title', el.textContent.trim());
-    } else {
-      el.removeAttribute('title');
-    }
   },
 
   _selectedDocChanged(doc, old) {
--- a/elements/search/nuxeo-search-form.js
+++ b/elements/search/nuxeo-search-form.js
@@ -44,6 +44,9 @@ import { html } from '@polymer/polymer/lib/utils/html-tag.js';
 import { Debouncer } from '@polymer/polymer/lib/utils/debounce.js';
 import { timeOut } from '@polymer/polymer/lib/utils/async.js';
 import { FormatBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-format-behavior.js';
+import { ensureSearchResultTooltipStyles } from './nuxeo-search-result-tooltip-styles.js';
+
+ensureSearchResultTooltipStyles();
 
 /**
  `nuxeo-search-form`
@@ -366,8 +369,13 @@ Polymer({
                     </div>
                     <span class="list-item-title ellipsis">[[item.title]]</span>
                     <!-- Kept outside the span so the name text stays exactly the title, and
-                         anchored on the parent row since an id would not be unique per stamp. -->
-                    <nuxeo-tooltip aria-hidden="true">[[item.title]]</nuxeo-tooltip>
+                         anchored on the parent row since an id would not be unique per stamp.
+                         Opens to the right of the row so it never covers the result list, and the
+                         name is wrapped in an element so the document-level rule that caps its
+                         width can match it once nuxeo-tooltip clones it onto the body. -->
+                    <nuxeo-tooltip position="right" offset="8" aria-hidden="true"
+                      ><span class="nuxeo-search-result-tooltip-name">[[item.title]]</span></nuxeo-tooltip
+                    >
                   </div>
                 </div>
               </div>
--- a/elements/search/nuxeo-search-form.js
+++ b/elements/search/nuxeo-search-form.js
@@ -238,12 +238,14 @@ Polymer({
       }
 
       /* Show the name on up to two lines before clipping it, so results sharing a long
-         prefix stay distinguishable. Overrides the single-line .ellipsis rule, which the
-         panel header still relies on. */
+         prefix stay distinguishable. Kept self-contained rather than leaning on the
+         .ellipsis class these spans also carry, which the panel header still needs for
+         single-line truncation. */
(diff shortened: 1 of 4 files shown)


elements/search/nuxeo-search-form.js
```diff
@@ -237,6 +237,17 @@ Polymer({
...
+      /* Show the name on up to two lines before clipping it, so results sharing a long
+         prefix stay distinguishable. Overrides the single-line .ellipsis rule, which the
+         panel header still relies on. */
+      .list-item-title {
+        display: -webkit-box;
+        -webkit-line-clamp: 2;
+        -webkit-box-orient: vertical;
+        overflow-wrap: break-word;
+        white-space: normal;
+      }
```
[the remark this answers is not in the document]
vivekchaturved013: Good catch - fixed in 75cb40565 (and 2a47995be on the lts-2025 PR). `overflow: hidden` is now declared on `.list-item-title` itself, so the two-line clamp no longer depends on `.ellipsis` being present. The comment was reworded too, since the rule no longer overrides `.ellipsis` so much as stand on its own - the panel header still needs that class for single-line truncation.

elements/search/nuxeo-search-form.js
```diff
@@ -44,6 +44,9 @@ import { html } from '@polymer/polymer/lib/utils/html-tag.js';
 import { Debouncer } from '@polymer/polymer/lib/utils/debounce.js';
 import { timeOut } from '@polymer/polymer/lib/utils/async.js';
 import { FormatBehavior } from '@nuxeo/nuxeo-ui-elements/nuxeo-format-behavior.js';
+import { ensureSearchResultTooltipStyles } from './nuxeo-search-result-tooltip-styles.js';
+
+ensureSearchResultTooltipStyles();
```
madhurkulshrestha-hyland, line 49: Will this not work if the style is added inside the Polymer template function ? If it works, this will eliminate usage of a separate file.

madhurkulshrestha-hyland (changes requested):

elements/search/nuxeo-search-form.js
vivekchaturved013: Good question — that was the first thing I tried, and it does not work with this widget.

`nuxeo-tooltip` does not render its content in place. On `show()` it creates a `paper-tooltip`, appends it to `document.body`, and clones the slot content into that. The node that actually renders therefore lives outside this element's shadow root, and a rule declared in the Polymer template is scoped to that root, so it never matches the clone. The widget documents this itself:

> On show, assigned slot nodes are cloned into a `paper-tooltip` attached to `document.body`. Styles from the consumer shadow tree are not applied to the clone.

I verified the mechanism rather than assuming it — with the rule declared in the template's `<style>` block, the cloned node's computed style does not carry it. And the overflow it leaves behind is not marginal: a 241-character name renders 2067px wide with roughly 787px off-screen at a 1280px viewport (measured on another consumer of the same widget, the grid thumbnail).

`nuxeo-tooltip` solves this for itself the same way. `ensureClonedContentStyles()` injects a document-level stylesheet, and the existing `data-nx-tooltip-role="resize-handle"` rule caps its width there. This module is that same pattern applied from the consumer side; the class name travels with the clone, which is what keeps the rule scoped to search results rather than leaking to every tooltip.

If the concern is the extra file rather than the mechanism, I can fold `ensureSearchResultTooltipStyles()` into `nuxeo-search-form.js` so there is no separate module — it just cannot live inside the template. Happy to do that if you would prefer it.

madhurkulshrestha-hyland (approved):

narasimhahyland (approved):

AnilKumarVanga (approved):

PR #1532, opened 2022-07-22
WEBUI-450: fix isSavedSearch flag condition updated

Diksha-Verma-Dev:
https://jira.nuxeo.com/browse/WEBUI-450

In default search save button is hidden
Screenshot 2022-07-22 at 11 25 16 AM

PR #3375, opened 2026-07-29
WEBUI-1946: do not run the default NXQL query on page load [LTS-2025]

vivekchaturved013:
## Problem

The Administration > NXQL Search page runs a broad query as soon as it is opened:

```
SELECT * FROM Document WHERE ecm:mixinType != 'HiddenInNavigation' AND ecm:isProxy = 0 AND ecm:isVersion = 0 AND ecm:isTrashed = 0
```

On customers with large repositories this unbounded query is executed on every visit to the page and destabilises the instance and its database. The request is that the query only run when the user presses Search.

Ticket: https://hyland.atlassian.net/browse/WEBUI-1946

## Root cause

Two independent code paths fetch on page load, and neither depends on `auto` (which is already `false` for this page):

1. `nuxeo-results-view._formChanged()` searches as soon as the search form layout loads — addressed by the paired nuxeo-elements PR.
2. `nuxeo-results._viewChanged()` calls `this.fetch()` as soon as the display view (the data table) initialises.

Because both funnel into a fetch, suppressing only one of them still leaves the query running.

## Changes

- `nuxeo-results`: new opt-in `deferInitialFetch`. While it is set and no explicit fetch has happened yet, view initialisation no longer fetches. `fetch()` records that it ran, so switching display mode (table/grid) after a real search still refetches as before.
- `nuxeo-search-page`: new `deferInitialSearch` property, passed through to `nuxeo-results-view`.
- `nuxeo-web-ui-bundle.html`: `defer-initial-search` on the NXQL admin search page.
- `nuxeo-nxql-search-results.html`: `defer-initial-fetch` on its `nuxeo-results`.
- `admin.feature`: the NXQL scenario now asserts the page opens with no results **and** that pressing Search returns them, so a regression in either the deferral or Search itself fails the test.

Both flags default to `false`, so every other search page keeps searching on load. The default query stays prefilled in the query box as today — it simply is not executed until the user presses Search or Ctrl+Enter.

## Test plan

- `npm test` — **2309 passed, 0 failed** on both `lts-2025` and `maintenance-3.1.x`.
- `Deferred Initial Fetch (WEBUI-1946)` suite in `test/nuxeo-results.test.js` covers: fetching on view initialisation by default; not fetching when `deferInitialFetch` is set; fetching on view initialisation once an explicit fetch has happened; and an explicit fetch still reaching the view while deferred. The tests drive the `view` property so Polymer's observer runs the real initialisation path.
- The NXQL ftest scenario proves both halves of the behaviour — empty on load, results after pressing Search — instead of only asserting that nothing is on screen, and the empty-state step waits for the results area to render rather than sleeping for a second.
- `npm run lint:eslint` clean. `npm run lint:prettier` flags the whole repository when run on Windows because `prettier.config.js` sets no `endOfLine` and the working copy uses CRLF — a pre-existing local-only artifact. The touched files pass with `--end-of-line crlf`.

## Notes

- Merge order: `deferInitialFetch` is a Web UI flag on our own `nuxeo-results`, so it takes effect immediately against published elements. The remaining path — `nuxeo-results-view._formChanged()` → `_search()` → `nuxeo-search-results-layout.fetch()` — only stops querying once the paired nuxeo-elements PR (ELEMENTS-2016) is released; in released elements, `fetch()` no-ops unless `this.results` has been assigned in its `afterNextRender`, so whether the query runs depends on which layout import resolves first. **nuxeo-elements must therefore be released before this merges**, otherwise the NXQL ftest here would be asserting import-ordering-dependent behaviour.
- Clearing the filters keeps the query deferred, handled in the paired nuxeo-elements PR: Search followed by Clear no longer re-runs the broad query.
- Unchanged on purpose: the results table keeps its existing `searchResults.noResults` empty label before the first search.
- Raised on both `lts-2025` and `maintenance-3.1.x`.


naveen-konda (commented):
Reviewed together with the paired nuxeo-elements PR nuxeo/nuxeo-elements#1532.

The approach looks right, and I like that both flags are opt-in and default to `false` so every other search page is untouched. Spotting the second fetch path in `nuxeo-results._viewChanged()` was the key insight here — suppressing only the results-view path would indeed have left the query running. A few points before I approve.

### 1. Merge order and the ftest change

The description says this PR is inert until the elements change is released, and that behaviour is "unchanged rather than broken" until then. I don't think that holds, and it matters because this PR also flips `admin.feature` to `And I can see no search results`.

`deferInitialFetch` is a Web-UI-local flag on our own `nuxeo-results`, so it applies immediately with the currently published `@nuxeo/nuxeo-ui-elements` — path 2 (view initialisation) is suppressed regardless of the elements version. The path that remains is `nuxeo-results-view._formChanged()` → `_search()` → `nuxeo-search-results-layout.fetch()`, which in released elements is:

```js
fetch() {
  if (this.results) {
    this.results.fetch();
  } else {
    return Promise.resolve();
  }
}
```

`this.results` there is assigned inside an `afterNextRender` once the results layout import resolves. So if the form layout resolves first this no-ops and no query runs; if the results layout resolves first, the query runs. That's import-ordering dependent rather than "unchanged".

CI is green today because the cross-repo workflow resolves the nuxeo-elements branch by name (`get-branch` with `branch: ${{ github.event.inputs.branch_name }}`) and both PRs use the same branch name, so the ftest is running with the elements fix in place. Once this merges, the ftest on `lts-2025` runs against published elements. Could we release nuxeo-elements#1532 first, and drop or adjust the "either order" note in both descriptions?

### 2. `fetch()` lifts the deferral for every caller, not just Search

`fetch()` sets `_fetched = true` unconditionally, and it is reached from a few places besides the Search button:

- `_sortChanged()`
- `_refreshAndFetch()` (toolbar refresh)
- `_scheduleQuickFilterFetch()`

Since `nuxeo-nxql-search-results.html` sets `display-quick-filters`, a user can toggle a quick filter before ever pressing Search, which runs the broad query and permanently lifts the deferral. These are all user-initiated so it's arguably fine — I mainly want to confirm it's intentional rather than incidental, since the JSDoc says "until a fetch is requested explicitly (e.g. by the enclosing search view's search button)".

### 3. The new ftest step can pass for the wrong reason

```js
Then(/^I can see no search results$/, async function () {
  await driver.pause(1000);
  ...
  if (output !== 0) { throw ... }
});
```

A fixed 1s sleep followed by an assertion that *nothing* is there will also pass when the server is simply slow, so it won't reliably catch a regression where the deferral stops working. Could we strengthen the scenario so it proves both halves — assert no results on load, then press the page's Search button and assert results appear? That would need a new step for the results-view Search button (the existing `I click the QuickSearch button` targets the top-nav quick search, not this one), but it would make the test fail loudly if either the deferral or Search itself regressed.

### Nits

- `_fetched` doesn't need to be a declared Polymer property — nothing binds or observes it — and unlike `deferInitialFetch` right next to it, it has no JSDoc.

vivekchaturved013:
Thanks @naveen-konda — all three addressed in 831c62b3f, cherry-picked to #3376 as 1036f47ce.

### 1. Merge order and the ftest change

You're right and my note was wrong. `deferInitialFetch` is our own flag, so it applies immediately against published elements, and I confirmed the remaining path behaves as you describe — in the published `@nuxeo/nuxeo-ui-elements`, `nuxeo-search-results-layout.fetch()` no-ops unless `this.results` has been assigned, and that assignment happens inside the `afterNextRender` in its `_formChanged`, so whether the query runs depends on which layout import resolves first.

All four descriptions now state that nuxeo-elements #1532 / #1533 must be released before these merge, and the "either order" note is gone.

### 2. `fetch()` lifting the deferral for every caller

Intentional, and now documented rather than implied:

> If `true`, initializing a display mode does not fetch results, so loading the page executes no query. Only the deferred *initial* fetch is suppressed: any user action that asks for results — the enclosing search view's Search button, a sort change, the toolbar refresh or a quick filter toggle — fetches as usual and lifts the deferral for later display-mode changes.

The reasoning: sort, refresh and quick filters are all explicit requests for results, and suppressing them would leave visible controls that silently do nothing until Search is pressed. Only the automatic page-load fetch is suppressed, which is what the ticket asks for. Happy to narrow it to the Search button alone if you would rather have the stricter behaviour.

### 3. The ftest passing for the wrong reason

Agreed — the scenario now proves both halves:

```gherkin
When I click "nxql" in the administration menu
Then I can see the nxql search page
And I can see no search results
When I click the nxql search button
Then I can see more than 1 search results
```

For the new step I added `nxqlSearchButton` to the administration page object (`paper-button.search` within `nuxeo-search-page#nxql`, which the shadow-piercing selectors resolve), so it targets this page's Search button rather than the top-nav quick search. The empty-state step no longer sleeps: it waits for the results area to render and resolves the display mode before counting, so an empty count means no query ran rather than one still being in flight. If the deferral regresses, the first assertion fails; if Search regresses, the last one does.

### Nit

`_fetched` is now a plain instance field initialised in `created()` with a comment, out of `properties`.

`npm test`: **2309 passed, 0 failed**. eslint clean; the touched files pass prettier with `--end-of-line crlf` (the repo-wide prettier noise on Windows is the pre-existing CRLF artifact). The ftests will run on CI.

Separately, the Clear button is no longer out of scope — your first point on nuxeo/nuxeo-elements#1532 was right, so clearing the filters on this page now returns to the deferred state instead of re-running the broad query.

naveen-konda (dismissed):

elements/nuxeo-results/nuxeo-results.js
```diff
@@ -707,8 +730,7 @@ Polymer({
         view.quickFilters = restoredQuickFilters.slice();
       }
 
-      // fetch after state restore
-      this.fetch();
+      this._fetchOnViewInit();
```
madhurkulshrestha-hyland, line 733: Any reason for adding the defer-intial-fetch attribute ? We ALWAYS want the fetch not to be triggered on load. So why not just remove **_this.fetch()_** on line 711

madhurkulshrestha-hyland (changes requested):

elements/nuxeo-results/nuxeo-results.js
vivekchaturved013: this.fetch() there is the initial fetch for every listing that uses nuxeo-results, not just NXQL. Folder browse, the folder trash tab, picturebook and nuxeo-default-results have no other trigger — no auto provider, no explicit fetch() — so removing it opens folders with an empty table (11 of 45 ftest feature files depend on it). defer-initial-fetch is exactly that toggle, defaulted to false so only the NXQL page opts out.
The page has a second, independent trigger: nuxeo-results-view._formChanged() → _search() in nuxeo-elements, which fires when the form loads regardless of this element. Suppressing either one alone still leaves a query running on load, which is why the fix is paired — deferInitialSearch guards that path, deferInitialFetch guards this one, and both stay opt-in so no other consumer changes behaviour.

madhurkulshrestha-hyland (dismissed):

narasimhahyland (approved):

naveen-konda (approved):

PR #3384, opened 2026-07-30
WEBUI-482: keep the validation error summary in view

narasimhahyland:
## Problem

Web UI signals an unfilled required field with colour and a thicker underline only, failing WCAG 2.1
SC 1.4.1 (Use of Color) and SC 3.3.3 (Error Suggestion).

The substance of the fix lives in nuxeo-elements: nuxeo/nuxeo-elements#1544 makes
`nuxeo-layout` report which widgets failed validation, gives them a message naming the field, and
has `nuxeo-document-layout` list the failing fields in an alert summary above the form.

This PR fixes the one thing on the Web UI side that stopped that summary from being seen.

## Root cause

`nuxeo-document-layout` scrolls its error summary into view when validation fails, but both form
hosts then run `invalidField.scrollIntoView()` on the first invalid widget. The default
`block: 'start'` aligns that widget with the top of the dialog scroller, which pushes the summary
sitting above the form straight back out of view — so the newly added list of failing fields was
never visible.

## Changes

- `elements/document/nuxeo-document-create.js` and `elements/document/nuxeo-document-form-layout.js`:
  scroll the first invalid field with `{ block: 'nearest' }`, which still brings it into view and
  still focuses it, without scrolling the summary away.

## Test plan

- `npm run lint` and `npm test` pass locally (2305 passed, 0 failed).
- Verified end to end against a running instance with nuxeo/nuxeo-elements#1544 linked in, on the
  File creation form, with a DOM probe rather than pixels alone. Before: no `aria-invalid`, no
  `aria-required`, no message, no summary. After: `aria-invalid="true"`, `aria-required="true"`,
  a visible message under the field, and an `alert` summary above the form naming the field —
  both visible in the same viewport. Screenshots and screen recordings are attached to the Jira
  issue.

## Notes

- Depends on nuxeo/nuxeo-elements#1544; on its own this change is a no-op improvement to scroll
  behaviour.
- The new user facing strings live in nuxeo-elements `ui/i18n/messages.json` and are English only;
  the other locales need translation.


narasimhahyland:
Addressed the Copilot feedback on both bases: the invalid field is now focused with `preventScroll: true`, so focusing it cannot re-scroll the container and undo the `nearest` alignment that keeps the validation error summary in view.

naveen-konda (dismissed):

vivekchaturved013 (dismissed):

madhurkulshrestha-hyland (dismissed):

narasimhahyland:
Mirrored from the maintenance-3.1.x twin #3385, where a review found that `nuxeo-document-import.js` still had the bare `scrollIntoView()` / `focus()` pair this PR fixes in the create and edit forms — so the import form kept the same WCAG SC 1.4.1 / 3.3.3 failure. Fixed in 8f4675603.

Also swapped the scroll/focus assertions in all three form tests from `calledOnceWithExactly` to comparing the recorded args: with the source reverted, the spy matcher's failure never reaches the reporter and the Web Test Runner session hangs until `testsFinishTimeout` with `0 passed, 0 failed`, whereas the args comparison fails cleanly in 2s. Same strictness, but a regression now shows up as a failed test rather than as flaky infrastructure.

Gates: `npm run lint` clean, `npm test` 2560 passed / 0 failed.

madhurkulshrestha-hyland (approved):

naveen-konda (approved):

PR #3394, opened 2026-07-31
WEBUI-899: Make focus order target quick search before main content [lts-2025]

narasimhahyland:
## Problem
Keyboard focus reached the quick search **after** entering the main-content landmark, and it cost two Tab presses to get there. Activating **Skip to main content** and pressing Tab landed on Quick Search instead of on the document, and one Tab stop earlier the focus ring disappeared entirely because it was sitting on an invisible element. Jira: [WEBUI-899](https://hyland.atlassian.net/browse/WEBUI-899)

## Root cause
In `elements/nuxeo-app.js`, `<nuxeo-suggester id="suggester" tabindex="0">` was rendered as the first child of `<main id="mainContent">`, even though its only visible control (`#searchButton`) is `position: fixed` in the top app bar (`--nuxeo-suggester-button` in `themes/base.js`). Two consequences, both measured with a DOM probe that walks `document.activeElement` through the shadow roots on every Tab:

* Quick search was **inside** the `main` landmark, so focus entered main content before reaching it — `skipLink` → Enter → `main#mainContent` → Tab → Quick Search.
* `tabindex="0"` on the wrapper added a tab stop on an element that renders nothing of its own (measured `1090x0`), so Tab first landed on an invisible element with no focus indicator and only the next Tab reached the actual button.

Tab order before the fix (steps 8-11 of the probe):

```
08  nuxeo-app >> nuxeo-resize-handle#drawerResizeHandle
09  nuxeo-app >> nuxeo-suggester#suggester              rect={"w":1090,"h":0}  inMainLandmark=true
10  nuxeo-app >> nuxeo-suggester#suggester >> paper-icon-button#searchButton   inMainLandmark=true
11  nuxeo-app >> nuxeo-browser#browser >> nuxeo-breadcrumb >> a
```

## Changes
* **`elements/nuxeo-app.js`**: move `<nuxeo-suggester>` out of `<main id="mainContent">` and into the app header toolbar next to `#drawerToggle`, and drop the redundant `tabindex="0"`. Quick search is now reached before the main-content landmark, as a single tab stop on the button that is actually visible. No positive `tabindex` is introduced.
* **`test/nuxeo-app.test.js`**: assert the suggester is in `app-header app-toolbar`, is not contained by `#mainContent`, and carries no `tabindex`.

Tab order after the fix:

```
08  nuxeo-app >> nuxeo-resize-handle#drawerResizeHandle
09  nuxeo-app >> nuxeo-suggester#suggester >> paper-icon-button#searchButton   inMainLandmark=false
10  nuxeo-app >> nuxeo-browser#browser >> nuxeo-breadcrumb >> a
```

Nothing moves on screen: the search button and the search overlay are positioned with `position: fixed` from the theme mixin, so they are independent of their DOM parent. The before/after screenshots of the document view are byte-identical.

## Test plan
- [x] `npm run lint` passes
- [x] `npm test` passes — 2306 passed, 0 failed
- [x] Tab-order probe on a running 2025 instance, before and after: quick search moves from tab stop 9+10 (inside `main`) to a single stop 9 (outside `main`); after **Skip to main content**, the next Tab now goes to the document breadcrumb instead of Quick Search.
- [x] Quick search still opens with the search button, with the `/`, `ctrl+space` and `s` shortcuts, and the overlay's arrow/enter/esc keys still work (the `nuxeo-keys` targets listen on the host, which still receives composed events).

## Notes
Backport of the same commit to `maintenance-3.1.x` is in a companion PR.
The `nuxeo-web-ui-ftest` page objects reach quick search by id (`#suggester`, `#searchButton`), so they are unaffected by the move.

[WEBUI-899]: https://hyland.atlassian.net/browse/WEBUI-899?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/elements/nuxeo-app.js
+++ b/elements/nuxeo-app.js
@@ -410,6 +410,7 @@ Polymer({
     </style>
     <header role="banner">
       <a href="#mainContent" id="skipLink" class="skip-link">[[i18n('app.skiptoMainContent.message')]]</a>
+      <nuxeo-suggester id="suggester"></nuxeo-suggester>
 
       <nuxeo-offline-banner message="[[i18n('app.offlineBanner.message')]]"></nuxeo-offline-banner>
 
@@ -545,7 +546,6 @@ Polymer({
               aria-label$="[[i18n('command.menu')]]"
               tabindex="-1"
             ></paper-icon-button>
-            <nuxeo-suggester id="suggester"></nuxeo-suggester>
           </app-toolbar>
         </app-header>
 
--- a/elements/nuxeo-app.js
+++ b/elements/nuxeo-app.js
@@ -545,11 +545,11 @@ Polymer({
               aria-label$="[[i18n('command.menu')]]"
               tabindex="-1"
             ></paper-icon-button>
+            <nuxeo-suggester id="suggester"></nuxeo-suggester>
           </app-toolbar>
         </app-header>
 
         <main id="mainContent" tabindex="-1">
-          <nuxeo-suggester id="suggester" tabindex="0"></nuxeo-suggester>
           <iron-pages id="pages" selected="[[page]]" attr-for-selected="name" selected-attribute="visible">
             <nuxeo-slot name="PAGES" model="[[actionContext]]"></nuxeo-slot>
 
--- a/test/nuxeo-app.test.js
+++ b/test/nuxeo-app.test.js
@@ -1844,10 +1844,12 @@ suite('nuxeo-app', () => {
   suite('accessibility and menu keyboard', () => {
     test('quick search sits in the app header, ahead of the main content landmark', () => {
       const { suggester, mainContent } = app.$;
+      const toolbar = app.shadowRoot.querySelector('app-header app-toolbar');
       expect(suggester).to.be.ok;
       expect(mainContent).to.be.ok;
+      expect(toolbar).to.be.ok;
       expect(mainContent.contains(suggester)).to.be.false;
-      expect(app.shadowRoot.querySelector('app-header app-toolbar').contains(suggester)).to.be.true;
+      expect(toolbar.contains(suggester)).to.be.true;
       // The focusable control is the inner search button; a tabindex on the wrapper
       // would add a tab stop on an element that renders nothing.
       expect(suggester.hasAttribute('tabindex')).to.be.false;
--- a/test/nuxeo-app.test.js
+++ b/test/nuxeo-app.test.js
@@ -1842,6 +1842,17 @@ suite('nuxeo-app', () => {
   });
 
   suite('accessibility and menu keyboard', () => {
+    test('quick search sits in the app header, ahead of the main content landmark', () => {
+      const { suggester, mainContent } = app.$;
+      expect(suggester).to.be.ok;
+      expect(mainContent).to.be.ok;
+      expect(mainContent.contains(suggester)).to.be.false;
+      expect(app.shadowRoot.querySelector('app-header app-toolbar').contains(suggester)).to.be.true;
+      // The focusable control is the inner search button; a tabindex on the wrapper
+      // would add a tab stop on an element that renders nothing.
+      expect(suggester.hasAttribute('tabindex')).to.be.false;
+    });
+
     test('skipLinkEvent focuses main content on Enter', () => {
       const { skipLink, mainContent } = app.$;
       if (!skipLink || !mainContent) {
--- a/test/nuxeo-app.test.js
+++ b/test/nuxeo-app.test.js
@@ -1842,14 +1842,19 @@ suite('nuxeo-app', () => {
   });
 
   suite('accessibility and menu keyboard', () => {
-    test('quick search sits in the app header, ahead of the main content landmark', () => {
+    test('quick search is tabbable before main content without a transformed ancestor', () => {
       const { suggester, mainContent } = app.$;
-      const toolbar = app.shadowRoot.querySelector('app-header app-toolbar');
+      const banner = app.shadowRoot.querySelector('header[role="banner"]');
       expect(suggester).to.be.ok;
       expect(mainContent).to.be.ok;
-      expect(toolbar).to.be.ok;
+      expect(banner).to.be.ok;
+      const searchButton = suggester.shadowRoot.querySelector('#searchButton');
+      expect(searchButton).to.be.ok;
       expect(mainContent.contains(suggester)).to.be.false;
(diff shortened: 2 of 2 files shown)


test/nuxeo-app.test.js
```diff
@@ -1842,6 +1842,17 @@ suite('nuxeo-app', () => {
   });
 
   suite('accessibility and menu keyboard', () => {
+    test('quick search sits in the app header, ahead of the main content landmark', () => {
+      const { suggester, mainContent } = app.$;
+      expect(suggester).to.be.ok;
+      expect(mainContent).to.be.ok;
+      expect(mainContent.contains(suggester)).to.be.false;
+      expect(app.shadowRoot.querySelector('app-header app-toolbar').contains(suggester)).to.be.true;
```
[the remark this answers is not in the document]
narasimhahyland: Good catch — fixed in f1d8c10c2: the toolbar element is now captured in a `toolbar` variable and asserted with `expect(toolbar).to.be.ok` before `toolbar.contains(suggester)`, so a selector mismatch fails as an assertion instead of a TypeError. Mirrored onto the `maintenance-3.1.x` PR as 40cead64c.

naveen-konda (dismissed):

AnilKumarVanga (dismissed):

elements/nuxeo-app.js
```diff
@@ -545,11 +545,11 @@ Polymer({
               aria-label$="[[i18n('command.menu')]]"
               tabindex="-1"
             ></paper-icon-button>
+            <nuxeo-suggester id="suggester"></nuxeo-suggester>
```
vivekchaturved013: Two things follow from the suggester becoming a child of `app-toolbar` inside `<app-header reveals effects="waterfall">`.

**1. `app-header` always carries a `transform`, so it becomes the containing block for the suggester's fixed-position chrome.**

`_updateScrollState` calls `_transformHeader(top)` on the initial forced layout pass and on every scroll (`@polymer/app-layout/app-header/app-header.js` lines 565-570), and `_transformHeader` sets `style.transform` on the header host via `this.translate3d(0, (-y) + 'px', 0)` (line 619-620). A transformed ancestor becomes the containing block for `position: fixed` descendants, and every piece of visible suggester chrome is fixed:

- `#searchButton` - `position: fixed; top: 0; right: 0`, plus the themed `--nuxeo-suggester-button` giving it `top: var(--nuxeo-app-top); right: 0; 60x53`
- `#suggester` - `position: fixed; width: 100%; height: 100%`
- `.fade` - `position: fixed`, `top`/`bottom`/`left`, `width: 100%`

Those now resolve against the roughly 64px `app-header` box instead of the viewport, so the overlay becomes full-width by ~64px, `#results { height: calc(100% - 130px - ...) }` goes negative and collapses, and the icon rides up out of view whenever `reveals` hides the header. `overflow: hidden` on `app-header-layout` (line 137-144 of this file) will clip whatever survives.

If that reproduces, the banner block at lines 412-418 looks like a safer home - immediately after the skip link, outside `app-drawer-layout`, no transformed ancestor, and still ahead of the main-content landmark in both DOM and tab order.

**2. Flex growth in the toolbar.**

The host is `@apply --layout-flex`, i.e. `flex: 1 1 0.000000001px`. As a child of the flex-row `app-toolbar` it now absorbs the toolbar's free space. Nothing should visibly move today since the toolbar holds only `#drawerToggle`, but it becomes a live layout bug the moment a `nuxeo-slot` is added there. Given the host is now purely a positioning shell, pinning it with `flex: none` from `nuxeo-app` would close that off.

test/nuxeo-app.test.js
```diff
@@ -1842,6 +1842,19 @@ suite('nuxeo-app', () => {
   });
 
   suite('accessibility and menu keyboard', () => {
+    test('quick search sits in the app header, ahead of the main content landmark', () => {
```
vivekchaturved013: Agreed on the reasoning in the comment, and worth stating plainly since it is the real win here: everything the host renders is either `hidden` or `position: fixed`, so it was a zero-size tab stop where focus visibly disappeared, and `#searchButton` is a `paper-icon-button` that is already tabbable through IronControlState.

The test asserts DOM containment, which is a proxy for the actual requirement. Asserting the order itself would be closer to the ticket - for example that `suggester.shadowRoot.querySelector('#searchButton')` is tabbable and precedes `#mainContent` in a tabbable-order walk. Wiring the orphaned quick-search ftest steps into a `.feature` file would also give this area its first end-to-end coverage.

vivekchaturved013 (commented):
Review of the quick-search focus-order change. Moving the suggester out of `main` and dropping the wrapper `tabindex` both look like the right calls to me - see the inline note on the test for why the `tabindex` removal is a genuine fix. My concern is the new location: `app-header` permanently carries a `transform`, which changes the containing block for everything the suggester positions. Details inline on `nuxeo-app.js`. Applies equally to the `maintenance-3.1.x` twin (#3395), whose diff is byte-identical.

### Worth knowing: CI does not cover this area at all

The added unit test asserts DOM position only. The quick-search ftest steps do exist - `packages/nuxeo-web-ui-ftest/features/step_definitions/search.js` lines 283-300 define "I click the QuickSearch button", "I perform a QuickSearch for ..." and "I can see N QuickSearch results" - but no `.feature` file references any of them, so the green `ftests` run exercises none of it. Relatedly, `pages/ui.js` lines 150-151 still select `#mainContainer nuxeo-suggester`, an id that is not in the template any more.

So the green checks here should not be read as evidence that quick search still works. A manual pass would be worth it: click the quick-search icon, type a term, confirm the full-screen overlay and results render, then scroll a long result list and confirm the icon stays pinned.

### Regression surface

Header layout and tab order on every page; quick search under each theme (`--nuxeo-suggester-button` is themed per theme in `themes/base.js` lines 559-584, so dark/light/kawaii are each worth a look); RTL, which relies on `:host([dir='rtl']) #searchButton { left: 0 }`; and narrow/mobile, where the `@media (max-width: 1024px)` rules apply and `#drawerToggle` becomes visible in the same toolbar.

No i18n impact - the button's `aria-label` already goes through `i18n('pickerSearch.title')`.

vivekchaturved013 (approved):

naveen-konda (approved):

madhurkulshrestha-hyland (approved):

PR #3404, opened 2026-07-31
WEBUI-2180: fix Web UI failing to load when an addon chunk 404s

madhurkulshrestha-hyland:
JIRA: [WEBUI-2180](https://hyland.atlassian.net/browse/WEBUI-2180)

## Problem
On environments where an addon package such as `nuxeo-platform-3d` is installed (e.g. the LTS-2023 `rainforest-docker-2023` env), the Web UI fails to load with:

```
Uncaught (in promise) TypeError: Failed to fetch dynamically imported module:
https://.../nuxeo/ui/nuxeo-platform-3d.bundle.js
```

## Root cause
Two changes combined:
- **WEBUI-2061** (content-hashed `chunkFilename`) removed the stable `nuxeo-platform-3d.bundle.js`; a stale/cached `main.bundle.js` references chunk hashes removed on upgrade, so the primary dynamic import 404s and the fallback 404s too. WEBUI-2061 documented the prerequisite: `main.bundle.js` must be served with `Cache-Control: no-cache`.
- **WEBUI-1978** moved `loadAddons` **before** `loadRouting`, so a rejected addon load now short-circuits the bootstrap chain (previously it ran last and was harmless). That's why it surfaced only now.

## Fix
- **sw.js** — dedicated route for `main.bundle.js` using `fetch(request, { cache: 'no-cache' })`, registered before the generic `.js` route, so the entry bundle always revalidates and the webpack runtime chunk-hash map is current (enforces the WEBUI-2061 prerequisite client-side). Root-cause fix.
- **index.js** — isolate each addon load in `loadAddons` with a per-addon `.catch()` so a missing/broken optional addon logs a warning instead of aborting `loadRouting`/bootstrap. Defense-in-depth; does not change the WEBUI-1978 ordering.

## Testing
- `eslint` + `prettier --check` pass on both files.

## Follow-up (out of scope)
- Verify the deployed `nuxeo-platform-3d` package ships a resolvable client bundle.
- Confirm the server/CDN also honors `Cache-Control: no-cache` on `main.bundle.js`.

[WEBUI-2180]: https://hyland.atlassian.net/browse/WEBUI-2180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/index.js
+++ b/index.js
@@ -27,15 +27,18 @@ const loadAddons = async () => {
   // NXP-26977: await loading of addons
   await Promise.all(
     bundles.map((url) => {
-      if (url.endsWith('.html')) {
-        return new Promise((resolve, reject) => importHref(url, resolve, reject));
-      }
-      return import(
-        /* webpackChunkName: "[request]" */
-        /* webpackInclude: /addons\/[^\/]+\/index.js$/ */
-        // eslint-disable-next-line comma-dangle
-        `./addons/${url}`
-      ).catch(() => import(/* webpackIgnore: true */ `./${url}.bundle.js`));
+      const load = url.endsWith('.html')
+        ? new Promise((resolve, reject) => importHref(url, resolve, reject))
+        : import(
+            /* webpackChunkName: "[request]" */
+            /* webpackInclude: /addons\/[^\/]+\/index.js$/ */
+            // eslint-disable-next-line comma-dangle
+            `./addons/${url}`
+          ).catch(() => import(/* webpackIgnore: true */ `./${url}.bundle.js`));
+      // A missing or broken optional addon bundle (e.g. a server package like nuxeo-platform-3d
+      // that is installed on the server but has no resolvable client bundle) must not reject the
+      // whole bootstrap chain. Isolate each addon load and log a warning instead of failing startup.
+      return Promise.resolve(load).catch((e) => console.warn(`Failed to load addon bundle "${url}":`, e));
     }),
   );
 };
--- a/sw.js
+++ b/sw.js
@@ -74,23 +74,26 @@ workbox.routing.registerRoute(
 );
 
 /* ================================
-   Entry bundle → ALWAYS revalidate (bypass HTTP cache)
+   Entry bundle → ALWAYS revalidate
    main.bundle.js embeds the webpack runtime chunk-hash map. It MUST be fresh so
    dynamically-imported chunks (e.g. addon bundles like nuxeo-platform-3d) resolve to
    content-hashed filenames that still exist on the server after an upgrade. A stale
    entry bundle references removed hashes → 404 (WEBUI-2061 prerequisite:
    Cache-Control: no-cache for main.bundle.js). Registered before the generic .js
-   route below so it always wins for the entry bundle, regardless of the `ts` param.
-   Uses `no-cache` (not `reload`) so the browser still revalidates with a conditional
-   request and reuses the cached bytes on a 304 — cheap when the bundle is unchanged.
+   route below so it always wins for the entry bundle.
+   `no-cache` forces the browser to revalidate with a conditional request (reusing the
+   cached bytes on a 304, cheap when unchanged); the `ts` server-start param — matching
+   the generic .js route — additionally busts intermediate/CDN caches on every upgrade.
 ================================ */
 workbox.routing.registerRoute(
   ({ url }) => url.pathname.endsWith('/main.bundle.js'),
-  async ({ event }) =>
-    fetch(event.request, {
+  async ({ event }) => {
+    const request = TS ? new Request(`${event.request.url}?ts=${TS}`, { credentials: 'same-origin' }) : event.request;
+    return fetch(request, {
       cache: 'no-cache', // force revalidation of the entry bundle on every load
       credentials: 'same-origin',
-    }),
+    });
+  },
 );
 
 /* ================================
--- a/sw.js
+++ b/sw.js
@@ -81,19 +81,27 @@ workbox.routing.registerRoute(
    entry bundle references removed hashes → 404 (WEBUI-2061 prerequisite:
    Cache-Control: no-cache for main.bundle.js). Registered before the generic .js
    route below so it always wins for the entry bundle.
-   `no-cache` forces the browser to revalidate with a conditional request (reusing the
-   cached bytes on a 304, cheap when unchanged); the `ts` server-start param — matching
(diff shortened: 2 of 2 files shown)


sw.js
```diff
@@ -73,6 +73,26 @@ workbox.routing.registerRoute(
   },
 );
 
+/* ================================
+   Entry bundle → ALWAYS revalidate (bypass HTTP cache)
+   main.bundle.js embeds the webpack runtime chunk-hash map. It MUST be fresh so
```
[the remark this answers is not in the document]
madhurkulshrestha-hyland: Good catch — reworded the comment in c8cfa0b to say `no-cache` *revalidates* (reusing bytes on a 304) rather than fully bypassing the HTTP cache like `reload`.

AnilKumarVanga (dismissed):

naveen-konda (dismissed):

sw.js
```diff
@@ -73,6 +73,29 @@ workbox.routing.registerRoute(
...
+   route below so it always wins for the entry bundle.
+   `no-cache` forces the browser to revalidate with a conditional request (reusing the
+   cached bytes on a 304, cheap when unchanged); the `ts` server-start param — matching
+   the generic .js route — additionally busts intermediate/CDN caches on every upgrade.
+================================ */
+workbox.routing.registerRoute(
+  ({ url }) => url.pathname.endsWith('/main.bundle.js'),
+  async ({ event }) => {
+    const request = TS ? new Request(`${event.request.url}?ts=${TS}`, { credentials: 'same-origin' }) : event.request;
+    return fetch(request, {
```
narasimhahyland: This bare `fetch()` has no cache fallback, so an offline reload or a transient network failure now fails to load the entry bundle entirely — whereas the previous generic `NetworkFirst` route could boot the app from the same-version cached copy. The stale-bundle risk only arises across a server upgrade (when `?ts` changes anyway), so a `.catch(() => caches.match(request))` fallback — or a `NetworkFirst` strategy keyed on `?ts` — would keep the freshness guarantee while preserving offline boot. Non-blocking if offline boot isn't supported here.

index.js
```diff
@@ -27,15 +27,18 @@ const loadAddons = async () => {
...
+        : import(
+            /* webpackChunkName: "[request]" */
+            /* webpackInclude: /addons\/[^\/]+\/index.js$/ */
+            // eslint-disable-next-line comma-dangle
+            `./addons/${url}`
+          ).catch(() => import(/* webpackIgnore: true */ `./${url}.bundle.js`));
+      // A missing or broken optional addon bundle (e.g. a server package like nuxeo-platform-3d
+      // that is installed on the server but has no resolvable client bundle) must not reject the
+      // whole bootstrap chain. Isolate each addon load and log a warning instead of failing startup.
+      return Promise.resolve(load).catch((e) => console.warn(`Failed to load addon bundle "${url}":`, e));
```
narasimhahyland, line 41: This isolates each addon load nicely, but it now swallows failures for *all* addons — including non-optional ones (`nuxeo-spreadsheet` is always loaded per AGENTS.md). A broken core addon will now only `console.warn` instead of surfacing. Fine as defense-in-depth; just flagging it's an intentional behavior change so a required-addon regression won't fail loudly.

narasimhahyland (commented):
Reviewed the full `sw.js` (route ordering, the `TS`/generic `.js` route this must precede, precache behavior) and `index.js` `loadAddons`, plus the ticket and existing threads.

**Overall:** solid, well-documented fix with an accurate root-cause writeup. The new `main.bundle.js` route is correctly registered **before** the generic `.js` route (Workbox is first-match-wins), so the entry bundle now bypasses the `NetworkFirst` `JS_CACHE` that could otherwise serve a stale entry bundle referencing removed chunk hashes — that's the right root-cause fix. The per-addon `.catch()` in `loadAddons` is sensible defense-in-depth for the WEBUI-1978 ordering change. Two approvals already; I'd just flag one trade-off before merge.

### Worth a look (non-blocking)
- **Offline / flaky-network resilience of the entry bundle.** The old generic route cached `main.bundle.js` via `NetworkFirst`, so an offline reload (or a transient network blip) could still boot the app from the *same-version* cached entry bundle. The new route is a bare `fetch()` with no cache fallback, so if that fetch fails the app no longer boots at all. The stale-bundle danger only exists **across a server upgrade** (hashes changed); within the same version the cached copy is valid. Consider `NetworkFirst`/`StaleWhileRevalidate` keyed by the `?ts=` server-start param (so a stale-version copy is simply a cache miss), or at least a `.catch(() => caches.match(request))` fallback — that keeps the upgrade-time freshness guarantee while preserving offline boot. If offline boot isn't a supported scenario here, ignore this.

### Minor
- **`loadAddons` now swallows *all* addon failures**, including ones that aren't optional (`nuxeo-spreadsheet` is always loaded per AGENTS.md). A genuinely broken required addon now only logs a warning. Acceptable as defense-in-depth, but worth a conscious call — a broken core addon will fail silently rather than loudly.
- **No unit test.** `loadAddons`'s new isolation (a rejecting addon must not reject the aggregate) is cheap to cover with a small test and would lock in the behavior this PR is specifically adding; `sw.js` is understandably harder to test.

### CI
- `ftests` was still running when I looked; lint/unit/sonar/a11y/CodeQL all green.

## Verdict
Looks good to merge. The offline-resilience trade-off is the only thing I'd consciously confirm is intended.

sw.js
madhurkulshrestha-hyland: Addressed in 2c50878 (and 5d19245 on the lts-2025 PR): switched the entry-bundle route to `NetworkFirst` keyed on `?ts` with `fetchOptions.cache: 'no-cache'`. Revalidation still prevents a stale HTTP/CDN copy from being served, but the same-version (`?ts`-keyed) cached fallback now restores offline/transient-failure boot; a cross-upgrade `?ts` change still forces a fresh fetch. Thanks for the catch.

index.js
madhurkulshrestha-hyland: Correct, this is intentional. The whole point of WEBUI-2180 is that no single addon — optional or required — should take down the entire Web UI bootstrap; degrading one addon (with a `console.warn`) is preferable to a blank app. If we want required addons like `nuxeo-spreadsheet` to fail loudly, that'd be a separate, deliberate policy (e.g. an allowlist of fatal bundles); happy to follow up under a new ticket if desired, but keeping it uniform here for the hotfix.

PR #1498, opened 2022-06-20
WEBUI-723: fix selection in nuxeo-easysharefolder-view-layout

Gabez0r:

PR #3416, opened 2026-08-04
ELEMENTS-1616: transparent-pixel fallback for failed grid/list thumbnails [maintenance-3.1.x]

narasimhahyland:
## Problem

PR #3290 (already merged to `maintenance-3.1.x`) added `crossorigin="anonymous"` to the grid and list thumbnail `` tags. Unlike `nuxeo-document-thumbnail`, those two elements have **no `on-error` handler**, so if a cross-origin thumbnail request fails (e.g. S3 direct download to a bucket without CORS configured) the view now renders a broken-image icon instead of degrading gracefully.

This is the follow-up flagged in review of #3289 (the lts-2025 sibling). Because #3290 is already merged, the maintenance-3.1.x fix lands as this separate PR.

## Root cause

`crossorigin` turns a previously always-succeeding no-CORS `` request into one that fails when the response lacks `Access-Control-Allow-Origin`. Without a fallback, the failure is visible as a broken image.

## Changes

- `elements/nuxeo-data-grid/nuxeo-document-grid-thumbnail.js` — add `on-error="_error"` and an `_error` handler that swaps the failed thumbnail for a transparent pixel (mirrors `nuxeo-document-thumbnail`).
- `elements/nuxeo-data-list/nuxeo-document-list-item.js` — same.
- Tests assert the `_error` fallback replaces the `src` with an inline transparent PNG.

## Test plan

- `npm run lint` — clean
- Unit: new `_error` tests in both suites
- Manual: S3 direct download without bucket CORS → grid/list show transparent pixel instead of broken-image icon

## Notes

Pairs with the crossorigin work in ELEMENTS-1616 (elements #1498/#1499, web-ui #3289/#3290). Tracked under WEBUI-2124. See also the merge-order note: `nuxeo-document-thumbnail` gains `crossorigin` only via elements #1499 (still open) on this base.


narasimhahyland:

- **Handler naming** — renamed the thumbnail `on-error` handler from `_error` to `_onError` in both the grid and list components (template + method), to match the repo convention (`nuxeo-app`, `nuxeo-search-form`, `nuxeo-distribution-analytics`).
- **Test wiring** — the unit tests now dispatch a real `error` event on the `` (`img.dispatchEvent(new Event("error"))`) instead of calling the handler directly, so the declarative `on-error="_onError"` binding is exercised and a removed/renamed listener would be caught.

The transparent-pixel data URI stays centralized in the shared `elements/nuxeo-thumbnail-fallback.js` helper. Same changes applied to the lts-2025 PR (#3289) to keep the two branches in sync.

madhurkulshrestha-hyland (approved):

naveen-konda (approved):

PR #3421, opened 2026-08-07
WEBUI-2189: Return user to originating page after session-expiry re-login [lts-2025]

narasimhahyland:
## Problem
After WEBUI-1987 shipped, an inactivity/401 session expiry now redirects straight to the timeout login page and, after the user re-authenticates, drops them on the app root (dashboard) instead of the page they were viewing. The originating page context is lost. Jira: [WEBUI-2189](https://hyland.atlassian.net/browse/WEBUI-2189)

## Root cause
`NuxeoInactivityBehavior._logoutRedirect()` (`elements/behaviors/nuxeo-inactivity-behavior.js`) ends the server session and navigates to `login.jsp?nxtimeout=true` with no record of where the user was. It deliberately bypasses the SPA anonymous-login bounce (which nests/drops a `requestedUrl`), so the server never learns the origin page. `location.replace()` also removes it from history, so the back button can't recover it.

## Changes
* **`elements/behaviors/nuxeo-inactivity-behavior.js`**: before the logout navigation, `_saveRequestedUrl()` persists the current URL to `sessionStorage` (only on the inactivity/401 path — manual Sign Out is unaffected). New `_restoreRequestedUrlAfterLogin()` consumes it once on the next boot and navigates back **only** when the saved URL is same-origin (open-redirect guard using `origin + '/'`, not a bare prefix) and differs from the current page.
* **`elements/nuxeo-app.js`**: call `_restoreRequestedUrlAfterLogin()` at the start of `ready()`, before the timer/401 handlers are armed.
* **`test/nuxeo-inactivity-behavior.test.js`**: 9 new tests (save-on-logout, same-origin restore + consume-once, cross-origin/look-alike-host rejection, current-page no-op, empty no-op, sessionStorage-unavailable handling).

Why `sessionStorage` (not the ticket's suggested `localStorage`): it survives the same-tab logout → login → reload sequence but is scoped to the originating tab and cleared when the tab closes — bounding staleness and preventing a cross-tab hijack — while remaining same-origin.

## Test plan
- [x] `npm run lint` passes
- [x] `npm test` passes (2401 passed, 0 failed)
- [ ] Manual: set `session.timeout` low, open a document, go idle → land on `login.jsp` timeout message → log back in → verify you return to the document (not the dashboard). Also verify manual Sign Out still lands on login/home.

## Notes
Backport pair: this targets `lts-2025`; a mirrored PR targets `maintenance-3.1.x`. Follow-up to WEBUI-1987 (both bases already have the merged behavior).

[WEBUI-2189]: https://hyland.atlassian.net/browse/WEBUI-2189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/elements/behaviors/nuxeo-inactivity-behavior.js
+++ b/elements/behaviors/nuxeo-inactivity-behavior.js
@@ -290,7 +290,7 @@ export const NuxeoInactivityBehavior = {
     try {
       const payload = JSON.stringify({
         url: globalThis.location.href,
-        user: this.currentUser && this.currentUser.id,
+        user: this.currentUser?.id,
       });
       globalThis.sessionStorage.setItem(INACTIVITY_REQUESTED_URL_KEY, payload);
     } catch (e) {
@@ -346,7 +346,7 @@ export const NuxeoInactivityBehavior = {
     const baseUrl = this.baseUrl || '/';
     const normalizedBase = baseUrl.endsWith('/') ? baseUrl : `${baseUrl}/`;
     const base = `${globalThis.location.origin}${normalizedBase}`;
-    if (requestedUrl && requestedUrl.startsWith(base) && requestedUrl !== globalThis.location.href) {
+    if (requestedUrl?.startsWith(base) && requestedUrl !== globalThis.location.href) {
       this._redirect(requestedUrl);
     }
   },
--- a/elements/behaviors/nuxeo-inactivity-behavior.js
+++ b/elements/behaviors/nuxeo-inactivity-behavior.js
@@ -287,10 +287,17 @@ export const NuxeoInactivityBehavior = {
   // so _restoreRequestedUrlAfterLogin() can return the user here once they log back in. The owning user's
   // id is stored alongside the URL so restore only ever returns a page to the same user who left it.
   _saveRequestedUrl() {
+    // Only persist a page we can actually return later: without a resolved user id we couldn't match the
+    // saved URL to its owner on restore, so it would be consumed but never navigated. Skip the save instead
+    // of writing an unrestorable `{ user: undefined }` payload.
+    const userId = this.currentUser?.id;
+    if (!userId) {
+      return;
+    }
     try {
       const payload = JSON.stringify({
         url: globalThis.location.href,
-        user: this.currentUser?.id,
+        user: userId,
       });
       globalThis.sessionStorage.setItem(INACTIVITY_REQUESTED_URL_KEY, payload);
     } catch (e) {
--- a/elements/behaviors/nuxeo-inactivity-behavior.js
+++ b/elements/behaviors/nuxeo-inactivity-behavior.js
@@ -339,14 +339,32 @@ export const NuxeoInactivityBehavior = {
     if (savedUser !== currentUser.id) {
       return;
     }
-    // Restrict to the UI base path (e.g. `/nuxeo/ui/`, which implies same-origin) so a saved value can't
-    // slip past into an open redirect. baseUrl is absent in tests/edge cases → fall back to a bare '/'.
-    // Normalize to a single trailing '/' (mirroring the origin `+ '/'` reasoning): without it a baseUrl of
-    // `/nuxeo/ui` would let a look-alike sibling path such as `${origin}/nuxeo/ui-foo/...` pass startsWith().
-    const baseUrl = this.baseUrl || '/';
-    const normalizedBase = baseUrl.endsWith('/') ? baseUrl : `${baseUrl}/`;
-    const base = `${globalThis.location.origin}${normalizedBase}`;
-    if (requestedUrl?.startsWith(base) && requestedUrl !== globalThis.location.href) {
+    // The payload comes from (untrusted) sessionStorage: only a string URL is usable. Anything else — e.g. a
+    // tampered `{ "url": 1 }` — must be ignored, not fed to URL()/startsWith() below where it would throw and
+    // abort the currentUser observer that called us.
+    if (typeof requestedUrl !== 'string') {
+      return;
+    }
+    // Restrict the restore to the UI base path so a saved value can't slip past into an open redirect.
+    // baseUrl is usually a path (`/nuxeo/ui/`) but the property contract also allows an absolute URL, so
+    // resolve both it and the saved URL against the current origin. Normalize the base to a single trailing
+    // '/' — without it a base of `/nuxeo/ui` would let a sibling path like `${origin}/nuxeo/ui-foo/...` pass.
+    const rawBase = this.baseUrl || '/';
+    const normalizedBase = rawBase.endsWith('/') ? rawBase : `${rawBase}/`;
+    const { origin } = globalThis.location;
+    let base;
+    let target;
+    try {
+      base = new URL(normalizedBase, origin);
+      target = new URL(requestedUrl, origin);
+    } catch (e) {
+      // Unparseable base or saved URL — discard (already consumed above) and don't navigate.
+      this._inactivityStorageError = e;
+      return;
+    }
+    // Require the saved URL to be same-origin with the window we're navigating (the real open-redirect
+    // guard) AND under the UI base path, and to differ from the current page.
+    if (target.origin === origin && target.href.startsWith(base.href) && target.href !== globalThis.location.href) {
       this._redirect(requestedUrl);
     }
   },
--- a/elements/behaviors/nuxeo-inactivity-behavior.js
+++ b/elements/behaviors/nuxeo-inactivity-behavior.js
@@ -24,6 +24,14 @@ import { config } from '@nuxeo/nuxeo-elements';
 export const INACTIVITY_ACTIVITY_KEY = 'nuxeo-ui-inactivity-last-activity';
 
 /**
+ * WEBUI-2189: sessionStorage key holding the page the user was on when an inactivity/401 logout fired, so
+ * we can return them there after they re-authenticate. sessionStorage (not localStorage) scopes it to the
+ * originating tab and clears it when that tab closes — bounding its lifetime and avoiding a cross-tab
+ * hijack. It is consumed once, on the next app boot, and validated to be same-origin before navigating.
+ */
+export const INACTIVITY_REQUESTED_URL_KEY = 'nuxeo-ui-inactivity-requested-url';
+
+/**
  * Client-side session inactivity handling (WEBUI-1987, CWE-613 Insufficient Session Expiration).
  *
  * Arms an idle timer (from the `session.timeout` config, in minutes) that logs the user out after a
@@ -232,6 +240,9 @@ export const NuxeoInactivityBehavior = {
       return Promise.resolve();
     }
     this._loggingOut = true;
+    // WEBUI-2189: remember where the user was so we can bring them back after re-authentication. Only this
+    // inactivity/401 path saves it; the manual "Sign Out" link uses _logout() directly and must not.
+    this._saveRequestedUrl();
     // We want Nuxeo's native "Your session is inactive. Please log in." message, which login.jsp only
     // renders when it receives a top-level `nxtimeout` param. We can't reach that by navigating to /logout
     // with a requestedUrl: when anonymous auth is enabled the /ui SPA bounce re-nests our value inside its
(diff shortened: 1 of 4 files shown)


test/nuxeo-inactivity-behavior.test.js
```diff
@@ -465,4 +469,73 @@ suite('nuxeo-inactivity-behavior (WEBUI-1987)', () => {
...
+
+  suite('WEBUI-2189: restore requested URL after re-login', () => {
+    const REQUESTED_URL_KEY = INACTIVITY_REQUESTED_URL_KEY; // reuse the behavior's exported key
+    let redirect;
+
+    setup(() => {
+      redirect = sinon.stub(host, '_redirect');
+      host._loggingOut = false;
+      window.sessionStorage.removeItem(REQUESTED_URL_KEY);
+    });
```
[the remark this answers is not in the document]
narasimhahyland: Good catch — fixed in b830c8b5 (cherry-picked to #3422 as 561eb2e9). The WEBUI-2189 suite `setup()` now clears `host._inactivityStorageError` before each test, and both storage-error tests assert the specific recorded error (`message === "denied"`) so they can only pass when the code path under test actually records it.

AnilKumarVanga (dismissed):

elements/behaviors/nuxeo-inactivity-behavior.js
```diff
@@ -272,6 +283,45 @@ export const NuxeoInactivityBehavior = {
...
+      this._inactivityStorageError = e;
+    }
+  },
+
+  // WEBUI-2189: after re-authentication the app reloads at its root, losing the page the user was on when
+  // the session expired. If we saved one on this tab, send them back to it. Wired from the host's ready()
+  // so it runs once per boot, before the timer/401 handlers are armed. The saved value is consumed once
+  // and must be same-origin (open-redirect protection) and different from the current page before we
+  // navigate.
+  _restoreRequestedUrlAfterLogin() {
```
vivekchaturved013, line 316: This restores unconditionally at boot rather than on a successful re-authentication. It is called from `nuxeo-app`'s `ready()`, before `currentUser` resolves.

With anonymous authentication enabled - a configuration this file already reasons about, in the `_logoutRedirect` comment about the `/ui` SPA bounce re-nesting `requestedUrl` - the SPA boots as Guest without anyone logging in. The key then gets consumed and the Guest is navigated to the previous user's deep link, most likely landing on a permission error. Worse, the saved URL is now gone, so the real re-login restores nothing and the feature silently does not work on exactly that configuration.

Gating on an authenticated, non-anonymous `currentUser` would avoid it - a one-shot flag driven off the `currentUser` observer, or firing after `#nxcon` resolves. Storing the username alongside the URL and only restoring on a match would also protect against restoring one user's page for another.

elements/behaviors/nuxeo-inactivity-behavior.js
```diff
@@ -272,6 +283,45 @@ export const NuxeoInactivityBehavior = {
...
+    } catch (e) {
+      // sessionStorage unavailable; nothing to restore.
+      this._inactivityStorageError = e;
+      return;
+    }
+    // Same-origin guard: compare against `origin + '/'`, not a bare `origin`, so a look-alike host such as
+    // https://evil-<origin-host> can't slip past the prefix check and cause an open redirect.
+    if (
+      requestedUrl &&
+      requestedUrl.startsWith(`${globalThis.location.origin}/`) &&
```
vivekchaturved013: The trailing `/` is the right call and the look-alike test is a good one.

Minor, and theoretical since only our own `location.href` is ever written here: the guard accepts any same-origin URL, so a saved non-UI path would still be honoured. Tightening the prefix to the UI base path costs nothing if you want the guard to be self-evidently safe rather than safe-by-caller.

elements/nuxeo-app.js
```diff
@@ -849,6 +849,9 @@ Polymer({
 
     this.removeAttribute('unresolved');
 
+    // WEBUI-2189: if the previous session ended via an inactivity/401 logout, return the user to the page
+    // they were on (saved on this tab) now that they have re-authenticated — before arming anything else.
+    this._restoreRequestedUrlAfterLogin();
```
vivekchaturved013: This call site is the thing that makes the feature work, and it has no test - `test/nuxeo-app.test.js` covers `ready()` behavior elsewhere, so a case asserting the call would fit.

Worth one manual check too: because the saved URL usually differs from the post-login URL only by fragment, `_redirect`'s `location.replace` performs a same-document hash navigation rather than a reload. That is the behavior you want, but it depends on the router already listening by the time `ready()` fires - otherwise the hash changes and no page transition happens.

test/nuxeo-inactivity-behavior.test.js
```diff
@@ -465,4 +469,76 @@ suite('nuxeo-inactivity-behavior (WEBUI-1987)', () => {
...
+      host._inactivityStorageError = undefined;
+      window.sessionStorage.removeItem(REQUESTED_URL_KEY);
+    });
+
+    teardown(() => {
+      redirect.restore();
+      window.sessionStorage.removeItem(REQUESTED_URL_KEY);
+    });
+
+    test('_logoutRedirect saves the current page before navigating away', async () => {
```
vivekchaturved013: Good set of cases overall - the cross-origin, look-alike, same-page, nothing-saved and both storage-throw branches are all here.

Missing is the `_loggingOut` guard: `setup()` sets `host._loggingOut = false`, so the early-return path in `_logoutRedirect` is never exercised for the save. A case with `_loggingOut = true` asserting the key is not written would cover it.

vivekchaturved013 (commented):
Review of the post-login redirect. The shape of this reads well to me: `sessionStorage` scopes the value to the originating tab, it is consumed exactly once, the same-origin guard compares against `origin + '/'` so both `https://evil-<host>` and `<origin>.evil.example.com` are rejected (nice that the look-alike case has its own test), self-redirects are skipped, and both storage error branches are covered. It also follows this file's existing convention of recording `_inactivityStorageError` rather than swallowing the exception, matching `_recordSharedActivity` and `_getLastActivity`.

One thing I would want changed before merge, plus some test-coverage gaps - details inline. Applies equally to the `maintenance-3.1.x` twin (#3422), whose diff is byte-identical.

### Test coverage gaps

- Nothing asserts that `nuxeo-app`'s `ready()` calls `_restoreRequestedUrlAfterLogin`. The wiring is the part that makes the feature work, and it is the one part with no test - `test/nuxeo-app.test.js` would be the place.
- Nothing asserts that `_logoutRedirect` skips the save when `_loggingOut` is already `true`. It does, since the early return precedes `_saveRequestedUrl()`, but it is untested.
- The comment on line 244 claims the manual Sign Out link must not save. I checked and it holds today - the Sign Out menu item is a plain link built by `_logout(url)` (`nuxeo-app.js` line 512), so it never enters `_logoutRedirect`. Since a comment asserts it, a test or a pointer to that line would keep it from regressing if sign-out is ever routed through the behavior.

### Small note

`_saveRequestedUrl` stores the full `href` including query params, so if the expiry happens on a URL that already carries something one-shot (`nxtimeout`, for instance) it will be replayed on restore. Probably harmless, but worth a thought.

No matching nuxeo-elements change is needed for this.

elements/behaviors/nuxeo-inactivity-behavior.js
narasimhahyland: Fixed in lts-2025 `2b3e6be99` (cherry-picked to maintenance-3.1.x as `a86128785`). The restore no longer runs from `ready()`; it's now triggered from `_observeCurrentUser()` once a real user resolves. `_restoreRequestedUrlAfterLogin()` returns early (without reading/consuming the key) when there is no `currentUser` or the user is anonymous (`isAnonymous` or id `Guest`), so the key survives the Guest boot for the real re-login. The saved payload now also records the owning user's id and we only navigate when it matches the current user (consume-but-don't-navigate on mismatch).

elements/behaviors/nuxeo-inactivity-behavior.js
narasimhahyland: Fixed in lts-2025 `2b3e6be99` (cherry-picked to maintenance-3.1.x as `a86128785`). The guard now requires the saved URL to start with `globalThis.location.origin + (this.baseUrl || '/')` (e.g. `/nuxeo/ui/`), which implies same-origin and is self-evidently safe. Test host has no `baseUrl` so it falls back to `'/'` and existing tests still pass.

elements/nuxeo-app.js
narasimhahyland: Fixed in lts-2025 `2b3e6be99` (cherry-picked to maintenance-3.1.x as `a86128785`). Added `test/nuxeo-app.test.js` coverage asserting that when a non-anonymous `currentUser` resolves, `_restoreRequestedUrlAfterLogin` is invoked (stubbed on the instance), plus a case asserting no restore when there is no resolved user. Since restore now runs on currentUser resolution, the router is already listening (addressing the fragment-navigation timing note).

test/nuxeo-inactivity-behavior.test.js
narasimhahyland: Fixed in lts-2025 `2b3e6be99` (cherry-picked to maintenance-3.1.x as `a86128785`). Added a test for the `_loggingOut` guard: with `host._loggingOut = true`, `await host._logoutRedirect()` takes the early-return path and the sessionStorage key is NOT written.

elements/behaviors/nuxeo-inactivity-behavior.js
```diff
@@ -272,6 +283,70 @@ export const NuxeoInactivityBehavior = {
...
+    // Never restore one user's page for another: only navigate when the saved user matches the current one.
+    if (savedUser !== currentUser.id) {
+      return;
+    }
+    // Restrict to the UI base path (e.g. `/nuxeo/ui/`, which implies same-origin) so a saved value can't
+    // slip past into an open redirect. baseUrl is absent in tests/edge cases → fall back to a bare '/'.
+    const base = `${globalThis.location.origin}${this.baseUrl || '/'}`;
+    if (requestedUrl && requestedUrl.startsWith(base) && requestedUrl !== globalThis.location.href) {
+      this._redirect(requestedUrl);
+    }
```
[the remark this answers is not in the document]
narasimhahyland: Fixed in lts-2025 `1da4ac6d0` (cherry-picked to maintenance-3.1.x as `968669d53`). The base is now normalized to always end with a single trailing `/` before the `startsWith` check: `const baseUrl = this.baseUrl || '/'; const normalizedBase = baseUrl.endsWith('/') ? baseUrl : `${baseUrl}/`;`. So a `baseUrl` of `/nuxeo/ui` no longer lets a sibling path like `${origin}/nuxeo/ui-foo/...` pass. Same-origin intent and the `requestedUrl !== current href` check are unchanged. Added tests proving the no-trailing-slash sibling look-alike is rejected (key consumed, not navigated) and that a proper `${origin}/nuxeo/ui/#!/...` URL is still restored.

test/nuxeo-app.test.js
```diff
@@ -1431,6 +1431,30 @@ suite('nuxeo-app', () => {
       expect(app.$.tasksProvider.params).to.deep.equal({ userId: 'user-1' });
     });
 
+    // WEBUI-2189: the post-login restore is driven by the currentUser observer (not ready()), so it runs
+    // once a real user has authenticated and the router is already listening.
+    test('_observeCurrentUser triggers the post-login URL restore when a real user resolves', () => {
+      const restore = sinon.stub(app, '_restoreRequestedUrlAfterLogin');
+      try {
+        app.currentUser = { id: 'jdoe', isAnonymous: false, properties: {} };
+        app._observeCurrentUser();
```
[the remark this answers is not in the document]
narasimhahyland: Good catch — fixed in lts-2025 `11a1e297d` (cherry-picked to maintenance-3.1.x #3422 as `5a7cbd80a`). `currentUser` is observer-backed, so assigning it already invokes `_observeCurrentUser()`; the explicit call fired the restore twice and made the `calledOnce` assertion test double-invocation instead of the observer wiring. Removed the manual invocation so the test now asserts the assignment alone drives the restore exactly once. (The sibling "no resolved user" test keeps its direct call — its `not.called` assertion is immune to double-invocation and it deliberately unit-tests the guard.)

elements/behaviors/nuxeo-inactivity-behavior.js
```diff
@@ -272,6 +283,74 @@ export const NuxeoInactivityBehavior = {
...
   },
 
+  // WEBUI-2189: persist the current page (full hashbang URL) before an inactivity/401 logout navigation,
+  // so _restoreRequestedUrlAfterLogin() can return the user here once they log back in. The owning user's
+  // id is stored alongside the URL so restore only ever returns a page to the same user who left it.
+  _saveRequestedUrl() {
+    try {
+      const payload = JSON.stringify({
+        url: globalThis.location.href,
+        user: this.currentUser?.id,
```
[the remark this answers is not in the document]
narasimhahyland: Good edge case — fixed in lts-2025 `3f1c5b48f` (cherry-picked to maintenance-3.1.x as `9a2f9fbb6`). `_saveRequestedUrl()` now returns early when `this.currentUser?.id` is absent, so we never write an unrestorable `{ user: undefined }` payload that would be consumed but never navigated. Added a test asserting no `sessionStorage.setItem` when there is no resolved user.

elements/nuxeo-app.js
```diff
@@ -1509,6 +1509,12 @@ Polymer({
 
   _observeCurrentUser() {
     if (this.currentUser) {
+      // WEBUI-2189: run the post-login restore here rather than from ready(): with anonymous auth enabled
+      // the app boots as Guest before a real user resolves, and restoring from ready() would consume the
+      // saved page for the wrong (or no) user. By the time a real currentUser resolves the router is
+      // listening, so navigating to the saved deep link works. The restore method itself skips anonymous
+      // users and only navigates for the user who saved the page.
+      this._restoreRequestedUrlAfterLogin();
```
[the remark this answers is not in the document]
narasimhahyland: Agreed, and thanks for confirming it is safe — the key is consume-once so repeat observer fires are cheap no-ops, hence no code change. Restore is intentionally driven off the `currentUser` observer (rather than a one-shot) so it also covers the anonymous→real-user boot transition; leaving as-is for that reason.

vivekchaturved013 (approved):

naveen-konda (approved):

madhurkulshrestha-hyland (approved):

PR #3458, opened 2026-08-20
WEBUI-1491: Fix nuxeo-diff-page header when comparing doc with long title [LTS-2025]

AnilKumarVanga:
https://hyland.atlassian.net/browse/WEBUI-1491


madhurkulshrestha-hyland (dismissed):

narasimhahyland (changes requested):
Text is cutting at end

PR #3484, opened 2026-08-31
WEBUI-1786: add an administrator announcement banner

narasimhahyland:
## Problem

Administrators have no way to tell every Web UI user about something that affects the whole
instance — planned downtime, an upgrade window, a policy change. The only options are out-of-band
channels such as email, which miss exactly the users who are in the application at the time.

WEBUI-1786 asks for a banner in the header that administrators can turn on and off, whose text they
can edit, and to which they can attach an optional link.

## Root cause

Not a defect — this is new functionality. There was no place in Web UI to author an instance-wide
message, and no server-side store the UI could write an app-wide setting to (`Nuxeo.UI.config` is
generated once per page load from `nuxeo.conf` and is read-only at runtime).

## Changes

**Storage (Web UI addon, XML only — no new Java)**

* `plugin/.../OSGI-INF/webui-announcement-contrib.xml` + `schemas/webui-announcement.xsd` contribute a
  `webUIAnnouncement` directory holding a single entry (id `announcement`) with `enabled`, `message`,
  `linkUrl` and `linkLabel`.
* A directory rather than a document keeps the value out of navigation and search, and gives the
  security the ticket asks for straight from the platform defaults: **every authenticated user can
  read it, only administrators can write it**. Verified against a real server — a `members` user gets
  `403 Not allowed to edit directory` on `PUT`, so the restriction does not depend on the client-side
  filter that hides the screen.
* `<types><type>system</type></types>` keeps it out of the Vocabularies administration screen.

**Administration screen**

* `elements/nuxeo-admin/nuxeo-announcement-management.js` — toggle, message, link and link text, with a
  single Save. Registered as `ADMINISTRATION_MENU` / `ADMINISTRATION_PAGES` slot content at `order=70`,
  wrapped in `user.isAdministrator`. No routing change was needed: `page('/admin/:tab?')` already
  covers `#!/admin/announcement`.
* Creates the entry on first save and updates it afterwards (`POST` on a duplicate id returns 409).

**Banner**

* `elements/nuxeo-app/nuxeo-announcement-banner.js`, mounted in `<header role="banner">` in
  `nuxeo-app`. It looks the announcement up once a user is connected, and refreshes on the
  `announcement-updated` event so an administrator sees the result of Save without reloading.
* It is laid out **in the normal flow**, so it pushes the application down instead of covering the top
  of the page, and republishes its own height as `--nuxeo-app-top` — the variable the drawer, logo,
  menu, page toolbars and result views already use to reserve vertical space, since those are viewport
  positioned and cannot be pushed by the flow. A `ResizeObserver` keeps it correct when the message
  wraps on a narrow viewport. With no announcement enabled the element is `display: none` and
  `--nuxeo-app-top` is untouched, so the layout is byte-for-byte what it is today.
* If the lookup fails (e.g. a server whose addon predates the directory) the banner stays hidden and
  nothing is reported to the user — it must never break the application shell.

**Safety**

* The message is rendered as text and the link through a protocol check (`http:`/`https:` only), so an
  announcement cannot inject markup or a `javascript:` URL.
* Free-text fields are capped at the 255 characters the directory columns hold. The two link fields
  use `maxlength` (`nuxeo-input` forwards it to the native input); `nuxeo-textarea` has no
  `maxlength` support, so the message is truncated to the limit in `_messageChanged` and a counter
  under the field shows it. `_validate` keeps a length check as a backstop. This matters beyond
  tidiness: while an over-long message could sit in the field, the length check prevented an
  administrator from turning the banner *off* at all (fixed in a74618515).

**i18n / tests**

* 15 new `announcement*` keys plus `app.title.admin.announcement` in `i18n/messages.json`.
* 24 unit tests (`test/nuxeo-announcement-banner.test.js`, `test/nuxeo-announcement-management.test.js`).
* Gherkin scenarios + page objects + step definitions (`ftest/features/announcement.feature`).

## Test plan

Local gate (both branches): `npm run lint` clean, `npm test` **2678 passed / 0 failed** (2654 before,
+24 new).

Verified end to end against a throwaway `nuxeo/nuxeo:2025` container with this branch's build
deployed and the contribution loaded (component status `Pending: 0 / Missing: 0`):

| Check | Result |
|---|---|
| `Announcement` appears in the administration menu for an administrator | pass |
| Save with toggle on → banner appears with no reload, `--nuxeo-app-top: 46px` | pass |
| Banner still there on another page and after a full browser reload | pass |
| A `members` user sees the banner, has no administration icon, cannot open `#!/admin/announcement` | pass |
| `PUT` on the directory as a `members` user | `403` |
| Save with toggle off → banner gone, `--nuxeo-app-top` back to `0px` | pass |
| Optional link renders, opens in a new tab, `rel="noopener noreferrer"` | pass |
| Long message at 1400px / 420px wide → wraps, `--nuxeo-app-top` follows (46px / 134px), no page overflow | pass |
| RTL (`dir="rtl"`) → banner mirrors, no clipping, no overflow | pass |
| `javascript:` link | dropped, banner still shows the message |
| Message > 255 chars | truncated in the field (real keyboard input: 280 typed → 255 held), counter shows `255 of 255 characters` |
| Turning the banner **off** after an over-long message was typed | saves (`PUT`), banner hides, server holds `enabled=false` — the trap this used to hit is gone |
| Removing a previously configured link | anchor is fully removed (`restamp`), no hidden element left holding the old href |
| Markup in the message (``) | escaped as text, no element injected |
| Patched UI against a server whose addon predates the directory (`404`) | banner stays hidden, no toast, no layout shift, app fully usable |
| Only an unrelated directory entry exists (no reserved id) | banner hidden and admin form empty; saving creates the reserved entry with `POST` |
| Skip link focused while a banner is displayed | paints above the banner and is the topmost element at its centre |
| Offline banner while an announcement is displayed | stacks directly below it (y 46→99), no overlap |
| Exact directory column boundary | 255 chars `201`, 256 chars `400`, `PUT` on a missing id `404` (no silent data loss) |

Screenshots and before/after recordings are attached to WEBUI-1786.

## Notes

* Out of scope per the ticket: scheduling, per-group targeting, multiple simultaneous banners.
* Six existing screens size an inner table with a fixed `calc(100vh - Npx)` that does not account for
  `--nuxeo-app-top` (cloud consumers/providers, audit search, authorized apps, user cloud services,
  picturebook, default results). While a banner is displayed those tables are ~46px taller than their
  container and get an inner scrollbar. Cosmetic only, and pre-existing for the offline/expired-session
  banners; left alone to keep this change scoped.
* The ticket's `fixVersions` lists `3.1.x` only; opened on both bases so `lts-2025` does not regress
  when the maintenance line is merged forward.


packages/nuxeo-web-ui-ftest/pages/ui/admin/announcement.js
```diff
@@ -0,0 +1,58 @@
...
+    const checked = await toggle.getProperty('checked');
+    if (checked !== enabled) {
+      await toggle.click();
+    }
+  }
+
+  async fillMessage(message) {
+    const field = await this.messageInput;
+    await field.waitForVisible();
+    const textarea = await field.$('paper-textarea');
```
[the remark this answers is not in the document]
narasimhahyland: Fixed in 25ffa7908 — the page object now targets `#textarea`, the selector the functional-test field registry uses for `nuxeo-textarea`, instead of the `paper-textarea` wrapper.

packages/nuxeo-web-ui-ftest/pages/ui/admin/announcement.js
```diff
@@ -0,0 +1,58 @@
...
+    await textarea.setValue(message);
+  }
+
+  async fillLink(linkUrl, linkLabel) {
+    if (linkUrl !== undefined) {
+      const field = await this.linkUrlInput;
+      await field.waitForVisible();
+      await (await field.$('paper-input')).setValue(linkUrl);
```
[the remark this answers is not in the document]
narasimhahyland: Fixed in 25ffa7908 — both link fields now go through `.input-element input`, matching the `nuxeo-input` entry in the field registry.

elements/nuxeo-app/nuxeo-announcement-banner.js
```diff
@@ -0,0 +1,257 @@
...
+    const linkLabel = typeof entry.linkLabel === 'string' ? entry.linkLabel.trim() : '';
+    this._message = message;
+    this._linkUrl = linkUrl;
+    this._linkLabel = linkUrl ? linkLabel || this.i18n('announcementBanner.moreDetails') : '';
+    this._opened = true;
+  },
+
+  _entryOf(response) {
+    const entries = (response && response.entries) || [];
+    const entry = entries.find((e) => e && e.id === ANNOUNCEMENT_ENTRY_ID) || entries[0];
```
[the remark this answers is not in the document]
narasimhahyland: Agreed, fixed in 25ffa7908 — `_entryOf` now only accepts the entry whose id is `ANNOUNCEMENT_ENTRY_ID` and returns `null` otherwise. Covered by a new test that feeds an entry with a different id and asserts the banner stays hidden.

elements/nuxeo-admin/nuxeo-announcement-management.js
```diff
@@ -0,0 +1,285 @@
...
+   * Loads the current announcement.
+   * @return {Promise} resolved once the form reflects the server state.
+   */
+  refresh() {
+    this.$.announcement.path = ANNOUNCEMENT_ENTRY_PATH;
+    this.$.announcement.data = null;
+    return this.$.announcement.get().then(
+      (response) => {
+        const entries = (response && response.entries) || [];
+        const entry = entries.find((e) => e && e.id === ANNOUNCEMENT_ENTRY_ID) || entries[0];
```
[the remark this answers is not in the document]
narasimhahyland: Agreed, fixed in 25ffa7908 — `refresh()` now looks up the reserved id only, so `_exists` stays `false` and the form stays empty when the announcement is absent. New test added.

elements/nuxeo-app/nuxeo-announcement-banner.js
```diff
@@ -0,0 +1,257 @@
...
+    if (!this.user) {
+      return Promise.resolve();
+    }
+    return this.$.announcement.get().then(
+      (response) => this._update(response),
+      // A server without the announcement directory (or an unreachable one) simply has no
+      // announcement to show: stay hidden rather than failing the whole application shell.
+      () => this._update(null),
+    );
+  },
```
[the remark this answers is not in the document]
narasimhahyland: Good catch, fixed in 25ffa7908 — every lookup takes a `_requestId`; a response is applied only while it is still the newest one and the element is attached with a connected user. `_requestId` is also bumped when `user` becomes null and on `detached()`. Three tests added: out-of-order completion, logout during a pending lookup, and detach during a pending lookup.

elements/nuxeo-admin/nuxeo-announcement-management.js
```diff
@@ -0,0 +1,285 @@
...
+              error-message="[[_messageError]]"
+            ></nuxeo-textarea>
+
+            <nuxeo-input
+              class="field"
+              id="linkUrl"
+              name="linkUrl"
+              type="url"
+              label="[[i18n('announcementManagement.linkUrl')]]"
+              placeholder="https://example.com/maintenance"
```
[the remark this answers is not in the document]
narasimhahyland: Fixed in 25ffa7908 — added `announcementManagement.linkUrl.placeholder` to `i18n/messages.json` and bound the attribute through `i18n(...)`.

i18n/messages.json
```diff
@@ -42,6 +42,20 @@
   "addToCollectionDocumentsButton.tooltip": "Add to Collection",
   "admin.usersAndGroups.heading": "Users & Groups",
   "analytics.after": "After",
+  "announcementBanner.moreDetails": "More details",
```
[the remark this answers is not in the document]
narasimhahyland: Fixed in 25ffa7908 — the whole `announcement*` block now sits after the last `analytics.*` key, so the file is alphabetically grouped again.

elements/nuxeo-admin/nuxeo-announcement-management.js
```diff
@@ -0,0 +1,302 @@
...
+              class="field"
+              id="message"
+              name="message"
+              rows="3"
+              label="[[i18n('announcementManagement.message')]]"
+              value="{{_entry.message}}"
+              required$="[[_entry.enabled]]"
+              disabled$="[[_loading]]"
+              invalid="[[_messageInvalid]]"
+              error-message="[[_messageError]]"
```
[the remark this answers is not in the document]
narasimhahyland: Thanks — the gap is real, but I did not apply the suggested fix as written, because `maxlength` on `<nuxeo-textarea>` would be a dead attribute. I checked the component and then verified it in a browser against a running instance:

- `'maxlength' in nuxeoTextareaEl` → `false` (the component declares no such property, and its template does not forward one to `paper-textarea`)
- host attribute after binding → `null`
- `maxlength` on the native `<textarea>` reached through the shadow trees → `null`
- typing 300 characters → the message field accepted all 300, while `#linkUrl` stopped at 255 (`nuxeo-input` *does* declare `maxlength` and forwards it, so those two fields were genuinely capped)

So the diagnosis was right and the remedy would have silently done nothing.

Fixed properly in a74618515 (and 16b4ec0be on `maintenance-3.1.x`) by applying the limit where it can be applied:

- `_messageChanged` truncates `_entry.message` to `ANNOUNCEMENT_MAX_LENGTH`, which is exactly what `maxlength` does natively on the two link fields (a longer paste is truncated), so the field can never hold more than the directory column;
- a counter under the field (`255 of 255 characters`) surfaces the limit instead of only revealing it at Save;
- a pending validation error is cleared as soon as the message is edited.

Re-verified with real keyboard input through the browser (not a synthetic `input` event, which does not reach this component's binding): typing 280 characters leaves `_entry.message` at 255, the native textarea at 255, and the counter at `255 of 255 characters`.

This also removes the more serious consequence you and the other reviewer both flagged — see the sibling thread on PR #3485 — so `_validate` no longer has an over-long value to reject through the UI; the check stays as a backstop for a value that never went through the observer, and is covered by its own test.

narasimhahyland:
### Note on the current red checks — not caused by this PR

`unit-test` and `Build and analyze` are failing here for an environmental reason that affects every PR in the repo right now, not because of this change.

Both jobs fail at the dependency step with:

```
::error::Could not resolve a common RC version across all @nuxeo packages
```

`Build and analyze` exits 1 there. `unit-test` continues with a mismatched `@nuxeo/*` version and then reports **exactly one** failing test:

```
❌ nuxeo-search-form > gives the saved searches dropdown an always visible label (WEBUI-489)
   AssertionError: expected null to equal 'Search Filters'
```

Evidence that it is not this branch:

- **Unrelated PRs fail identically.** #3504 and #3505 (`task-avoid-optional-chaining-skill`, no overlap with these files) fail with the same single test and the same `common RC version` error.
- **All of this PR's tests pass in CI.** The run executed 2693 tests, `2692 passed, 1 failed`, and the one failure is the `nuxeo-search-form` test above. Zero announcement tests failed.
- **The previous commit on this same branch was green** (`unit-test` SUCCESS on 4d1ae2a0c), and the only changes since are in `nuxeo-announcement-*`, `i18n/messages.json` and their tests.
- Locally the full suite is **2693 passed / 0 failed**, because the local checkout resolves `@nuxeo/nuxeo-ui-elements` at the version that contains the WEBUI-489 label change that this test asserts.

So the failing assertion is a `@nuxeo` package-version resolution problem in CI, not a regression here. Happy to rebase or re-run once the RC resolution is fixed — flagging it rather than papering over it, since touching that unrelated test from this PR would be wrong.

elements/nuxeo-admin/nuxeo-announcement-management.js
```diff
@@ -0,0 +1,336 @@
...
+/**
+`nuxeo-announcement-management`
+
+Administration screen letting an administrator turn the announcement banner on or off, edit its
+message and attach an optional link. The value is shared by every user of the instance.
+
+@group Nuxeo UI
+@element nuxeo-announcement-management
+*/
+Polymer({
```
madhurkulshrestha-hyland: Use Polymer 3 class syntax

elements/nuxeo-app/nuxeo-announcement-banner.js
```diff
@@ -0,0 +1,281 @@
...
+
+While the banner is displayed it publishes its own height as `--nuxeo-app-top`
+on the document root, which is the variable the drawer, the page toolbars and
+the result views already use to reserve vertical space. That way the banner
+pushes the application down instead of covering it.
+
+@group Nuxeo UI
+@element nuxeo-announcement-banner
+*/
+Polymer({
```
madhurkulshrestha-hyland: Polymer 3 syntax to be used

packages/nuxeo-web-ui-ftest/pages/ui/announcementBanner.js
```diff
@@ -0,0 +1,48 @@
+/**
+ * The announcement banner lives inside the `nuxeo-app` shadow root and renders its message in its
+ * own shadow root, so its state is read through the browser rather than with plain selectors.
+ */
+export default class AnnouncementBanner {
```
madhurkulshrestha-hyland, line 5: Shouldnt the message take in a complete html text with inline links in multiple areas that can be added in the banner ?

madhurkulshrestha-hyland (changes requested):

elements/nuxeo-admin/nuxeo-announcement-management.js
```diff
@@ -0,0 +1,336 @@
...
+          }`,
+        });
+      },
+    );
+  },
+
+  _save() {
+    if (this._loading || !this._validate()) {
+      return Promise.resolve();
+    }
```
vivekchaturved013, line 301: `_save()` only bails on `_loading`, which is never set while a save is in flight, so the Save button stays enabled for the whole request.

`nuxeo-resource` aborts its own previous request whenever a new one starts on the same instance — it replaces `_controller` on every `execute()` unless `uncancelable` is set. So a double-click on Save aborts the first request: the second succeeds, but the first rejects with an `AbortError`, and `_persist` reports it to the administrator as `ERROR: The user aborted a request.` A failure toast on a save that actually worked.

Suggest a dedicated `_saving` flag set here and cleared in both `_persist` handlers, added to the `#save` button's `disabled$`.

elements/nuxeo-admin/nuxeo-announcement-management.js
```diff
@@ -0,0 +1,336 @@
...
+
+  /**
+   * Loads the current announcement.
+   * @return {Promise} resolved once the form reflects the server state.
+   */
+  refresh() {
+    this.$.announcement.path = ANNOUNCEMENT_ENTRY_PATH;
+    this.$.announcement.data = null;
+    this._loading = true;
+    return this.$.announcement.get().then(
```
vivekchaturved013, line 269: Same re-entrancy issue on the read side, and it also undercuts the `_loading` doc comment above.

`_visibleChanged` calls `refresh()` unconditionally and `iron-pages` re-asserts `visible` on selection, so clicking away to another admin tab and back while the GET is in flight starts a second one — which aborts the first, whose rejection shows a spurious `ERROR: The user aborted a request.` toast. If instead both complete, the older response can land last and overwrite what the administrator typed after the first one re-enabled the form — exactly what the `_loading` comment claims to prevent.

`nuxeo-announcement-banner` already has the right pattern (monotonic `_requestId` + `_applyIfCurrent`); worth reusing it here.

Minor: line 233 is dead — `nuxeo-resource` ignores `data` on GET (`if (this.data && this.method !== 'get')`).

elements/nuxeo-admin/nuxeo-announcement-management.js
```diff
@@ -0,0 +1,336 @@
...
+        // Let the banner of the current session pick the new value up without a reload.
+        document.dispatchEvent(new CustomEvent(ANNOUNCEMENT_UPDATED_EVENT));
+      },
+      (err) => {
+        this.notify({
+          message: `${this.i18n('label.error').toUpperCase()}: ${
+            err?.message || this.i18n('announcementManagement.errorSaving')
+          }`,
+        });
+      },
```
vivekchaturved013, line 338: `err?.message || this.i18n(...)` means the i18n string is only used when the error carries no message — i.e. almost never — so real failures show untranslated server or browser text instead of `announcementManagement.errorSaving`. Same shape at line 255 for `errorLoading`.

Suggest inverting it: notify with the i18n message, `console.warn` the raw error so it stays diagnosable, and skip the toast entirely when `err.name === 'AbortError'`.

elements/nuxeo-app/nuxeo-announcement-banner.js
```diff
@@ -0,0 +1,281 @@
...
+    this._linkLabel = linkUrl ? linkLabel || this.i18n('announcementBanner.moreDetails') : '';
+    this._opened = true;
+  },
+
+  _entryOf(response) {
+    // The announcement is the entry with the reserved id. Any other entry of the directory belongs
+    // to nobody and must never be promoted to an instance wide banner.
+    const entry = (response?.entries || []).find((e) => e?.id === ANNOUNCEMENT_ENTRY_ID);
+    return entry?.properties || null;
+  },
```
vivekchaturved013: This is byte-identical to the lookup in `nuxeo-announcement-management.refresh()` (line 239). `nuxeo-announcement.js` already exists as the shared contract between the two, so this belongs there as e.g. `findAnnouncementEntry(response)` — both to keep the reserved-id rule in one place and because Sonar flags duplicated blocks.

elements/nuxeo-app/nuxeo-announcement.js
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+/** Directory holding the announcement. */
+export const ANNOUNCEMENT_DIRECTORY = 'webUIAnnouncement';
+
+/** Id of the one and only entry of that directory. */
+export const ANNOUNCEMENT_ENTRY_ID = 'announcement';
+
+/** REST path (relative to the API root) used to read/create the announcement. */
+export const ANNOUNCEMENT_ENTRY_PATH = `directory/${ANNOUNCEMENT_DIRECTORY}`;
```
vivekchaturved013: This is the directory root, so the banner GETs every row and filters client-side. `directory/webUIAnnouncement/announcement` returns just the entry and 404s when it's absent — which the banner's error path already handles by staying hidden. Cheaper, and it removes the need for the reserved-id filter on the read path altogether.

Worth pairing with a cache on the contribution: as written this is an uncached directory read on the critical path of every page load for every authenticated user, even when no announcement is configured.

elements/nuxeo-app/nuxeo-announcement-banner.js
```diff
@@ -0,0 +1,281 @@
...
+       * chrome (drawer, logo, menu, toolbars) is viewport positioned and cannot be pushed that way,
+       * so the height is also published as --nuxeo-app-top, which that chrome already honours.
+       * The z-index stays below the skip link so the skip link remains visible when focused.
+       */
+      :host {
+        display: none;
+        position: relative;
+        z-index: 999;
+        background: var(--nuxeo-announcement-banner-background, #fee066);
+        color: var(--nuxeo-announcement-banner-text, #3a3a54);
```
vivekchaturved013: These are the same two colours as `nuxeo-offline-banner` (`#fee066` on `#3a3a54`), so the two banners are visually indistinguishable: a user seeing a yellow bar can't tell "you are offline" from "maintenance on Saturday", and when both are showing they stack as two identical yellow bars. Probably a quick design confirmation rather than a code change.

Contrast itself is fine (~8.4:1, AAA). Note also that neither banner's custom properties are defined in any theme, so this stays bright yellow under `dark` / `hyland-dark`.

plugin/web-ui/addon/src/main/resources/OSGI-INF/webui-announcement-contrib.xml
```diff
@@ -0,0 +1,32 @@
...
+      <schema>webUIAnnouncement</schema>
+      <table>webui_announcement</table>
+      <idField>id</idField>
+      <autoincrementIdField>false</autoincrementIdField>
+      <createTablePolicy>on_missing_columns</createTablePolicy>
+      <!-- `system` keeps the directory out of the Vocabularies administration screen -->
+      <types>
+        <type>system</type>
+      </types>
+    </directory>
```
vivekchaturved013, line 31: No cache is declared, so every banner lookup is a DB round trip — and since the banner reads on every page load for every user, this sits on a hot path. A `cacheEntryName` / `cacheMaxSize` here would be cheap insurance at scale.

Also worth flagging for whoever owns the platform side: this adds a new `webui_announcement` table, i.e. a DB schema addition on upgrade, shipped from the Web UI addon rather than the platform. Given the feature is landing on the maintenance lines, an explicit OK on using a `system` directory as the store for app-wide UI config seems worth having on the ticket.

packages/nuxeo-web-ui-ftest/features/step_definitions/announcement.js
```diff
@@ -0,0 +1,59 @@
...
+  await this.ui.administration.goToAnnouncement();
+});
+
+Given('the announcement banner is turned off', async function () {
+  const page = await this.ui.administration.announcement;
+  await page.waitForVisible();
+  await page.setEnabled(false);
+  await page.save();
+  await AnnouncementBanner.waitForDisplayed(true);
+});
```
vivekchaturved013, line 49: The announcement is persistent server-side state and there's no `After` hook, so a scenario that fails between enabling and disabling leaves the banner on for every later feature in the run. Because it also sets `--nuxeo-app-top` globally, that shifts every page by ~46px and can cascade into unrelated features' selectors — and with `maxInstances: 1` the entire rest of the suite runs after it.

An `After` hook that force-disables, ideally via a direct REST call rather than driving the UI, would make this self-healing.

vivekchaturved013 (commented):
Review of the announcement banner feature. Posting as a comment rather than a change request — nothing here blocks the concept, and the two real bugs are both in the admin screen's request handling.

Overall this is solid: the banner has a proper monotonic request guard that it invalidates on *every* abandon path (disconnect, detach), the link goes through a protocol allowlist, the entry id is a hard-coded constant so there is no unencoded-id problem, and the tests deliberately cover the stale-response and error branches rather than just the happy path. The test plan in the description is unusually thorough and the out-of-scope notes are honest.

### Worth fixing before merge

Both in `nuxeo-announcement-management.js`, and both stemming from load and save sharing one cancellable `nuxeo-resource` with no re-entry guard:

1. Double-clicking Save, or leaving and re-entering the admin tab mid-request, makes the second request abort the first — and the abort is reported to the administrator as `ERROR: The user aborted a request.`, i.e. a failure toast on an operation that actually succeeded.
2. `err?.message || this.i18n(...)` means the new `errorLoading` / `errorSaving` keys are almost never used; real failures surface untranslated server and browser text instead.

### Judgement calls (inline)

Reading the whole directory instead of the single entry (plus no directory cache, on a per-page-load hot path), the entry lookup duplicated between the banner and the admin screen, the banner being pixel-identical to the offline banner, and the ftests leaving persistent server state behind when a scenario fails.

### One process point

The Notes section documents that six existing screens using a hard-coded `calc(100vh - Npx)` will overflow by ~46px whenever a banner is displayed. Leaving that out of scope is the right call, but a PR body isn't a tracker — that should be a linked follow-up ticket, otherwise it's lost on merge. It's also slightly worse than "pre-existing for the offline banner", since the offline banner is transient whereas an announcement can be on for days.

### Regression surface

The part I'd test hardest is the `--nuxeo-app-top` write, since it's a global mutation on `document.documentElement` consumed by ~20 files: drawer and its scroll region, logo, menu, suggester, selection toolbar, `nuxeo-page`, `nuxeo-results`, document tree, search form, collections, favourites, clipboard, tasks, publications, trash content and picturebook. With no announcement enabled the host is `display: none` and the variable is never written, so the default layout is genuinely untouched.

---
<sub>Reviewed with AI assistance (Cursor); findings verified against the branch locally.</sub>

elements/nuxeo-admin/nuxeo-announcement-management.js
narasimhahyland: Addressed in 1bfe1284: the administration component now uses Polymer 3 class syntax (`class ... extends ...` plus `customElements.define`).

elements/nuxeo-app/nuxeo-announcement-banner.js
narasimhahyland: Addressed in 1bfe1284: the banner now uses Polymer 3 class syntax (`class ... extends ...` plus `customElements.define`).

packages/nuxeo-web-ui-ftest/pages/ui/announcementBanner.js
narasimhahyland: No code change needed: WEBUI-1786 defines editable text plus one optional URL/label. The message is intentionally rendered as text so administrator input cannot inject markup; the separate URL and label fields provide the supported safe link.

elements/nuxeo-admin/nuxeo-announcement-management.js
narasimhahyland: Addressed in 94a6cb8f: `_saving` now contributes to `_busy`, disables the controls, and makes `_save()` return while a save is in flight. Unit coverage verifies that a second save does not start another request.

elements/nuxeo-admin/nuxeo-announcement-management.js
narasimhahyland: Addressed in 94a6cb8f and 1bfe1284: refreshes use a monotonic request id, refresh is skipped during a save, superseded/aborted loads do not notify, and the GET now targets the reserved entry directly without a request body.

elements/nuxeo-admin/nuxeo-announcement-management.js
narasimhahyland: Addressed in 94a6cb8f: user notifications now always use the localized loading/saving keys, raw errors are retained in `console.warn`, and `AbortError` is suppressed.

elements/nuxeo-app/nuxeo-announcement-banner.js
narasimhahyland: Addressed in 1bfe1284: both consumers now request the reserved announcement entry directly, removing the duplicated whole-directory lookup and filter.

elements/nuxeo-app/nuxeo-announcement.js
narasimhahyland: Addressed in 1bfe1284: the banner reads `directory/webUIAnnouncement/announcement` directly, and the directory contribution now declares bounded entry caches.

elements/nuxeo-app/nuxeo-announcement-banner.js
narasimhahyland: No correctness change needed here: the announcement is distinguished by its information icon and content, its contrast passes, and both colors remain theme-overridable custom properties. A different palette would be a separate design decision.

plugin/web-ui/addon/src/main/resources/OSGI-INF/webui-announcement-contrib.xml
narasimhahyland: Addressed in 1bfe1284: the contribution now declares entry and without-references caches, each bounded to one entry. The single `system` directory remains the documented storage architecture for this app-wide setting.

packages/nuxeo-web-ui-ftest/features/step_definitions/announcement.js
narasimhahyland: Addressed in a2c6903e: the tagged `Before` hook snapshots the original server-side entry and an ordered `After` hook restores it (or deletes the test entry) through the directory client, even when a scenario fails.

PR #3531, opened 2026-09-08
WEBUI-2253: Complete an incomplete assertion and make two silent test skips honest [maintenance-3.1.x]

vivekchaturved013:
## Problem

Three SonarCloud maintainability findings, all in test code, all with the same symptom: **a test that reports green without having verified anything.** Jira: [WEBUI-2253](https://hyland.atlassian.net/browse/WEBUI-2253)

| Rule | Severity | File | Line |
| --- | --- | --- | --- |
| `javascript:S2970` | **BLOCKER** | `test/nuxeo-document-tree.test.js` | 420 |
| `javascript:S8968` | HIGH | `test/nuxeo-performance.test.js` | 403 |
| `javascript:S8968` | HIGH | `test/nuxeo-performance.test.js` | 473 |

This is 8 of 9 in the maintainability triage (see WEBUI-2246). It is the smallest ticket in the set but contains the backlog's only maintainability BLOCKER, and it lands first so that [WEBUI-2252](https://hyland.atlassian.net/browse/WEBUI-2252) (the `performance.timing` rewrite) starts from tests that honestly report their own status.

## Root cause

**S2970 — an assertion that can never fail.** `test/nuxeo-document-tree.test.js:420` ended the "Tree should collapse when clicking on a document" test with a chai chain that terminates before it asserts:

```js
expect(isElementVisible(documentTree.shadowRoot.querySelector('.parents')));
```

`expect(value)` with no matcher just constructs an Assertion object and discards it. The line executes and counts toward coverage, but no outcome is ever checked — it is strictly worse than a missing test, because it reads as coverage.

The adjacent "Tree breadcrumb is present" test at `:344` had the **same defect in an even weaker form** — a bare call whose return value is dropped on the floor:

```js
isElementVisible(documentTree.shadowRoot.querySelector('.parents'));
```

Sonar could not see that one because S2970 only matches `expect()` chains, and eslint could not either because `no-unused-expressions` is deliberately off for test files. Both lines sit directly under the comment `// assert that we have one parent only and it is visible`, which states the intent the code failed to implement.

**S8968 — two silent skips.** `test/nuxeo-performance.test.js:403` and `:473` both bail out with a bare `return` when `PerformancePaintTiming` is unavailable. Mocha treats a returning test as a **pass**, so on any engine lacking the paint-timing API these two report green having executed nothing.

## Changes

* **`test/nuxeo-document-tree.test.js`** (lines 344, 420): both `.parents` checks now assert `.to.be.true`, matching the intent stated in their own comment. What they now assert: *after the tree collapses to a single ancestor, the `.parents` breadcrumb container is actually rendered and visible* (non-zero box, non-hidden, opacity > 0). `.parents` is bound to `hidden$="[[_noPermission]]"` in `nuxeo-document-tree.js`, so this is a real guard against the breadcrumb silently disappearing.
* **`test/nuxeo-performance.test.js`** (lines 403, 473): the bare `return` is replaced by `this.skip()`, which reports a skip and makes the condition visible in the CI test report. Both callbacks were converted from arrow functions to `function ()` so that `this` is the Mocha context — `this.skip` is not available on an arrow function's lexical `this`.

Test code only. **No product code is touched**, so there is nothing for QA to exercise — this ticket has no QA sub-task and is verified entirely by CI.

## Test plan

- [x] `npm run lint` — eslint clean on both changed files. (Prettier reports repo-wide differences on this Windows checkout because `core.autocrlf=true` writes CRLF; re-running with `--end-of-line crlf` reports **zero** differences on the changed files, and untouched files such as `webpack.config.js` are flagged identically. CI runs on Linux with LF and is unaffected.)
- [x] `npm test` — **2736 passed, 1 failed**. The single failure is `nuxeo-search-form > gives the saved searches dropdown an always visible label (WEBUI-489)`, which is **pre-existing on `lts-2025` and unrelated**: re-running the full suite with the base versions of both changed files reproduces the identical result (2736 passed, same single failure).
- [x] **The completed S2970 assertion was seen to fail against deliberately broken behaviour** (AC 2). With `_noPermission = true` forced on a real `nuxeo-document-tree` fixture — the actual condition that hides `.parents` — a throwaway probe confirmed all three of:
  - `isElementVisible(parents)` is genuinely `false` (the breadcrumb really is hidden);
  - the **old** form `expect(isElementVisible(parents))` does **not** throw — proving it could never fail;
  - the **new** form `expect(isElementVisible(parents)).to.be.true` **throws** `expected false to be true` — proving the assertion is now real.
- [x] **The S8968 tests report as skipped, not passed** (AC 3). Chromium defines `PerformancePaintTiming`, so the guard is not taken in CI; a throwaway probe replicating both shapes confirmed the runner reports `1 skipped` for the `this.skip()` form while the old `return` form was silently counted among the passes, and that `typeof this.skip === 'function'` inside the non-arrow callback.
- [x] Commits are signed and show **Verified** on GitHub, authored by the CLA-covered identity with no `cursoragent@` co-author trailer.

Both probe files were deleted before committing; the diff is 6 changed lines across the two test files and nothing else.

## Notes

* **Coverage may not move at all**, and that is expected: every one of these lines was already executing. The coverage metric was never wrong — the assertions were.
* Completing the assertions did **not** expose a defect in `nuxeo-document-tree`, so AC 4 does not apply and no follow-up Bug is raised. The breadcrumb genuinely is visible in both scenarios; the assertions pass on their own merits, not because they are trivially true.
* Line 344 is outside the three findings Sonar listed. It is included because it is the identical defect, in the same file, in the test immediately above, and fixing only the line Sonar can see would leave the ticket's own premise — a test that cannot fail — still true one test up.
* Paired PR for the other base is linked below; the two branches carry the same commit.

Paired PR: [#3530](https://github.com/nuxeo/nuxeo-web-ui/pull/3530) (`lts-2025`) - same commit, same diff.

[WEBUI-2253]: https://hyland.atlassian.net/browse/WEBUI-2253?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WEBUI-2252]: https://hyland.atlassian.net/browse/WEBUI-2252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/nuxeo-performance.test.js
+++ b/test/nuxeo-performance.test.js
@@ -398,9 +398,9 @@ suite('Performance', () => {
   });
 
   suite('branch coverage: paint and timing fallbacks', () => {
-    test('getFirstPaint returns null when no first-paint entry', () => {
+    test('getFirstPaint returns null when no first-paint entry', function () {
       if (typeof PerformancePaintTiming === 'undefined') {
-        return;
+        this.skip();
       }
       const stub = sinon.stub(performance, 'getEntriesByType').callsFake((type) => {
         if (type === 'paint') {
@@ -468,9 +468,9 @@ suite('Performance', () => {
       }
     });
 
-    test('getFirstContentfulPaint returns null when no fcp entry', () => {
+    test('getFirstContentfulPaint returns null when no fcp entry', function () {
       if (typeof PerformancePaintTiming === 'undefined') {
-        return;
+        this.skip();
       }
       const stub = sinon.stub(performance, 'getEntriesByType').callsFake((type) => {
         if (type === 'paint') {
--- a/test/nuxeo-document-tree.test.js
+++ b/test/nuxeo-document-tree.test.js
@@ -341,7 +341,7 @@ suite('nuxeo-document-tree', () => {
       expect(documentTree.parents).to.be.not.empty;
       expect(documentTree.parents).to.have.length(1);
       expect(documentTree.parents[0].uid).to.be.equal(4);
-      isElementVisible(documentTree.shadowRoot.querySelector('.parents'));
+      expect(isElementVisible(documentTree.shadowRoot.querySelector('.parents'))).to.be.true;
     });
 
     test('Tree breadcrumb is present with root document', async () => {
@@ -417,7 +417,7 @@ suite('nuxeo-document-tree', () => {
       expect(documentTree.parents).to.be.not.empty;
       expect(documentTree.parents).to.have.length(1);
       expect(documentTree.parents[0].uid).to.be.equal(4);
-      expect(isElementVisible(documentTree.shadowRoot.querySelector('.parents')));
+      expect(isElementVisible(documentTree.shadowRoot.querySelector('.parents'))).to.be.true;
     });
   });
 


narasimhahyland (approved):

naveen-konda (approved):

test/nuxeo-performance.test.js
```diff
@@ -398,9 +398,9 @@ suite('Performance', () => {
   });
 
   suite('branch coverage: paint and timing fallbacks', () => {
-    test('getFirstPaint returns null when no first-paint entry', () => {
+    test('getFirstPaint returns null when no first-paint entry', function () {
```
madhurkulshrestha-hyland, line 401: Removal of arrow functions not required

test/nuxeo-performance.test.js
```diff
@@ -468,9 +468,9 @@ suite('Performance', () => {
       }
     });
 
-    test('getFirstContentfulPaint returns null when no fcp entry', () => {
+    test('getFirstContentfulPaint returns null when no fcp entry', function () {
```
madhurkulshrestha-hyland, line 471: Removal of arrow functions not required

madhurkulshrestha-hyland (changes requested):

test/nuxeo-performance.test.js
vivekchaturved013: The conversion is required - with an arrow function `this.skip()` throws instead of skipping.

Arrow functions bind `this` lexically, and at the top level of an ES module `this` is `undefined`, so `this.skip` is unreachable. Running both shapes through this repo's Mocha (11.7.6) under the TDD UI, with the precondition deliberately unmet so the guard is actually taken:

```
  this.skip() and the callback form
    1) ARROW callback
    - FUNCTION callback

  0 passing, 1 pending, 1 failing

  1) ARROW callback:
     TypeError: Cannot read properties of undefined (reading 'skip')
```

The arrow form fails the test; the `function ()` form is correctly reported as pending (skipped).

Worth flagging why this is easy to miss: reverting would **not** turn CI red. Chromium always defines `PerformancePaintTiming`, so the guard is never taken and the broken line never executes. CI would stay green while the skip path throws on any engine that genuinely lacks the paint-timing API - reintroducing the "reports green without verifying anything" defect this ticket exists to remove, in a form CI cannot catch.

This is also called out in the ticket: "`this.skip()` requires a non-arrow function for the test callback so that `this` is the Mocha context."

So I've kept `function ()` as-is. For reference, the identical diff is approved on #3530 (`lts-2025`).

test/nuxeo-performance.test.js
vivekchaturved013: The conversion is required - with an arrow function `this.skip()` throws instead of skipping.

Arrow functions bind `this` lexically, and at the top level of an ES module `this` is `undefined`, so `this.skip` is unreachable. Running both shapes through this repo's Mocha (11.7.6) under the TDD UI, with the precondition deliberately unmet so the guard is actually taken:

```
  this.skip() and the callback form
    1) ARROW callback
    - FUNCTION callback

  0 passing, 1 pending, 1 failing

  1) ARROW callback:
     TypeError: Cannot read properties of undefined (reading 'skip')
```

The arrow form fails the test; the `function ()` form is correctly reported as pending (skipped).

Worth flagging why this is easy to miss: reverting would **not** turn CI red. Chromium always defines `PerformancePaintTiming`, so the guard is never taken and the broken line never executes. CI would stay green while the skip path throws on any engine that genuinely lacks the paint-timing API - reintroducing the "reports green without verifying anything" defect this ticket exists to remove, in a form CI cannot catch.

This is also called out in the ticket: "`this.skip()` requires a non-arrow function for the test callback so that `this` is the Mocha context."

So I've kept `function ()` as-is. For reference, the identical diff is approved on #3530 (`lts-2025`).

test/nuxeo-performance.test.js
madhurkulshrestha-hyland: Agreed that `this.skip()` needs the Mocha context and therefore can't be called from an arrow-function test body — no argument there.

The point of my note is that keeping `function ()` isn't the only way to get an honest skip. The guard can live in a `beforeEach` hook, which lets the test callbacks stay arrow functions:

```js
suite('performance paint timing', () => {
  beforeEach(function () {
    if (!window.PerformancePaintTiming) {
      this.skip();
    }
  });

  test('getFirstContentfulPaint returns null when no fcp entry', () => {
    // arrow function preserved
  });
});
```

Only the hook is a `function ()`; every `test(...)` stays an arrow. This reports a real pending/skip on engines lacking the paint-timing API — same honest behaviour your change delivers — while keeping the arrow-function style consistent with the rest of the file.

So it's not a correctness objection to your version; it's a preference for keeping arrow functions where we can. If you'd rather not restructure into a hook, your `function ()` form is fine to merge as-is. But since I'm going with the hook approach in the paired work, flagging it here so both PRs stay consistent.

vivekchaturved013: Thanks - and agreed the hook pattern is valid Mocha in general. It doesn't hold for this particular suite though, for one reason that I think matters more for your paired work than for this PR.

**This suite mixes two opposite preconditions.** `branch coverage: paint and timing fallbacks` has five tests, not just the two that carry the guard:

| Line | Test | Needs `PerformancePaintTiming` |
| --- | --- | --- |
| 401 | `getFirstPaint returns null when no first-paint entry` | **present** (guarded) |
| 415 | `getFirstPaint uses non-blank paint timing when paint API missing` | **absent** |
| 437 | `getFirstPaint uses msFirstPaint when timeToNonBlankPaint missing` | **absent** |
| 459 | `getFirstContentfulPaint returns null when paint timing type missing` | **absent** |
| 471 | `getFirstContentfulPaint returns null when no fcp entry` | **present** (guarded) |

The three middle tests each `delete window.PerformancePaintTiming` and assert the fallback path. A suite-level hook that skips when the API is unavailable skips all five - including the three written specifically for the engine that lacks it. On the exact engine where those three matter most, they would stop running. Running the hook form with the API unavailable:

```
  suite-level setup() guard, API unavailable
    - needs API PRESENT  - getFirstPaint returns null when no first-paint entry
    - needs API ABSENT   - uses non-blank paint timing when paint API missing
    - needs API ABSENT   - uses msFirstPaint when timeToNonBlankPaint missing
    - needs API ABSENT   - getFirstContentfulPaint null when paint type missing
    - needs API PRESENT  - getFirstContentfulPaint returns null when no fcp entry

  0 passing, 5 pending
```

A per-test guard is what keeps the skip scoped to the two tests that actually need the API present.

Minor, but it would bite on first run: the snippet uses `beforeEach`, which isn't defined here - this repo runs Mocha's TDD UI (`suite` / `test` / `setup`). A quick probe under `--ui tdd` reports `typeof beforeEach = undefined` and `typeof setup = function`, so it would throw a ReferenceError; `setup(function () { ... })` is the equivalent hook.

Mainly flagging this for the paired work: if that also touches this suite, a suite-level guard there would silently drop those three fallback tests.

madhurkulshrestha-hyland (approved):

